Issue #12, opened 2021-08-03
Milestone 2 tasks


VijayanB:
In this milestone, we focus on cleaning up existing code and make it ready for OpenSearch development

- [x] Remove X-Pack and non oss features/tests
- [x] Update namespace
- [x] Updates license headers
- [x] Remove docs
- [ ] Remove commits not related to >7.10 from *main* branch
- [ ] Update CI to run with  *OpenDistro* docker image with out security plugin

AmiStrn:
Do we intend to change also the types of searches not supported by opensearch that exist in the client? i.e. `combined_fields` search?

VachaShah:
@VijayanB do you think we can close this if all the tasks are addressed?

PR #97, opened 2022-02-04
Update endpoint id with prefix opensearch

imRishN:
### Description
Update endpoint id with prefix opensearch
 
### Issues Resolved
#12 
 
### Check List


mitalawachat (approved):

java-client/src/main/java/org/opensearch/client/opensearch/tasks/ListRequest.java
```diff
@@ -333,7 +333,7 @@ public ListRequest build() {
 	 * Endpoint "{@code tasks.list}".
 	 */
 	public static final Endpoint<ListRequest, ListResponse, ErrorResponse> _ENDPOINT = new SimpleEndpoint<>(
-			"es/tasks.list",
+			"opensearch/tasks.list",
```
VijayanB, line 336: Will removing "es/" instead of replacing have any failures? if not, i would remove it instead of replacing it.

PR #132, opened 2022-03-30
Breaking changes for type removal

VachaShah:
### Description
This PR contains the first set of breaking changes for 2.0 related to OpenSearch 2.0 - https://github.com/opensearch-project/OpenSearch/issues/2480

The following changes are included in this PR:

1. A CI workflow `test-integration-unreleased.yml` which runs integ tests for the client against unreleased OpenSearch. This will help in future development to prepare clients with the latest unreleased OpenSearch versions. Currently this workflow will test against 2.0 OpenSearch.
2. The CI workflow `test-integration.yml` is updated to work for released versions which the client is compatible with.
3. Upgraded the OpenSearch version in the client to 2.0.0-alpha1-SNAPSHOT.
4. Type mapping removal across the APIs and related test changes.
5. Remove `indices.exists_type` and `indices.flush_synced` and related files.
 
### Issues Resolved
Part of #139 
 
### Check List


.github/workflows/build.yml
```diff
@@ -17,6 +17,19 @@ jobs:
         with:
           java-version: '14'
           distribution: 'adopt'
+ 
+      - name: Checkout OpenSearch
```
reta, line 21: Hm ... why we are building `opensearch-project/opensearch` instead of using published snapshots from `https://aws.oss.sonatype.org/content/repositories/snapshots/`?

java-client/src/main/java/org/opensearch/client/opensearch/core/BulkRequest.java
```diff
@@ -580,15 +550,6 @@ public BulkRequest build() {
 					buf.append("/_bulk");
 					return buf.toString();
 				}
-				if (propsSet == (_index | _type)) {
-					StringBuilder buf = new StringBuilder();
-					buf.append("/");
-					SimpleEndpoint.pathEncode(request.index, buf);
```
reta, line 586: Seems like `_index` should not be dropped

VachaShah: The `_index` part is before this if condition. https://github.com/opensearch-project/opensearch-java/pull/132/files/1e01fbeae03463bf2d47b8d395bce3701901448d#diff-2bd95eccdb4274c5118be928fb9f466034f19fc2cdcfc3e5c82609585af94e0aL576

reta: Ah ... not visible in changeset, thank you!

.github/workflows/build.yml
VachaShah: Ah my bad! I can use that for build and unit test. I will update the workflows for the same.

.github/workflows/test-integration-unreleased.yml
```diff
@@ -0,0 +1,54 @@
...
+      - name: Assemble OpenSearch and Publish to Maven Local
+        run: |
+          cd opensearch
+          ./gradlew assemble
+
+        # This step runs the docker image generated during gradle assemble in OpenSearch. It is tagged as opensearch:test.
+        # Reference: https://github.com/opensearch-project/OpenSearch/blob/2.0/distribution/docker/build.gradle#L190
+      - name: Run Docker Image
+        run: |
+          docker run -p 9200:9200 -p 9600:9600 -d -e "discovery.type=single-node" -e "bootstrap.memory_lock=true" opensearch:test
```
reta, line 43: I am 100% sure there is a staging repository for Docker images, @dblock @peterzhuamazon could you please refer to the right one?

VachaShah: Yes there is one, but the staging image is with all the plugins which isn't ready yet. This is to test against min OpenSearch image.

VachaShah: I also tried the route of writing Dockerfile and related things to create a docker image for min OpenSearch. But then I found this gradle task that runs in assemble for OpenSearch so used this one. LMK if you think the Dockerfile route is better. I can use that instead.

VachaShah:
I will merge the changes from #134 once that is approved and merged and update the workflows with the appropriate JDK versions.

.github/workflows/test-integration-unreleased.yml
reta: I think publishing min Docker images (along with artifacts) would be the best option, may be we could quickly address that? Another possibility is to use OpenSearch test scaffolding (available as set of Gradle plugins) to run clusters from the test suites. What do you think?

VachaShah: I didn't use the Gradle plugins since I wanted to create a CI way for all the clients. It can work for java client but we would still end up in the same place for other clients.

reta: Agree, thank you, publishing min Docker images (along with artifacts) would be the answer, right?

VachaShah: Yes that would be the best option. Till that happens, is it fine if we use this change so we can start the development for 2.0? WDYT?

reta (approved):

CEHENKLE (approved):

.github/workflows/test-integration-unreleased.yml
VijayanB: I believe we will be adding API for plugins too and we need unreleased complete distribution for testing. Shall we create an issue to track this?

reta: Yes, please do, I think it will useful for many

VachaShah:
@reta @CEHENKLE @VijayanB A review again please? Just merged the changes from the main branch for the workflows and removed `indices.exists_type` as part of the breaking changes.

CEHENKLE (approved):

reta (approved):

owaiskazi19:
@VachaShah I see java client is using checkstyle for formatting. Moving forward do we have any plan to have Spotless as the formatting tool?

VachaShah:
> @VachaShah I see java client is using checkstyle for formatting. Moving forward do we have any plan to have Spotless as the formatting tool?

No plans as such but you can create a PR :)

VijayanB (approved):
LGTM. Thanks for making the changes.

VachaShah:
Not sure why the whitesource check is failing, the dependency was updated in #129. Will keep an eye on it.

java-client/build.gradle.kts
```diff
@@ -122,13 +123,14 @@ val integrationTest = task<Test>("integrationTest") {
...
 
-    val opensearchVersion = "1.2.4"
+    val opensearchVersion = "2.0.0-alpha1-SNAPSHOT"
     val jacksonVersion = "2.12.6"
     val jacksonDatabindVersion = "2.12.6.1"
 
     // Apache 2.0
     implementation("org.opensearch.client", "opensearch-rest-client", opensearchVersion)
     testImplementation("org.opensearch.test", "framework", opensearchVersion)
+    implementation("org.apache.logging.log4j", "log4j-core", "2.17.2")
```
dermot-hardy, line 133: Is this introducing a hard-coded dependency on the Log4j framework?  Surely the client should only be dependent on the facade?

dermot-hardy: #175

PR #134, opened 2022-03-30
Reducing test matrix to use JDK 11 and/or 17

### LINKED ISSUE #133: [FEATURE] Reduce the size of the test matrix
**Is your feature request related to a problem?**

We run too many tests in the CI matrix, takes a long time and is wasteful.

**What solution would you like?**

- Remove JDK 14, 15, only run 11 and 17
- Run integration tests against OpenSearch on a single JDK (e.g. some on 11, some on 17)
 
**Do you have any additional context?**

Use a matrix like this: https://github.com/slack-ruby/slack-ruby-bot-server/blob/master/.github/workflows/test-mongodb.yml#L10

VachaShah:
Signed-off-by: Vacha Shah <vachshah@amazon.com>

### Description
Reducing test matrix to use JDK 11 or 17 across OpenSearch versions.
 
### Issues Resolved
#133 
 
### Check List

### DIFF
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -14,33 +14,28 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        opensearch_version:
-          - 1.0.0
-          - 1.0.1
-          - 1.1.0
-          - 1.2.0
-          - 1.2.1
-          - 1.2.2
-          - 1.2.3
-          - 1.2.4
-          - 1.3.0
-        java:
-          - 8
-          - 14
-          - 15
-          - 17
+        entry:
+          - { opensearch_version: 1.0.0, java: 11 }
+          - { opensearch_version: 1.0.1, java: 11 }
+          - { opensearch_version: 1.1.0, java: 11 }
+          - { opensearch_version: 1.2.0, java: 11 }
+          - { opensearch_version: 1.2.1, java: 11 }
+          - { opensearch_version: 1.2.2, java: 11 }
+          - { opensearch_version: 1.2.3, java: 11 }
+          - { opensearch_version: 1.2.4, java: 11 }
+          - { opensearch_version: 1.3.0, java: 11 }
     steps:
-      - name: Set up JDK ${{ matrix.java }}
+      - name: Set up JDK ${{ matrix.entry.java }}
         uses: actions/setup-java@v1
         with:
-          java-version: ${{ matrix.java }}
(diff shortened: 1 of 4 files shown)


VijayanB (approved):

.github/workflows/test-integration.yml
```diff
@@ -14,33 +14,28 @@ jobs:
...
+        entry:
+          - { opensearch_version: 1.0.0, java: 11 }
+          - { opensearch_version: 1.0.1, java: 11 }
+          - { opensearch_version: 1.1.0, java: 11 }
+          - { opensearch_version: 1.2.0, java: 11 }
+          - { opensearch_version: 1.2.1, java: 11 }
+          - { opensearch_version: 1.2.2, java: 11 }
+          - { opensearch_version: 1.2.3, java: 11 }
+          - { opensearch_version: 1.2.4, java: 11 }
+          - { opensearch_version: 1.3.0, java: 11 }
```
saratvemulapalli, line 26: I understand we'd like to reduce the test matrix to LTS versions. 
For the versions which we have already shipped, isn't it worth testing with the version of bundled JDK?

VachaShah: Sure, is there a list of the versions of bundled JDK that we shipped?

saratvemulapalli: Thats a great question, I dont think there is a better way today. 
Scrappy way I can think of is via: https://github.com/opensearch-project/OpenSearch/blob/main/buildSrc/version.properties#L5 in each branch.

saratvemulapalli (approved):

.github/workflows/test-integration.yml
VachaShah: Sure thank you!

VijayanB: @saratvemulapalli @dblock Here jdk version corresponds to client’s jdk. Is there any reason that it has to be aligned with OpenSearch’s bundled jdk version?

VijayanB: I believe most (if not all) client users don’t have to know which jdk version their OpenSearch server is being used. IMO, we should maintain separate compatibility matrix for clients since same client can connect to multiple OpenSearch servers.

saratvemulapalli: Absolutely, 100%!

dblock: No reason to align. I would support as many JDKs as possible for the client.

Issue #158, opened 2022-05-16
[BUG] can't sort on field when document of that field contains null. fixed in elasticsearch-java.


joelbolt35:
**What is the bug?**
A full convo can be found here regarding the bug:
https://discuss.elastic.co/t/failure-with-sorts-and-new-co-elastic-clients-elasticsearch-java-client/291511/12

tl;dr: Sorting on a field where documents contain null breaks.

**What is the expected behavior?**
I am able to sort on a field where some documents contain NULL in those fields.

**Elastic Searches Fix**
https://github.com/elastic/elasticsearch-java/pull/68/files

dblock:
@joelbolt35 Please feel free to reimplement a fix in this project. If you do, please confirm that you're not copying from non-APLv2 elasticsearch-java.

chibenwa:
Hello there,

As a notice, on a migration effort for Apache James to OpenSearch, with the rest client 2.0.0 and OpenSearch 2.0.0 we encouter the issue...

https://github.com/Arsnael/james-project/blob/970ef4dce2f48f581376ace450dc5891f3f0ef7b/mailbox/opensearch/src/test/java/org/apache/james/mailbox/opensearch/ElasticSearchIntegrationTest.java#L426

Arsnael:
Hi guys! 

Tried to take a crack at it as mentioned by my friend above we encountered that issue too. I took an approach completely different from the fix from es so I don't think it should be an issue.

Cheers!

Arsnael:
Thanks for merging it team! Just want to know when can we expect a next release of the client with the fix?

dblock:
I believe we can close this with https://github.com/opensearch-project/opensearch-java/pull/169. 

We usually release clients along with the server, cc: @CEHENKLE

Arsnael:
> We usually release clients along with the server, cc: @CEHENKLE

@dblock No offense but I saw other clients, even the high level java one, having a 2.0.1 release along with the server already, but this one didn't follow up it seems (2.0.0 still being the latest). Should I be worried?

dblock:
@Arsnael I think we didn't release 2.0.1 because it wasn't needed. The 2.0 release works with 2.0.1.

I do understand that you want this bug fix released. Maybe we should start doing this more methodically? Care to open an issue saying "Release v. Next" and say "I need bug fix XYZ" or something like that?

PR #169, opened 2022-06-16
Sort field in Hit deserializer should handle null events

Arsnael:
Signed-off-by: Rene Cordier <rcordier@linagora.com>

### Description

Adds a new `StringOrNullDeserializer` to allow to handle the null events with strings (like with Int or Double), which then allows us to use this for the sort field in Hit deserializer, thus getting the null event case being handled in the array deserializer.

Would like to add I built it locally and tested on my project and it seems to solve the issue for sorts on my end.

Let me know if this is good for you guys :)

### Issues Resolved

Attempt to solve #158

### DIFF
--- a/java-client/src/main/java/org/opensearch/client/json/JsonpDeserializer.java
+++ b/java-client/src/main/java/org/opensearch/client/json/JsonpDeserializer.java
@@ -204,6 +204,10 @@ public interface JsonpDeserializer<V> {
         return new JsonpDeserializerBase.IntOrNullDeserializer(defaultValue);
     }
 
+    static JsonpDeserializer<String> stringOrNullDeserializer() {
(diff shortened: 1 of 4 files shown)


dblock:
Are there other instances where we should be using this?

dblock (approved):

Arsnael:
> Are there other instances where we should be using this?

No idea to be honest, the only issue I had so far was with sorts. Easy to fix later though if it's necessary for other instances IMO

chibenwa:
Any ETA of this fix in a public release @dblcok ?

VachaShah (approved):
Thank you!

dblock:
> Any ETA of this fix in a public release @dblcok ?

We release clients in sync with OpenSearch usually, I don't have any exact ETA though. @CEHENKLE maybe can comment on something more precise?

PR #173, opened 2022-06-30
Fix issue 172

mtimmerm:
Fix issue 172 -- erroneous implementation of double-check locking caused NPE in clients of LazyDeserializer.

### DIFF
--- a/java-client/src/test/java/org/opensearch/client/opensearch/json/JsonpMapperTest.java
+++ b/java-client/src/test/java/org/opensearch/client/opensearch/json/JsonpMapperTest.java
@@ -32,6 +32,7 @@
 
 package org.opensearch.client.opensearch.json;
 
+import org.opensearch.client.json.JsonpDeserializer;
 import org.opensearch.client.json.JsonpMapper;
 import org.opensearch.client.json.jackson.JacksonJsonpMapper;
 import org.opensearch.client.json.jsonb.JsonbJsonpMapper;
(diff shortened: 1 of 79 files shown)


dblock:
This needs a test, please?

java-client/src/test/java/org/opensearch/client/opensearch/json/JsonpMapperTest.java
```diff
@@ -116,6 +119,60 @@ public void testJacksonCustomJsonFactory() {
...
+        thread1.setDaemon(true);
+        thread1.start();
+        Thread thread2 = new Thread(threadProc);
+        thread2.setDaemon(true);
+        thread2.start();
+
+        // Wait until both threads are blocked waiting LazyDeserializer resolution
+        do {
+            try {
+                Thread.sleep(5);
```
dblock, line 162: Does it really take 5 seconds? Doubt it, maybe this should be very short to avoid hanging the test unnecessarily?

dblock (approved):
LGTM, will wait for an answer/update for below before merging

java-client/src/test/java/org/opensearch/client/opensearch/json/JsonpMapperTest.java
dblock: Oh 🤦 long day :)

ecastilloseekr:
Hello, thank you for this fix! Do you guys plan to create a tag with this fix anytime soon? Thank you.  @mtimmermansTa @Arsnael

dblock:
We have #174 open, working in it, but no immediate ETA, cc: @bbarani @gaiksaya.

PR #175, opened 2022-07-03
Remove unused Log4j 2 dependency

dermot-hardy:
Signed-off-by: Dermot Hardy <dermot.hardy@microfocus.com>

### Description
- Removes an unused dependency
- Removes a dependency on a logging framework

### Issues Resolved
A library intended for reuse from different projects should only depend on a logging facade like `slf4j-api` or `log4j-api`.  It should not depend on an actual logging framework.  This is somewhat academic anyway since so far as I can see the dependency is not used at all.


VachaShah (approved):
Thank you @dermot-hardy!

dblock (approved):

Issue #190, opened 2022-07-29
[FEATURE] Automate incrementing for the next developer iteration


dblock:
### Is your feature request related to a problem?

The workflow in #188 and #187 takes a tag and releases it.

Incrementing the version to the next developer iteration is manual.

### What solution would you like?

Add a workflow that:

1. Creates a tag for the current version
2. (the workflow added in #188 will run)
3. Increments the version for the next developer iteration.

PR #193, opened 2022-07-29
Extract version into gradle.properties.

dblock:
Signed-off-by: dblock <dblock@amazon.com>

### Description

The current construct of putting a version in `config/version.txt` is not very Java like. Moved the version setting to gradle.properties so it can be manipulated as needed in the future in #190. 


VachaShah (approved):
LGTM

gaiksaya (approved):

gradle.properties
```diff
@@ -0,0 +1 @@
+systemProp.version = 2.1.0
```
gaiksaya, line 1: I assume the `version` upper/lower case does not matter here?

dblock: Right.

PR #235, opened 2022-10-08
Remove "THIS CODE IS GENERATED" comments (#218)

### LINKED ISSUE #218: [PROPOSAL] Separate generated code or remove "THIS CODE IS GENERATED" comments
### What are you proposing?
Remove
```
//----------------------------------------------------
// THIS CODE IS GENERATED. MANUAL EDITS WILL BE LOST.
//----------------------------------------------------
```
where it does not make sense, or put the generated source code into a separate directory (source set) for clear separation.

### How did you come up with this proposal?
I was browsing the code to see if I could reuse the models/classes for a kotlin client when I came across many of these comments. From #197 #182 https://github.com/opensearch-project/opensearch-clients/issues/19 I see that the comment may be incorrect, at least for now. 

### What is the user experience going to be?
should be transparent to users of the library, but hopefully this makes the code more maintainable for contributors.

### Why should it be built? Any reason not to?
Maybe keeping common package code together? But I think the benefits of having a separate directory outweigh this point, even if only to make sure the code generator does not overwrite non-generated files.

### What will it take to execute?
time and review

### What are remaining open questions?
I have not yet worked with smithy, so are there any considerations to be made for that given it may be used in the near future?
Also, I am assuming that the entirety of the client is not (or will not be) generated. Is that a correct assumption?

AarjavP:
### Description
Removed comments that said the code is generated. Unfortunately generating java client from smithy spec is not possible yet
https://github.com/awslabs/smithy/issues/102#issuecomment-1267321093

Should this be removed as well?
https://github.com/opensearch-project/opensearch-java/blob/f8d9a5e0ac4ad4c5047d081b7da24feb0587bea2/config/checkstyle/checkstyle_suppressions.xml#L39-L43

### Issues Resolved
Closes #218

### DIFF
--- a/java-client/src/main/java/org/opensearch/client/opensearch/OpenSearchAsyncClient.java
+++ b/java-client/src/main/java/org/opensearch/client/opensearch/OpenSearchAsyncClient.java
@@ -30,10 +30,6 @@
  * GitHub history for details.
  */
 
-//----------------------------------------------------
-// THIS CODE IS GENERATED. MANUAL EDITS WILL BE LOST.
-//----------------------------------------------------
-
 package org.opensearch.client.opensearch;
 
 import org.opensearch.client.ApiClient;
--- a/java-client/src/main/java/org/opensearch/client/opensearch/OpenSearchClient.java
+++ b/java-client/src/main/java/org/opensearch/client/opensearch/OpenSearchClient.java
@@ -30,10 +30,6 @@
  * GitHub history for details.
  */
 
-//----------------------------------------------------
-// THIS CODE IS GENERATED. MANUAL EDITS WILL BE LOST.
-//----------------------------------------------------
-
 package org.opensearch.client.opensearch;
 
 import org.opensearch.client.ApiClient;
--- a/java-client/src/main/java/org/opensearch/client/opensearch/_types/AcknowledgedResponseBase.java
+++ b/java-client/src/main/java/org/opensearch/client/opensearch/_types/AcknowledgedResponseBase.java
@@ -30,10 +30,6 @@
  * GitHub history for details.
  */
 
-//----------------------------------------------------
-// THIS CODE IS GENERATED. MANUAL EDITS WILL BE LOST.
-//----------------------------------------------------
-
 package org.opensearch.client.opensearch._types;
 
 import org.opensearch.client.json.JsonpDeserializer;
--- a/java-client/src/main/java/org/opensearch/client/opensearch/_types/BaseNode.java
+++ b/java-client/src/main/java/org/opensearch/client/opensearch/_types/BaseNode.java
@@ -30,10 +30,6 @@
  * GitHub history for details.
  */
 
-//----------------------------------------------------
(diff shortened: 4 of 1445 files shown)


AarjavP:
The three line additions are coming from adding end of line at end of few files. 
- https://github.com/opensearch-project/opensearch-java/pull/235/files#diff-af65dcb0cbde2b9281acf79ab748b295ba4e8706938851dbc369dc41d97efdfb
- https://github.com/opensearch-project/opensearch-java/pull/235/files#diff-297be9b3c5f0513c74e5b6c2d97ae7f10778b6e023303c716229005bc087ff86
- https://github.com/opensearch-project/opensearch-java/pull/235/files#diff-4af3908c346925783d425653507de8ec3617d6f0cf362e8d16bc3f7d01d3ebf9

saratvemulapalli:
Thanks @AarjavP for opening up this PR. @VachaShah @harshavamsi for my learning do know the history ?

harshavamsi:
> Thanks @AarjavP for opening up this PR. @VachaShah @harshavamsi for my learning do know the history ?

My understanding is that the Java client along with a few others (rust?) were generated from [elastic spec](https://github.com/elastic/elasticsearch-specification). We don't generate from spec anymore. I think it makes sense to remove these lines for the sake of transparency. 

@AarjavP for context, you're right that we can't generate the java client from the smithy spec but we have successful internal prototypes of generating smithy specs from model and using the JSON AST to convert that to OpenAPI and use the [openapi-generator](https://github.com/OpenAPITools/openapi-generator) to generate clients. We're in the process of publishing our findings.

AarjavP:
just curious, where will I be able to find that report? Will it be a blog post?

I did try out generating a kotlin client but I dont think it generated all the models that this library currently has. Haven't looked into it further if there's help needed in writing additional smithy files, I can give it a shot!

VachaShah:
> My understanding is that the Java client along with a few others (rust?) were generated from [elastic spec](https://github.com/elastic/elasticsearch-specification). We don't generate from spec anymore. I think it makes sense to remove these lines for the sake of transparency.

Hi @AarjavP thank you for opening this PR! To answer your question, we have [opensearch-api-specification](https://github.com/opensearch-project/opensearch-api-specification) repository in which we are developing Smithy models to generate Smithy and OpenAPI specs. Please take a look :)

VachaShah (commented):
I think its a good idea to remove these lines since the code is not generated anymore. @AarjavP Do you think it would be better to separate the change into multiple PRs since it spans more than 1400 files?

AarjavP:
Sure I can separate them, how would you like it to be done? multiple commits in single PR or separate PRs? And separate the changes by package? or do you have certain top level packages in mind?

VachaShah:
> Sure I can separate them, how would you like it to be done? multiple commits in single PR or separate PRs? And separate the changes by package? or do you have certain top level packages in mind?

I think separate PRs would be helpful. I don't have preferences for the packages, but maybe divide them according to the number of files they have?

Issue #292, opened 2022-12-15
[BUG] SPI lookups on the HTTP client eventloop using JacksonJsonpMapper


chibenwa:
### What is the bug?

Using `JacksonJsonpMapper` causes frequent SPI lookups upon deserializing JSON which is very harmful:
 - Blocks the event loop (RandomAccessFile)
 - Costs significant CPU (26% of the event loop CPU from what I could see).

### How can one reproduce the bug?

```
        RestClientTransport transport = new RestClientTransport(lowLevelRestClient, new JacksonJsonpMapper());

        return new OpenSearchAsyncClient(transport);
```

### What is the expected behavior?

I expect that no SPI lookup to be done after resources are initialized. I expect SPI calls to be done statically upon APP startup.

### What is your host/environment?
_Operating system, version._

### Do you have any screenshots?

![Screenshot from 2022-12-15 10-57-19](https://user-images.githubusercontent.com/6928740/207793086-009063e2-38e3-45fe-acd9-4f2b13e34da3.png)

### Do you have any additional context?

Fix for Apache James: https://github.com/apache/james-project/pull/1061/commits/eb5a780527d42d3aa7cdd2acee15b797c8ab633b

PR #293, opened 2022-12-15
ISSUE-292 Prevent SPI calls at runtime

chibenwa:
Closes https://github.com/opensearch-project/opensearch-java/issues/292

### DIFF
--- a/java-client/src/main/java/org/opensearch/client/json/jackson/JsonValueParser.java
+++ b/java-client/src/main/java/org/opensearch/client/json/jackson/JsonValueParser.java
@@ -49,11 +49,13 @@ import java.io.IOException;
  * object (e.g. START_OBJECT, VALUE_NUMBER, etc).
  */
 class JsonValueParser {
-    private final JsonProvider provider = JsonProvider.provider();
+    private static class DefaultJsonProvider {
+        private static final JsonProvider INSTANCE = JsonProvider.provider();
+    }
 
     public JsonObject parseObject(JsonParser parser) throws IOException {
 
-        JsonObjectBuilder ob = provider.createObjectBuilder();
+        JsonObjectBuilder ob = DefaultJsonProvider.INSTANCE.createObjectBuilder();
 
         JsonToken token;
         while((token = parser.nextToken()) != JsonToken.END_OBJECT) {
@@ -68,7 +70,7 @@ class JsonValueParser {
     }
 
     public JsonArray parseArray(JsonParser parser) throws IOException {
-        JsonArrayBuilder ab = provider.createArrayBuilder();
+        JsonArrayBuilder ab = DefaultJsonProvider.INSTANCE.createArrayBuilder();
 
         while(parser.nextToken() != JsonToken.END_ARRAY) {
             ab.add(parseValue(parser));
@@ -94,22 +96,22 @@ class JsonValueParser {
                 return JsonValue.NULL;
 
             case VALUE_STRING:
-                return provider.createValue(parser.getText());
+                return DefaultJsonProvider.INSTANCE.createValue(parser.getText());
 
             case VALUE_NUMBER_FLOAT:
             case VALUE_NUMBER_INT:
                 switch(parser.getNumberType()) {
                     case INT:
-                        return provider.createValue(parser.getIntValue());
+                        return DefaultJsonProvider.INSTANCE.createValue(parser.getIntValue());
                     case LONG:
-                        return provider.createValue(parser.getLongValue());
+                        return DefaultJsonProvider.INSTANCE.createValue(parser.getLongValue());
                     case FLOAT:
                     case DOUBLE:
-                        return provider.createValue(parser.getDoubleValue());
+                        return DefaultJsonProvider.INSTANCE.createValue(parser.getDoubleValue());
                     case BIG_DECIMAL:
-                        return provider.createValue(parser.getDecimalValue());
+                        return DefaultJsonProvider.INSTANCE.createValue(parser.getDecimalValue());
                     case BIG_INTEGER:
-                        return provider.createValue(parser.getBigIntegerValue());
+                        return DefaultJsonProvider.INSTANCE.createValue(parser.getBigIntegerValue());
                 }
 
             default:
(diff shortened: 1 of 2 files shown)


wbeckler:
What's the issue this addresses?

wbeckler:
Ah I see it here: https://github.com/opensearch-project/opensearch-java/issues/292

java-client/src/main/java/org/opensearch/client/json/jackson/JsonValueParser.java
```diff
@@ -49,7 +49,7 @@
  * object (e.g. START_OBJECT, VALUE_NUMBER, etc).
  */
 class JsonValueParser {
-    private final JsonProvider provider = JsonProvider.provider();
+    private static final JsonProvider provider = JsonProvider.provider();
```
reta, line 52: Let's postpone the provider initialization till it is really needed? 

```suggestion
    private static class DefaultJsonProvider {
        private static final JsonProvider INSTANCE = JsonProvider.provider();
    }    
```

And than just replace `provider` with `DefaultJsonProvider.INSTANCE` ?

dblock:
Can we please add a unit test for this, too?

wbeckler:
I think it makes sense that this would go testless. On the other hand, it
would be awesome if you could open an issue suggesting an ongoing approach
to performance, to perhaps detect other low hanging fruit.

chibenwa:
Next week i'll propose a JMH bench for this changeset.

Regarding "detection of low hanging fruits" I use asyncprofiler on my workloads (memory & cpu). Can't recall other massive painpoints but I will get a second look, and share flame graphs here too.

Happy new year,

Regards,

Benoit

chibenwa:
[opensearch_2.4.0.zip](https://github.com/opensearch-project/opensearch-java/files/10334079/opensearch_2.4.0.zip)

Here is the flame graph I diagnosed this issue from. Taken with https://github.com/jvm-profiling-tools/async-profiler :

```
./profiler -d 120 -e itimer -f opensearch_2.4.0.html 1
```

(From within my applicative container)

Then searching `opensearch` (top left) gives you:
  - The driver thread (on the right) with the httpclient event loop
  - Applicatives calls to the driver from applicative threads (small stacktraces in the center)
  
Observations for the driver event loop:
  - Call to the SPI are very easy to notice CF issue. Not only it eats CPU/heap allocations but also does it blocks the HTTP event loop, which is catastrophic (my app is not OpenSearch heavy so that is not a big issue to me).
  - 0.5% of my application CPU (and 5% of the eventLoop CPU - similar footprint on heap allocation) is eaten by HTTP-client basic auth caching... :-( **Likely a low hanging fruit, shall I create an issue for this one?**
  - Unsurprisingly JSON parsing takes ~11% of the drivers CPU and 24% of heap allocation.
  - Event loop busyness (push stuff to the kernel network stack, SSL) takes around 60% of the event loop CPU, 25% of heap allocations, which again feels normal to me.
  - 2.3% of the event loop CPU is our binding to Reactor reactive library, which, again is normal: converting futures and enqueuing tasks takes time.

chibenwa:
## JMH output backing this change

### Before

```
Benchmark                Mode  Cnt   Score   Error  Units
JMHFieldBench.jsonBench  avgt    5  71.790 ± 2.125  us/op
```

### After

```
Benchmark                Mode  Cnt  Score   Error  Units
JMHFieldBench.jsonBench  avgt    5  1.418 ± 0.024  us/op
```

### Conclusion

x50 fastening of the JSOn parsing by getting rid of the SPI lookups, not event taking into account blocking operations... Huge win!

### Code of the benchmark

```
package org.apache.james.backends.opensearch;

import java.io.IOException;
import java.util.concurrent.TimeUnit;

import org.apache.james.backends.opensearch.json.jackson.JacksonJsonpParser;
import org.junit.Test;
import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.Mode;
import org.openjdk.jmh.infra.Blackhole;
import org.openjdk.jmh.runner.Runner;
import org.openjdk.jmh.runner.options.Options;
import org.openjdk.jmh.runner.options.OptionsBuilder;
import org.openjdk.jmh.runner.options.TimeValue;

import com.fasterxml.jackson.core.JsonParser;
import com.fasterxml.jackson.databind.ObjectMapper;

public class JMHFieldBench {

    public static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();

    @Test
    public void launchBenchmark() throws Exception {
        Options opt = new OptionsBuilder()
            .include(this.getClass().getName() + ".*")
            .mode (Mode.AverageTime)
            .timeUnit(TimeUnit.MICROSECONDS)
            .warmupTime(TimeValue.seconds(5))
            .warmupIterations(3)
            .measurementTime(TimeValue.seconds(2))
            .measurementIterations(5)
            .threads(1)
            .forks(1)
            .shouldFailOnError(true)
            .shouldDoGC(true)
            .build();

        new Runner(opt).run();
    }

    @Benchmark
    public void jsonBench(Blackhole bh) throws IOException {
        final JsonParser parser = OBJECT_MAPPER.createParser("[\"a\",\"b\"]");
        try {
            bh.consume(new JacksonJsonpParser(parser).getArrayStream());
        } catch (Exception e) {
            // do nothing
        } finally {
            parser.close();
        }
    }
}
```

reta:
It is known that `ObjectMapper` is expensive to create, making it  static certainly helps

chibenwa:
> It is known that ObjectMapper is expensive to create, making it static certainly helps

I do not see how this affects this benchmark

reta:
> > It is known that ObjectMapper is expensive to create, making it static certainly helps
> 
> I do not see how this affects this benchmark

Apologies, your benchmark checks the impact of changing the `JsonProvider.provider()` initialization (instance -> static), for some reason I thought that `JsonProvider.provider()` was instantiating ObjectMapper instances (which is not correct).

wbeckler:
Would you be up for adding a change to the changelog?

reta (approved):

dblock (approved):

Issue #287, opened 2022-12-03
[FEATURE] Support OpenSearch Serverless request signing


pingleig:
### Is your feature request related to a problem?

Current builtin `AwsSdk2Transport` does not work with serverless https://aws.amazon.com/opensearch-service/features/serverless/ because:

- It is only in 2.x version while severless is still using 1.x
- Different signing requirement https://docs.aws.amazon.com/opensearch-service/latest/developerguide/serverless-clients.html#serverless-signing
   - sign uses `es` while serverless requires `aoss`
   - sign`Content-Length` header while serverless does not use it
   - uses `Aws4Signer` while serverless uses ` Aws4UnsignedPayloadSigner`

External library does not work due to `Content-Length`

- `AwsRequestSigningApacheInterceptor.java`
https://github.com/acm19/aws-request-signing-apache-interceptor/blob/031791a8f11c3ecb2d687137ba36cc805442687a/src/main/java/io/github/acm19/aws/interceptor/http/AwsRequestSigningApacheInterceptor.java#L134

The official guide is using AWS SDK v1 https://github.com/awsdocs/amazon-opensearch-service-developer-guide/blob/master/sample_code/AmazonOpenSearchJavaClient-main/src/main/java/AmazonOpenSearchServiceSample.java

### What solution would you like?

- Document what customer can use when they want to interact with OpenSearch serverless using Java client. The full interceptor code is actually not long, e.g. [my modified version](https://github.com/pingleig/opensearch-unofficial-demo/blob/main/src/main/java/org/example/OpenSearchServerlessApacheInterceptor.java)
- Allow configure service name, signer and `Content-Length` behavior in the `AwsSdk2Transport` (but I am not sure if we can valid it ...

### What alternatives have you considered?

- Update https://github.com/acm19/aws-request-signing-apache-interceptor with an option to always skip content length 
- Backport the sign transport to v1 client

### Do you have any additional context?

I have a demo repo https://github.com/pingleig/opensearch-unofficial-demo

dblock:
Thanks for reporting this @pingleig! 

- I opened https://github.com/acm19/aws-request-signing-apache-interceptor/issues/88 
- We need to add whatever support is necessary to opensearch-java for Serverless (this issue)
- A custom interceptor should/can be added to the AWS docs, just like https://github.com/awsdocs/amazon-opensearch-service-developer-guide/pull/63

dblock:
[...]

PR #324, opened 2023-01-11
Added support for signing service name (partial support for OpenSearch Serverless)

dblock:
Signed-off-by: dblock <dblock@amazon.com>

### Description

Added support for signing service name. This is a necessary prerequisite for OpenSearch Serverless.

### Issues Resolved

Related to https://github.com/opensearch-project/opensearch-java/issues/287


java-client/src/test/java/org/opensearch/client/opensearch/integTest/aws/AwsSdk2TransportTestCase.java
```diff
@@ -150,6 +156,11 @@ protected String getTestClusterHost() {
         return cluster;
     }
 
+    protected String getTestClusterServiceName() {
+        String cluster = System.getProperty("tests.awsSdk2support.domainServiceName");
```
VachaShah, line 160: Call it `serviceName`?

java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
```diff
@@ -118,7 +143,54 @@ public AwsSdk2Transport(
...
+
+    /**
+     * Create an {@link OpenSearchTransport} with an ASYNCHRONOUS AWS Http client
+     * <p>
+     * Note that synchronous OpenSearch requests sent through this transport will be dispatched
+     * using the asynchronous client, but the calling thread will block until they are complete.
+     *
+     * @param asyncHttpClient HTTP client to use for OpenSearch requests
+     * @param host            The target host
+     * @param signingRegion The AWS region for which requests will be signed.  This should typically match
```
VachaShah, line 157: Also add `signingServiceName` for parameter documentation?

java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
```diff
@@ -96,20 +96,42 @@ public AwsSdk2Transport(
             @Nonnull String host,
             @Nonnull Region signingRegion,
             @CheckForNull AwsSdk2TransportOptions options) {
-        this(httpClient, null, host, signingRegion, options);
+        this(httpClient, null, host, "es", signingRegion, options);
```
reta, line 99:
```suggestion
        this(httpClient, null, host, "os", signingRegion, options);
```

reta: Seem like es -> elasticsearch? :)

reta: Oh sorry, I see it now - the value is coming from AWS

dblock: I think I'd prefer it to be a string because I am not so sure that list is finite. I'm trying to find out whether the signing service name can be other than `es` and `aoss` in the case of a proxy. Can we leave it as is for now?

reta (approved):

reta (approved):

VachaShah (approved):

Issue #215, opened 2022-08-30
[FEATURE] Support HTTP/2


arunx2:
### Is your feature request related to a problem?
No

### What solution would you like?
As the `opensearch-java` library requires Java 11, we can support HTTP/2 as it is much faster and more reliable than HTTP1. HTTP1 loads a single request for every TCP connection, while HTTP2 avoids network delay by using multiplexing. We can either leverage java `HttpClient` or square's `OkHttp` libraries

reta:
@arunx2 see please https://github.com/opensearch-project/OpenSearch/issues/4257

dblock:
What changes do we need to do in this library after the above was merged? @arunx2 want to contribute? At the least we need to document how to enable HTTP/2.

reta:
> What changes do we need to do in this library after the above was merged?

We basically need to introduce HTTP/2 compatible client transports, there are few ways we could approach that:
 - [Migrate client transports to Apache HttpClient / Core 5.x](https://github.com/opensearch-project/OpenSearch/issues/4256) - we that, we would get RestClient / RestHighLevelClient with HTTP/1.1 and HTTP/2 support out of the box
 - Introduce another implementation of the `OpenSearchTransport` (using fe Java's standard `HttpClient`)
Thanks.

dblock:
What do you think we should do @reta ?

wbeckler:
@reta what do you mean by "all supported OpenSearch clients." I would propose that opensearch-java be the only "supported" java client.

reta:
@wbeckler apologies for confusion, there are 3 clients right now: `RestClient`, `RestHighLevelClient` and `opensearch-java`. I think we want is to have OpenSearch and `opensearch-java` to live independent lives (no inter-dependencies in other words). That leaves us with: `RestClient` stays (to be used in core internally), `RestHighLevelClient` is deprecated (and hopefully removed), `opensearch-java`  becomes the official way to connect to OpenSearch clusters from JVM applications.

reta:
@dblock @wbeckler I think we could close this one: in 3.0.0 HTTP/2 is already supported (except AWS transport since it is based on Apache HttpClient 4.x).

dblock:
How about adding in README that this client supports HTTP/2, possibly a section in USER_GUIDE that explains how to use it (does it need to be turned on? does it just work?), and then closing this?

reta:
[...]

PR #330, opened 2023-01-16
Document HTTP/2 support

reta:
Signed-off-by: Andriy Redko <andriy.redko@aiven.io>

### Description
Document HTTP/2 support (https://github.com/opensearch-project/opensearch-java/issues/215)

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-java/issues/215

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -68,6 +68,8 @@ Transport transport = new RestClientTransport(restClient,
 OpenSearchClient client = new OpenSearchClient(transport);
 ```
 
+Upcoming OpenSearch `3.0.0` release brings HTTP/2 support and as such, the `RestClientTransport` would switch to HTTP/2 if available (for both HTTPS and/or HTTP protocols). The desired protocol could be forced using `RestClientBuilder.HttpClientConfigCallback`.
+
 ### Create a client using `ApacheHttpClient5Transport`
 
 ```java
@@ -77,6 +79,8 @@ final Transport transport = ApacheHttpClient5TransportBuilder
     .build();
 OpenSearchClient client = new OpenSearchClient(transport);
 ```
+
+Upcoming OpenSearch `3.0.0` release brings HTTP/2 support and as such, the `ApacheHttpClient5Transport` would switch to HTTP/2 if available (for both HTTPS and/or HTTP protocols). The desired protocol could be forced using `ApacheHttpClient5TransportBuilder.HttpClientConfigCallback`.
 
 ## Create an index
 
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -56,14 +56,14 @@ There are multiple low level transports which `OpenSearchClient` could be config
 ### Create a client using `RestClientTransport`
 
 ```java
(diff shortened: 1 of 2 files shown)


USER_GUIDE.md
```diff
@@ -68,6 +68,8 @@ Transport transport = new RestClientTransport(restClient,
 OpenSearchClient client = new OpenSearchClient(transport);
 ```
 
+Upcoming OpenSearch `3.0.0` release brings HTTP/2 support and as such, the `RestClientTransport` would switch to HTTP/2 if available (for both HTTPS and/or HTTP protocols). The desired protocol could be forced using `RestClientBuilder.HttpClientConfigCallback`.
```
dblock, line 71: How can I see what protocol has been chosen?
Example of how I would switch to/from HTTP/2?

reta: > How can I see what protocol has been chosen?

Basically this could be done by:
 - inspecting the traffic
 - logging requests and responses (they contain HTTP protocol version)

I am not sure what kind of prove we could provide here. 

>  Example of how I would switch to/from HTTP/2?
`RestClientBuilder.HttpClientConfigCallback` should be documented in core but I will add the configuration snippet for Apache HttpClient 5 transport.

USER_GUIDE.md
```diff
@@ -56,28 +56,46 @@ There are multiple low level transports which `OpenSearchClient` could be config
 ### Create a client using `RestClientTransport`
 
 ```java
-Transport transport = new RestClientTransport(restClient, new JacksonJsonpMapper()); 
+OpenSearchTransport transport = new RestClientTransport(restClient, new JacksonJsonpMapper());
```
reta, line 59: Just noticed that `OpenSearchClient` does not accept `Transport` but `OpenSearchTransport`, fixed that

dblock (approved):
Perfect.

USER_GUIDE.md
dblock: For documentation generally I ask myself "what would I do". So if you think logging is how you'd do it, I'd show here (or somewhere in the docs) how to log the HTTP version chosen.

reta: > For documentation generally I ask myself "what would I do". So if you think logging is how you'd do it, I'd show here (or somewhere in the docs) how to log the HTTP version chosen.

This is not something we should be worried about here (I think), there are few things involved on our behalf:
 - for HTTPS, we have the ALPN protocol in place 
 - for HTTP, we have `Upgrade` in place

The OpenSearch Java client could connect to `2.x` clusters fe, that would mean using HTTP/1.1 no matter what. Generally, forcing the protocol is the only reliable way (I am aware of) to make sure only HTTP/1.1 or only HTTP/2 has to be used. 

To this point, we would be building features on top of HTTP/2 very soon (I believe), so the presence of HTTP/2 would be obvious.

dblock: Fair enough.

reta:
@VachaShah mind a review please (documentation update)? thank you!

VachaShah (approved):

PR #339, opened 2023-01-21
Adding x-amz-content-sha256 header for signed requests

### LINKED ISSUE #287: [FEATURE] Support OpenSearch Serverless request signing

VachaShah:
### Description

X-Amz-Content-Sha256 is a required header for Serverless. 

_Tested this code against both Managed Service and Serverless._

### Issues Resolved
#287 


dblock (changes requested):
Add a test that checks that the `x-amz-...` header is added?

java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
```diff
@@ -363,6 +363,9 @@ private <RequestT> SdkHttpFullRequest prepareRequest(
             }
             req.putHeader("Content-Length", String.valueOf(body.getContentLength()));
             req.contentStreamProvider(body::getInputStream);
+            // To add the "X-Amz-Content-Sha256" header, it needs to set as required.
+            // It is a required header for Amazon OpenSearch Serverless.
+            req.putHeader("x-amz-content-sha256", "required");
```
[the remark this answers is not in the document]
VachaShah: Yeah it took me a while to find too. We use `Aws4Signer` which does not have the sign method overriden like `Aws4UnsignedPayloadSigner`.  According to documentation, the `Aws4UnsignedPayloadSigner` is similar to `Aws4Signer` but just adds `UNSIGNED-PAYLOAD` when protocol is HTTPS. Can we use `Aws4UnsignedPayloadSigner`?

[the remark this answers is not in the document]
dblock: If it works we sure can I think.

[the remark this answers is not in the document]
VachaShah: I tried the `Aws4UnsignedPayloadSigner` but it does not work for Amazon OpenSearch Service since it signs the payload with `UNSIGNED_PAYLOAD` over https protocol. Works for Amazon OpenSearch Serverless. I can PR this change on their repo may be as a new signer class but for now looks like we might have to use the hard-coded "required".

VachaShah:
> Add a test that checks that the `x-amz-...` header is added?

Added test.

java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
```diff
@@ -297,7 +297,7 @@ public void close() {
     }
 
     @CheckForNull
-    private <RequestT> OpenSearchRequestBodyBuffer prepareRequestBody(
+    public <RequestT> OpenSearchRequestBodyBuffer prepareRequestBody(
```
reta, line 300: Those should probably not be public

VachaShah: I had to do that since I had to write a test for the header :(

dblock: I think we need to find a way. How about making it protected and subclassing AwsSdk2Transport?

VachaShah: Do you mean something like this?

```
public class AwsSdk2TransportMock extends AwsSdk2Transport {

    public AwsSdk2TransportMock(
            @Nonnull SdkHttpClient httpClient,
            @Nonnull String host,
            @Nonnull String signingServiceName,
            @Nonnull Region signingRegion,
            @CheckForNull AwsSdk2TransportOptions options) {
        super(httpClient, null, host, signingServiceName, signingRegion, options);
    }

    @Override
    public <RequestT> OpenSearchRequestBodyBuffer prepareRequestBody(RequestT request,
            Endpoint<RequestT, ?, ?> endpoint, TransportOptions options) throws IOException {
        return super.prepareRequestBody(request, endpoint, options);
    }
    @Override
    public <RequestT> SdkHttpFullRequest prepareRequest(RequestT request, Endpoint<RequestT, ?, ?> endpoint,
            TransportOptions options, OpenSearchRequestBodyBuffer body) {
        return super.prepareRequest(request, endpoint, options, body);
    }
}
```
where the `prepareRequest` and `prepareRequestBody` in AwsSdk2Transport are `protected`?

java-client/src/test/java/org/opensearch/client/opensearch/integTest/aws/AwsSdk2TransportMock.java
```diff
@@ -0,0 +1,46 @@
...
+import software.amazon.awssdk.http.SdkHttpClient;
+import software.amazon.awssdk.http.SdkHttpFullRequest;
+import software.amazon.awssdk.regions.Region;
+
+import java.io.IOException;
+
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+
+public class AwsSdk2TransportMock extends AwsSdk2Transport {
```
dblock, line 25: I don't think that "mock" is correct naming here. That implies something that "fakes" a certain operation, which this does not.

dblock (changes requested):
I think while the inheritance here works around the problem, it's a bad idea. This test is way too complicated for something as simple as a header.

@VachaShah I vote to remove `testContentShaHeader` (and the new mock class), and manually confirm that `./gradlew integrationTest --tests "*AwsSdk2SearchIT*" -Dtests.awsSdk2support.domainHost=$ENDPOINT -Dtests.awsSdk2support.domainRegion=us-west-2 ...` works for both managed OpenSearch and Serverless.

@reta do you have any other ideas?

java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
dblock: Right. Looking at your implementation I think that was not a good suggestion, sorry. I say we open a bug for this and call it a day instead of adding the child class.

reta:
> @reta do you have any other ideas?

Sorry for delay, I thought about simple approach with test interceptor [1]:

```
    protected SdkHttpClient getHttpClient() {
        if (httpClient == null) {
            httpClient = ApacheHttpClient
            		.builder()
            		.buildWithDefaults(AttributeMap.builder()
            			.put(SdkClientOption.EXECUTION_INTERCEPTORS, List.of(/* test interceptor */))
            			.build());
        }
        return httpClient;
    }
```

By adding one, we should be able to verify presence of all headers before request is sent off.

[1] https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/ExecutionInterceptor.java#L198

VachaShah:
> I think while the inheritance here works around the problem, it's a bad idea. This test is way too complicated for something as simple as a header.
> 
> @VachaShah I vote to remove `testContentShaHeader` (and the new mock class), and manually confirm that `./gradlew integrationTest --tests "*AwsSdk2SearchIT*" -Dtests.awsSdk2support.domainHost=$ENDPOINT -Dtests.awsSdk2support.domainRegion=us-west-2 ...` works for both managed OpenSearch and Serverless.
> 
> @reta do you have any other ideas?

Sure! I removed the test and the sub class.

For confirmation, I have manually tested this change against both OpenSearch and OpenSearch Serverless with the integ tests and a client app.

VachaShah:
> > I think while the inheritance here works around the problem, it's a bad idea. This test is way too complicated for something as simple as a header.
> > @VachaShah I vote to remove `testContentShaHeader` (and the new mock class), and manually confirm that `./gradlew integrationTest --tests "*AwsSdk2SearchIT*" -Dtests.awsSdk2support.domainHost=$ENDPOINT -Dtests.awsSdk2support.domainRegion=us-west-2 ...` works for both managed OpenSearch and Serverless.
> > @reta do you have any other ideas?
> 
> Sure! I removed the test and the sub class.
> 
> For confirmation, I have manually tested this change against both OpenSearch and OpenSearch Serverless with the integ tests and a client app.

Actually, for OpenSearch Serverless, wait for refresh policy does not work. 

> > @reta do you have any other ideas?
> 
> Sorry for delay, I thought about simple approach with test interceptor [1]:
> 
> ```
>     protected SdkHttpClient getHttpClient() {
>         if (httpClient == null) {
>             httpClient = ApacheHttpClient
>             		.builder()
>             		.buildWithDefaults(AttributeMap.builder()
>             			.put(SdkClientOption.EXECUTION_INTERCEPTORS, List.of(/* test interceptor */))
>             			.build());
>         }
>         return httpClient;
>     }
> ```
> 
> By adding one, we should be able to verify presence of all headers before request is sent off.
> 
> [1] https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/ExecutionInterceptor.java#L198

Thats interesting. I will add this test in an upcoming PR.

dblock (dismissed):
I’m good with this. We can improve tests and what not later.

VachaShah:
> > I think while the inheritance here works around the problem, it's a bad idea. This test is way too complicated for something as simple as a header.
> > @VachaShah I vote to remove `testContentShaHeader` (and the new mock class), and manually confirm that `./gradlew integrationTest --tests "*AwsSdk2SearchIT*" -Dtests.awsSdk2support.domainHost=$ENDPOINT -Dtests.awsSdk2support.domainRegion=us-west-2 ...` works for both managed OpenSearch and Serverless.
> > @reta do you have any other ideas?
> 
> Sure! I removed the test and the sub class.
> 
> For confirmation, I have manually tested this change against both OpenSearch and OpenSearch Serverless with the integ tests and a client app.

Actually, I wrote too fast. The integ tests for serverless might need to be modified since there are certain things that don't work for Serverless like wait for refresh policy. Manual testing it with a client app works successfully.

dblock:
@VachaShah you can remove the refresh policy altogether, that’s what I did in the demos

make sure dev guide says all the right things

VachaShah:
> @VachaShah you can remove the refresh policy altogether, that’s what I did in the demos
> 
> make sure dev guide says all the right things

Removing the refresh policy address a problem but then other issues stem up. Integ tests would need to be modified to work with Serverless.

dblock:
> > @VachaShah you can remove the refresh policy altogether, that’s what I did in the demos
> > make sure dev guide says all the right things
> 
> Removing the refresh policy address a problem but then other issues stem up. Integ tests would need to be modified to work with Serverless.

Do they pass with a sleep, or at least once? Since we aren’t running them with automation rn, I would consider that good enough.

VachaShah:
> > > @VachaShah you can remove the refresh policy altogether, that’s what I did in the demos
> > > make sure dev guide says all the right things
> > 
> > 
> > Removing the refresh policy address a problem but then other issues stem up. Integ tests would need to be modified to work with Serverless.
> 
> Do they pass with a sleep, or at least once? Since we aren’t running them with automation rn, I would consider that good enough.

They passed with a sleep once without the refresh policy. But keep failing intermittently.

dblock (dismissed):
Serverless may have a cold start thing going on causing intermittent delays, I would ignore that.

VachaShah:
@dblock Review again please? Github dismissed your approval 🤦🏼‍♀️

dblock (approved):
@reta good with you?

java-client/src/test/java/org/opensearch/client/opensearch/integTest/aws/AwsSdk2SearchIT.java
```diff
@@ -77,12 +78,15 @@ void testClientAsync(boolean async) throws Exception {
...
         SimplePojo doc2 = new SimplePojo("Document 2", "The text of document 2");
-        CompletableFuture<IndexResponse> add2 = addDoc(client, "id2", doc2, false);
+        CompletableFuture<IndexResponse> add2 = addDoc(client, "id2", doc2);
         SimplePojo doc3 = getLongDoc("Long Document 3", 1000000);
         CompletableFuture<IndexResponse> add3 = CompletableFuture.allOf(add1, add2).thenCompose(
-                unused -> addDoc(client, "id3", doc3, true));
+                unused -> addDoc(client, "id3", doc3));
+
+        // wait for the document to index
+        Thread.sleep(1000);
```
reta, line 89: This is a recipe for flakyness ....

reta (approved):

reta:
> @reta good with you?

@dblock we need to improve tests right away, using something like `https://github.com/awaitility/awaitility` to replace `Thread.sleep()` calls.

VachaShah:
> > @reta good with you?
> 
> @dblock we need to improve tests right away, using something like `https://github.com/awaitility/awaitility` to replace `Thread.sleep()` calls.

@reta I will work on doing that as part of #343.

dblock:
> > @reta good with you?
> 
> @dblock we need to improve tests right away, using something like `https://github.com/awaitility/awaitility` to replace `Thread.sleep()` calls.

For sure. It’s not that simple and those tests don’t run. We’ll queue up getting infrastructure and repeatable tests for managed and serverless for all clients. Cc: @wbeckler

PR #352, opened 2023-01-26
Update USER_GUIDE to use exact AWS brand names

andrross:
Minor update to use exact brand names. "OpenSearch" is rather overloaded to I find it less confusing to be precise in the naming.


reta (dismissed):

USER_GUIDE.md
```diff
@@ -175,9 +175,9 @@ DeleteIndexResponse deleteIndexResponse = client.indices().delete(deleteIndexReq
 
 # Using different transport options
 
-## Amazon Managed OpenSearch
+## Amazon OpenSearch Service
```
dblock, line 178: Update TOC

Should we call these "Managed OpenSearch and OpenSearch Serverless"?

dblock (changes requested):
TOC needs updating

USER_GUIDE.md
andrross: The section header could be "Amazon OpenSearch Service and OpenSearch Serverless" though. Would that be better?

dblock: Yes, so we have 

https://docs.aws.amazon.com/opensearch-service/latest/developerguide/request-signing.html that says "Amazon OpenSearch Service" and https://docs.aws.amazon.com/opensearch-service/latest/developerguide/serverless-clients.html that says "Amazon OpenSearch Serverless" 

So I think that's the correct wording, if you want them combined "Amazon OpenSearch Service" + "Amazon OpenSearch Serverless" = ""Amazon OpenSearch Service and Serverless"?

dblock (approved):

VachaShah (approved):

Issue #284, opened 2022-11-29
[PROPOSAL] Experiment with using Smithy API spec to generate missing APIs


Xtansia:
### What are you proposing?
Performing an experiment using the [WIP Smithy API specification for OpenSearch](https://github.com/opensearch-project/opensearch-api-specification) to generate missing functionality in the client for a small API without re-creating the entire client.

### What users have asked for this feature?
- [Proposal for using API specifications to generate all clients](https://github.com/opensearch-project/opensearch-clients/issues/19)
- [Request to add remote store API](https://github.com/opensearch-project/opensearch-java/issues/168)

### What problems are you trying to solve?
- Reduce friction in adoption of new OpenSearch APIs & features in the language clients.
- Reduce maintenance overhead

### What is the developer experience going to be?
A script/tool that can be run to automatically fetch the API spec and generate the chosen code.

#### Are there any security considerations? 
n/a

#### Are there any breaking changes to the API
This experiment is only successful if it does not break compatibility.

### What is the user experience going to be?
Same as developer experience.

#### Are there breaking changes to the User Experience?
See above.

### Why should it be built? Any reason not to?
The learnings from this experiment, if successful:
- Will be broadly applicable to all language clients.
- Will reduce workloads for maintainers.
- Will speed up adoption of new APIs in clients.
- Will fill in gaps in API coverage
If the experiment is unsuccessful, it will not be merged, negating almost any potential downsides.
It may not be feasible to integrate generated code from an off-the-shelf generator with the pre-existing client code.

### What will it take to execute?
- Will use the `/_remotestore/_restore` API as a low-risk low-touch use-case
- Potentially configuring or customising an off-the-shelf generator, at worst hand-rolling a minimal generator for our use-case.

Xtansia:
## Options Explored

### Pre-existing OpenAPI generator
A quick inspection of the existing Java generator for [openapi-generator](https://github.com/OpenAPITools/openapi-generator/) shows it is generally only workable when generating completely from scratch, as it expects to be generating a whole client, project files and all. There is some flexibility within the CLI tool to reduce surface area of what you desire to generate, however not to level necessary. All the generated code is also specific to whatever HTTP & JSON libraries have had templates written for. 

Verdict: 👎  Any generated code would be completely independent of the existing client which is not desirable. 

### Custom OpenAPI templates/generator
[...]

PR #366, opened 2023-02-17
Generate client code from OpenAPI specification

Xtansia:
### Description
As I mentioned in https://github.com/opensearch-project/opensearch-java/issues/284#issuecomment-1427164576 this is a PoC creating a code generator that uses OpenAPI specs from scratch (ie. not based off openapi-generator). The code is not feature complete and not production ready.

Please direct general comments/discussion about the topic to the [issue](https://github.com/opensearch-project/opensearch-java/issues/284) to keep things consolidated.


reta:
@Xtansia do you have an examples of request / response models the generator produce? thank you!

Xtansia:
> @Xtansia do you have an examples of request / response models the generator produces? thank you!

@reta Yes, they're included in this PR, the remote store ones. Under the `java-client/src/generated/java` directory

reta:
> > @Xtansia do you have an examples of request / response models the generator produces? thank you!
> 
> @reta Yes, they're included in this PR, the remote store ones. Under the `java-client/src/generated/java` directory

Sorry missed that, they look pretty close to what we have inherited (comparing to just template changes).

harshavamsi:
@Xtansia this is amazing work! The generated client and req/res look very similar to the existing APIs. From a quick glance, I don't see a `Response.mustache` template. How is the response handler being generated?

dblock:
love it, how can we move it closer to something we can merge?

Xtansia:
@dblock & @reta, I have added the warning header & refactored the transform logic to what I feel is a lot clearer. I've also managed to pull the few more complicated mustache "lambdas"/getters out to primarily templates, with some more minor code logic just to handle recursiveness. 
There's a handful of features that'll need to be implemented eventually, for example when we want to regenerate some of the operations that merge multiple endpoints into one operation/request shape (ie. index a doc with or without ID). 
Otherwise it's primarily waiting for having an actual published spec from https://github.com/opensearch-project/opensearch-api-specification and probably some kind of configuration for filtering down the desired operations to generate.

reta:
@Xtansia this is pretty cool, just from your perspective as implementor: how difficult it would be to keep OpenAPI and our own generator in sync (if it makes at all)? Specifically, I have in mind bugfixes and missing OpenAPI 3.1 features (the 3.1 support is still experimental and not fully implemented). 

As for a spec, I've found this outdated but pretty comprehensive snippet [1], I converted it to 3.x using Swagger Editor, we could try to use it meantime (for a sake to filling any implementation gaps), thanks

Xtansia:
@dblock @reta @VachaShah This is ready to be looked at when you all have time. This PR doesn't include any generation of new code or automations yet. Aiming to have this merged to be able to further iterate on it and make piecemeal PRs generating new operations or re-generating existing ones. And adding whatever further improvements needed in future PRs. 

An example of what this generates is here: https://github.com/Xtansia/opensearch-java/commit/4a9898ec8466a1ddfcbe87bfdde255f20417905b

dblock (approved):
Did a cursory look, nothing major to note, I'd merge this as is.

reta:
> @dblock @reta @VachaShah This is ready to be looked at when you all have time.

Thanks @Xtansia , I will take a look shortly, just wanted to confirm you think that having this project as part of `opensearch-java` is a way to? The reason I am asking is that I expect there would be a lot of changes there and the release cadence of generator and client might be very different (at least initially), having a separate repo would mitigate that quite a lot.

java-codegen/build.gradle.kts
```diff
@@ -0,0 +1,290 @@
...
+    implementation("commons-cli", "commons-cli", "1.8.0")
+    implementation("commons-logging", "commons-logging", "1.3.2")
+    implementation("org.apache.commons", "commons-lang3", "3.14.0")
+    implementation("org.apache.commons", "commons-text", "1.12.0")
+    implementation("org.apache.logging.log4j", "log4j-api", "[2.17.1,3.0)")
+    implementation("org.apache.logging.log4j", "log4j-core", "[2.17.1,3.0)")
+    implementation("org.apache.logging.log4j", "log4j-slf4j2-impl", "[2.17.1,3.0)")
+
+    // Apache 2.0
+    implementation("com.fasterxml.jackson.core", "jackson-core", "2.15.2")
```
reta, line 143:
```suggestion
    implementation("com.fasterxml.jackson.core", "jackson-core", "2.17.1")
```

java-codegen/build.gradle.kts
```diff
@@ -0,0 +1,290 @@
...
+    implementation("com.fasterxml.jackson.core", "jackson-databind", "2.15.2")
```
reta, line 144:
```suggestion
    implementation("com.fasterxml.jackson.core", "jackson-databind", "2.17.1")
```

java-codegen/src/main/java/org/opensearch/client/codegen/openapi/OpenApiSpecification.java
```diff
@@ -0,0 +1,135 @@
...
+import javax.annotation.Nullable;
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.apache.commons.lang3.builder.ToStringBuilder;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.opensearch.client.codegen.exceptions.ApiSpecificationParseException;
+import org.opensearch.client.codegen.utils.Maps;
+
+public class OpenApiSpecification extends OpenApiElement<OpenApiSpecification> {
```
reta, line 31: You definitely put a lot of thoughts into it, but I am just curious do we really need to have these many wrappers around `io.swagger.v3.oas.models.*` ? Couldn't we use the OAS model classes directly?

Xtansia:
> > @dblock @reta @VachaShah This is ready to be looked at when you all have time.
> 
> Thanks @Xtansia , I will take a look shortly, just wanted to confirm you think that having this project as part of `opensearch-java` is a way to go? The reason I am asking is that I expect there would be a lot of changes there and the release cadence of generator and client might be very different (at least initially), having a separate repo would mitigate that quite a lot.

That's a fair point and I think in the long run it would make sense to have this in a separate repo. However at the moment it's not yet at a stable point where this makes sense to actively publish for others. The first primary goal is in getting part or all of the Java client being generated so this initial period of iteration and development is going to be inextricably tied to attempts to generate more operations and covering any missed edge cases etc. So I think it's reasonable for it to live here until it has stabilised somewhat as long as we aren't releasing it from here.

VachaShah (commented):
This is great!! The only comment I have (after a high level review) looking from previous comments is: if at all there is a chance for this generator to be moved out of this repo, could we make the generator classes restricted to use for consumers? Mostly no user would end up using them but if we move out public classes out of the repo, we might end up with a major version increment.

dblock (approved):
Merge?

reta:
> Merge?

No objections, we are not going to backport it to `2.x`, right?

dblock:
> > Merge?
> 
> No objections, we are not going to backport it to `2.x`, right?

Why not? Assuming we don't introduce breaking changes we should be able to swap manual authoring for a generator.

Leaving this to @Xtansia.

dblock:
> > Why not? Assuming we don't introduce breaking changes we should be able to swap manual authoring for a generator.
> 
> If we do, we need clearly document that this is new experimental module with all the limitations it has now - we do make releases from 2.x and the new module will be published

I don't think I'm following, the generator is just a tool, the actual code of the client being released will not change.

reta:
> I don't think I'm following, the generator is just a tool, the actual code of the client being released will not change.

Correct, but will we release the tool as part of client release?

dblock:
> > I don't think I'm following, the generator is just a tool, the actual code of the client being released will not change.
> 
> Correct, but will we release the tool as part of client release?

I don't think so. Extracting the tool into something releasing on itself could make sense, but otherwise no.

reta:
> I don't think so. Extracting the tool into something releasing on itself could make sense, but otherwise no.

No objections, sounds good to me

Xtansia:
> > I don't think I'm following, the generator is just a tool, the actual code of the client being released will not change.
> 
> Correct, but will we release the tool as part of client release?

@reta I've removed the publish config from the build.gradle of the generator so it won't get published

dblock:
@Xtansia merge at will

Xtansia:
@dblock I'm happy for it to be merged if everyone else is, I can't do it myself as I'm not a maintainer

reta (approved):

dblock:
> @dblock I'm happy for it to be merged if everyone else is, I can't do it myself as I'm not a maintainer

Oh yeah? ;) Looking forward to that spec to be fully generated.

Xtansia:
> This is great!! The only comment I have (after a high level review) looking from previous comments is: if at all there is a chance for this generator to be moved out of this repo, could we make the generator classes restricted to use for consumers? Mostly no user would end up using them but if we move out public classes out of the repo, we might end up with breaking changes and a major version increment.

@VachaShah I would think this should not be an issue as long as we aren't publishing the generator to maven etc. For someone to rely on the classes of the generator they'd need to be building from source. AFAIK Java doesn't have a good mechanism to restrict access to the classes in a useful way as they need to be `public` to be accessed by the classes in other parts of the generator.

PR #367, opened 2023-02-17
Upgrading compatibility to opensearch 2.5

### LINKED ISSUE #285: [BUG] Compatibility matrix is missing compatible server versions
### What is the bug?
There is no mention of server version 2.4 here: https://github.com/opensearch-project/opensearch-java/blob/main/COMPATIBILITY.md

It would also be useful to know whether the client can work with older versions of opensearch or elasticsearch, so people can know if they can upgrade their client before upgrading their server.

AmiStrn:
### Description
Client version compatibility matrix required updating to the later versions of OpenSearch
Upgrading the tests to 2.4, 2.5 and updating the COMPATIBILITY.doc

### Issues Resolved
resolves #285 

### DIFF
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -6,10 +6,10 @@
 The below matrix shows the compatibility of the [`opensearch-java-client`](https://search.maven.org/artifact/org.opensearch.client/opensearch-java) with versions of [`OpenSearch`](https://opensearch.org/downloads.html#opensearch).
 
 | Client Version | OpenSearch Version |
-| --- | --- |
-| 1.0.0 | 1.0.0-1.3.3 |
-| 2.0.0 | 1.3.3-2.0.1 |
-| 2.1.0 | 1.3.3-2.3.0 |
+| --- |--------------------|
+| 1.0.0 | 1.0.0-1.3.3        |
+| 2.0.0 | 1.3.3-2.0.1        |
+| 2.1.0 | 1.3.3-2.5.0        |
 
 ## Upgrading
 
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -6,10 +6,11 @@
 The below matrix shows the compatibility of the [`opensearch-java-client`](https://search.maven.org/artifact/org.opensearch.client/opensearch-java) with versions of [`OpenSearch`](https://opensearch.org/downloads.html#opensearch).
 
 | Client Version | OpenSearch Version |
-| --- |--------------------|
-| 1.0.0 | 1.0.0-1.3.3        |
-| 2.0.0 | 1.3.3-2.0.1        |
-| 2.1.0 | 1.3.3-2.5.0        |
+|----------------|--------------------|
+| 1.0.0          | 1.0.0-1.3.3        |
+| 2.0.0          | 1.3.3-2.0.1        |
+| 2.1.0          | 1.3.3-2.3.0        |
+| 2.2.0          | 1.3.3-2.5.0        |
 
 ## Upgrading
 
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -17,6 +17,8 @@ jobs:
           - { opensearch_version: 2.1.0, java: 11 }
           - { opensearch_version: 2.2.1, java: 11 }
           - { opensearch_version: 2.3.0, java: 11 }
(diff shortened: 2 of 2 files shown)


COMPATIBILITY.md
```diff
@@ -6,10 +6,10 @@
 The below matrix shows the compatibility of the [`opensearch-java-client`](https://search.maven.org/artifact/org.opensearch.client/opensearch-java) with versions of [`OpenSearch`](https://opensearch.org/downloads.html#opensearch).
 
 | Client Version | OpenSearch Version |
-| --- | --- |
-| 1.0.0 | 1.0.0-1.3.3 |
-| 2.0.0 | 1.3.3-2.0.1 |
-| 2.1.0 | 1.3.3-2.3.0 |
```
reta, line 12: Could you please add current version as well? `| 2.2.0 | 1.3.3-2.3.0 |`

AmiStrn:
Thanks @reta! 
I was just adding the changelog update - should I not commit that?

reta:
> Thanks @reta! I was just adding the changelog update - should I not commit that?

Thanks @AmiStrn,  I think we could skip changelog - no functional changes here

COMPATIBILITY.md
AmiStrn: sure

AmiStrn: just to make sure i got this. 
i am making the table like this? because if the current version is 2.2.0 i guess i am adding a new line and stating that it is compatible with these versions.

```
| Client Version | OpenSearch Version |
|----------------|--------------------|
| 1.0.0          | 1.0.0-1.3.3        |
| 2.0.0          | 1.3.3-2.0.1        |
| 2.1.0          | 1.3.3-2.3.0        |
| 2.2.0          | 1.3.3-2.5.0        |
```

AmiStrn: this is confusing since there is nothing to backport really. I think I am missing something here

reta: Table looks good, thanks! I didn't get the `backport` part, what is it about?

AmiStrn: never mind. was wondering how i can see that the test is of version 2.2.0 but that is clear to me now. 
I thought this was an issue since there are failing tests. then i saw all PR's failing for unpublished opensearch versions... so nothing special here.

reta (approved):

COMPATIBILITY.md
```diff
@@ -6,10 +6,11 @@
...
+|----------------|--------------------|
+| 1.0.0          | 1.0.0-1.3.3        |
+| 2.0.0          | 1.3.3-2.0.1        |
+| 2.1.0          | 1.3.3-2.3.0        |
+| 2.2.0          | 1.3.3-2.5.0        |
```
andrross, line 13: What do you think about claiming `1.3.3-2.x` for the latest client version? We are running the integration tests against the [latest unreleased version](https://github.com/opensearch-project/opensearch-java/blob/819cb7cde7cc976bb504e07bad1cbc1629ded26e/.github/workflows/test-integration-unreleased.yml#L19-L20).

andrross: (client 2.0.0 and 2.1.0 should continue to work against all 2.x server versions as well assuming we do semver correctly)

saratvemulapalli: +1

saratvemulapalli (approved):

COMPATIBILITY.md
reta: @andrross it is a bit far reaching I think, the exact number gives more clarify regarding current OpenSearch releases but I am not sure (we could also put 3.x here since we run against unreleased 3.x as well)

AmiStrn: @andrross it is a question of being explicit or not. We don't use this syntax in other client repo's so it would be different here. i prefer being explicit as well regarding the versions tested using the integrations tests.

AmiStrn:
Do we need 6 reviews for every change in the client?

AmiStrn:
Thanks @reta @andrross and @reta for reviewing :)

andrross (approved):

COMPATIBILITY.md
AmiStrn: @andrross This is a question perhaps to raise in https://github.com/opensearch-project/opensearch-clients because it may be relevant to many clients

Issue #379, opened 2023-02-28
[FEATURE] java api doesn't allow setting `index.mapping.total_fields.limit`


L00kian:
### Is your feature request related to a problem?
Current Java API doesn't contain definition of 'mapping' object within org.opensearch.client.opensearch.indices.IndexSettings and doesn't allow a non-declared json field to be added to the payload

### What solution would you like?
Either add `mapping.totalFields.limit` properties chain to IndexSettings object or consider allowing setting an arbitrary json property on any level of any API request objects

### What alternatives have you considered?
Using high-level rest client as a workaround

### Do you have any additional context?
Nope

dblock:
Looks right, want to contribute @L00kian ?

L00kian:
Hi @dblock, which option would be the way to go? arbitrary fields support on all the levels or just adding the field?

Can and will contribute :)

dblock:
Both! I think that arbitrary fields support on all the levels makes some sense because it will provide a workaround quickly, but that also means that one could theoretically add fields that don't exist. I would fix any missing field as 1 PR and then independently propose an arbitrary field solution and we can discuss that on its own merits. WDYT?

wbeckler:
I'm renaming this issue to refer to the remaining proposal.

PR #382, opened 2023-03-01
Add "mappings" field to IndexSettings object

### LINKED ISSUE #171: [FEATURE] Be able to set total number of fields on index
I cannot using the java client set index.mappings.total_fields.limit on an IndexSettings request, and have to do so manually since my project uses the java client instead of the REST client.

I would like to be able to set the total fields limit for an index, same as number of shards, on the IndexSettings object for an IndexTemplateMapping.

Using the REST client in addition to the java client, but I really don't want to have to use both clients in my project just so I can set the total fields on an index mapping update.

L00kian:
### Description
Allows to manage index mapping settings using java client

### Issues Resolved
Closes #171 
Relates to #379 

### DIFF
--- a/java-client/src/main/java/org/opensearch/client/opensearch/indices/IndexSettingsMapping.java
+++ b/java-client/src/main/java/org/opensearch/client/opensearch/indices/IndexSettingsMapping.java
@@ -1,0 +1,262 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.client.opensearch.indices;
+
+import jakarta.json.stream.JsonGenerator;
+import org.opensearch.client.json.JsonpDeserializable;
+import org.opensearch.client.json.JsonpDeserializer;
+import org.opensearch.client.json.JsonpMapper;
+import org.opensearch.client.json.JsonpSerializable;
+import org.opensearch.client.json.ObjectBuilderDeserializer;
+import org.opensearch.client.json.ObjectDeserializer;
+import org.opensearch.client.opensearch._types.Time;
+import org.opensearch.client.util.ObjectBuilder;
+import org.opensearch.client.util.ObjectBuilderBase;
+
+import javax.annotation.Nullable;
+import java.util.List;
+import java.util.function.Function;
+
+
+@JsonpDeserializable
+public class IndexSettingsMapping implements JsonpSerializable {
+
+	@Nullable
+	private final IndexSettingsMappingLimit totalFields;
+
+	@Nullable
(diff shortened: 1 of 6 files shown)


dblock:
Needs a CHANGELOG entry and otherwise LGTM. @VachaShah does this look right to you?

I noticed some code that says it was auto-generated. Was that something during build? I'm not super familiar with this stuff here.

java-client/src/main/java/org/opensearch/client/opensearch/indices/IndexSettingsMapping.java
```diff
@@ -0,0 +1,331 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+//----------------------------------------------------
```
VachaShah, line 33: Since this is a new file (and the other one too in this PR) being added, the license headers would just be for OpenSearch.

Also, this code is not generated so we can remove that line as well.

VachaShah:
> Needs a CHANGELOG entry and otherwise LGTM. @VachaShah does this look right to you?
> 
> I noticed some code that says it was auto-generated. Was that something during build? I'm not super familiar with this stuff here.

Yeah the code is not auto-generated, so we can remove that line,

java-client/src/test/java/org/opensearch/client/opensearch/experiments/ParsingTests.java
```diff
@@ -85,4 +86,44 @@ public void testFoo() throws Exception {
       throw new JsonParsingException(je.getMessage() + " at " + je.getLocation(), je, je.getLocation());
     }
   }
+
+  @Test
+  public void testIndexSettingsMappingParsing() {
```
VachaShah, line 91: Can we also add a test for index settings mapping during index creation ([here](https://github.com/opensearch-project/opensearch-java/blob/main/java-client/src/test/java/org/opensearch/client/opensearch/integTest/AbstractRequestIT.java#L114))?

java-client/src/main/java/org/opensearch/client/opensearch/indices/IndexSettingsMapping.java
L00kian: Will do the cleanup.

VachaShah:
@L00kian Thank you for addressing the comments! The checkstyle is failing due to the below error:

```
Error: eckstyle] [ERROR] /home/runner/work/opensearch-java/opensearch-java/java-client/src/test/java/org/opensearch/client/opensearch/integTest/AbstractRequestIT.java:66:8: Unused import - org.opensearch.client.opensearch.indices.IndexSettingsMapping. [UnusedImports]
```

Can you fix this?

dblock (approved):

Issue #388, opened 2023-03-06
[FEATURE] Add capability to cat indices with column names


MarinaRazumovsky:
### Is your feature request related to a problem?
Cat indices command does not return creation_date of indices in standard mode

### What solution would you like?
Option to choose columns

### What alternatives have you considered?
use org.opensearch.client.RestClient

wbeckler:
Would you be willing to add this?

AmiStrn:
I think we can close this one now :)

PR #391, opened 2023-03-06
Add option to set result columns into cat indices request

MarinaRazumovsky:
### Description
Add option to set result columns into cat indices request

### Issues Resolved
[#388] 

### DIFF
--- a/.github/workflows/publish-snapshots.yml
+++ b/.github/workflows/publish-snapshots.yml
@@ -27,4 +27,6 @@ jobs:
         run: |
           export SONATYPE_USERNAME=$(aws secretsmanager get-secret-value --secret-id maven-snapshots-username --query SecretString --output text)
           export SONATYPE_PASSWORD=$(aws secretsmanager get-secret-value --secret-id maven-snapshots-password --query SecretString --output text)
+          echo "::add-mask::$SONATYPE_USERNAME"
+          echo "::add-mask::$SONATYPE_PASSWORD"
           ./gradlew --no-daemon publishPublishMavenPublicationToSnapshotsRepository


AmiStrn (dismissed):
Thanks @MarinaRazumovsky ! looks pretty straightforward, I am surprised this was a missing capability

reta:
Thanks for addressing the issue, @MarinaRazumovsky , although your are looking into one specific request, the problem could be addressed at large:
 - list of headers is common for all `_cat` actions [1], I think it would make sense to move it to `CatRequestBase` 
 - we definitely need to add test cases, the number of issues arise from failed deserialization attempts, we have to be sure header values are properly extracted from the responses (there are surprises there)

Thank you.

[1] https://opensearch.org/docs/1.2/opensearch/rest-api/cat/index/#common-url-parameters

MarinaRazumovsky:
@reta thank you for your response. What is a right location for implementation of `_cat` tests?

MarinaRazumovsky:
Hi @reta. I started to implement common cat parameters in `CatRequestBase`.  Could you review current changes, if it looks good, I'll implement rest cat requests.

java-client/src/main/java/org/opensearch/client/opensearch/cat/CatRequestBase.java
```diff
@@ -37,21 +37,60 @@
...
 public abstract class CatRequestBase extends RequestBase {
-	public CatRequestBase() {
-	}
 
-	protected abstract static class AbstractBuilder<BuilderT extends AbstractBuilder<BuilderT>>
-			extends
-				ObjectBuilderBase {
-		protected abstract BuilderT self();
+    @Nullable
+    private final String columnNames;
```
reta, line 53: Super minor, I think the APi refers to this property as `headers`, could we change it please?

java-client/src/main/java/org/opensearch/client/opensearch/cat/IndicesRequest.java
```diff
@@ -44,310 +44,312 @@
...
+                    SimpleEndpoint.pathEncode(request.index.stream().map(v -> v).collect(Collectors.joining(",")), buf);
+                    return buf.toString();
+                }
+                throw SimpleEndpoint.noPathTemplateFound("path");
+
+            },
+
+            // Request parameters
+            request -> {
+                Map<String, String> params = request.queryParameters();
```
reta, line 333: new HashMap<>(request.queryParameters())
```suggestion
                Map<String, String> params = new HashMap<>(request.queryParameters());
```

Let's not modify the original request

java-client/src/main/java/org/opensearch/client/opensearch/cat/CatRequestBase.java
MarinaRazumovsky: @reta are you sure that it will not ambiguity with request headers?

java-client/src/main/java/org/opensearch/client/opensearch/cat/IndicesRequest.java
MarinaRazumovsky: ok, thanks

java-client/src/main/java/org/opensearch/client/opensearch/cat/CatRequestBase.java
reta: I don't see request headers in `RequestBase` (afaik we do not use model classes to pass headers through), which request headers you are referring to?

AmiStrn: Perhaps the headers in question are the ones sent with the clients request? E.g "x-opaque-id" for example? @MarinaRazumovsky i am not sure if that's what you meant but I was going to make a similar comment. Correct me if i am wrong in my assumption.

reta: Thanks @AmiStrn , if that is the case, I think out of the context the `headers` could be misleading, but the context is `_cat` API [1] where such headers are related to the output table view. I am certainly fine with better name (if `headers` are confusing)  to align it with `_cat` APIs 

[1] https://opensearch.org/docs/2.6/api-reference/cat/index/

AmiStrn: makes sense, @MarinaRazumovsky wdyt? headers is the way it is described in the docs, and internal code as well.

MarinaRazumovsky: you are right, I rename this parameter

reta:
@MarinaRazumovsky LGTM, thank you! Some checkstyle violations:

```
Task :java-client:checkstyleTest FAILED
```

java-client/src/test/java/org/opensearch/client/opensearch/integTest/AbstractCatClientIT.java
```diff
@@ -0,0 +1,203 @@
...
+        assertNotNull(allocationRecord.diskIndices());
+        assertNotNull(allocationRecord.diskUsed());
+        assertNotNull(allocationRecord.diskAvail());
+        assertNotNull(allocationRecord.diskTotal());
+        assertNotNull(allocationRecord.diskPercent());
+        assertNull(allocationRecord.host());
+    }
+
+    public void testCatPlugins() throws Exception {
+        PluginsResponse pluginsResponse = javaClient().cat().plugins();
```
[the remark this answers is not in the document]
MarinaRazumovsky: thank you, fixed it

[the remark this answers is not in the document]
reta: Oops sorry, the cause is not the same, looking into it

java-client/src/test/java/org/opensearch/client/opensearch/integTest/AbstractCatClientIT.java
```diff
@@ -0,0 +1,206 @@
...
+
+    public void testCatPlugins() throws Exception {
+        assumeTrue("There are no plugins installed when HTTP is used", isHttps());
+        PluginsResponse pluginsResponse = javaClient().cat().plugins();
+        assertNotNull(pluginsResponse.valueBody());
+        assertTrue(pluginsResponse.valueBody().size() > 0);
+    }
+
+    public void testCatSegments() throws Exception {
+        createIndex("cat-segments-test-index");
```
reta, line 184:
```suggestion
        createIndex("cat-segments-test-index");
        
        final IndexResponse index = javaClient().index(b -> b
            .index("test-cat-segments-index")
            .id("id")
            .refresh(Refresh.True)
            .document(Map.of()));
        assertTrue(index.result() == Result.Created);
```

reta: We need to add at lest one document to be 100% there would be at least one segment

PR #477, opened 2023-05-11
Fix suggestion classes for Term, Phrase, and Completion.

### LINKED ISSUE #459: [BUG] Request failed: [parse_exception] missing suggestion object
### What is the bug?
When I send a  Suggestionrequest the response cannot be parsed because of the following exception:

```
 org.opensearch.client.opensearch._types.OpenSearchException: Request failed: [parse_exception] missing suggestion object
	at org.opensearch.client.transport.aws.AwsSdk2Transport.parseResponse(AwsSdk2Transport.java:486)
	at org.opensearch.client.transport.aws.AwsSdk2Transport.executeSync(AwsSdk2Transport.java:393)
	at org.opensearch.client.transport.aws.AwsSdk2Transport.performRequest(AwsSdk2Transport.java:191)
	at org.opensearch.client.opensearch.OpenSearchClient.search(OpenSearchClient.java:1219)
	at com.heiland.searchservice.repository.AbstractOpenSearchRepository.autocomplete(AbstractOpenSearchRepository.java:48)
	... 176 common frames omitted
```

My implementation is as follows:

````
try {

  final CompletionSuggester completion = CompletionSuggester.of(c -> c.field("suggestions"));
  final FieldSuggester fieldSuggester = FieldSuggester.of(f -> ((FieldSuggester.Builder) f.completion(completion)).prefix("Haut"));
  final SearchRequest request = SearchRequest.of(s -> s.index(searchIndex()).suggest(Suggester.of(s -> s.suggesters("autocomplete", fieldSuggester))));
  
  return client.search(request, Object.class);

  } catch (final Exception e) {
    throw new RuntimeException(e);
  }
````

This implementation shall represent the following query:

`GET index/_search
{
  "suggest": {
    "autocomplete": {
      "prefix": "Haut",
      "completion": {
        "field": "suggestions"
      }
    }
  }
}`

I expect that the reponse can be parsed and the SearchRepsonse has the field suggest is filled propertly.

* SpringBoot 3
* opensearch-java v2.4.0
* OpenSearch Domain v2.5

### PR DESCRIPTION
**Pull request by harshavamsi**
### Description
This change refactors the suggestion classes for Term, Phrase, and Completion suggestion.

The previous logic used the`UnionDeserializer` class to deserialize the responses from these API calls. But due to a bug in the way the `UnionDeserializer` treats member object, phrase completion was broken. This PR uses an alternative approach with `ExternallyTaggedUnionDeserializer`. It is correctly able to differentiate between the three entities. 

Further, also refactored the Suggester classes for the three entities. The API is now more verbose and has the correct calling order. The previous was of accepting the options provided by the response was `response.suggest().get(suggesterName).get(0).options().get(0).phrase().text`. It is now `response.suggest().get(suggesterName).get(0).phrase().options().get(0).text`, which is more logically.

### Issues Resolved
Closes #459

### CODE REVIEW

File: USER_GUIDE.md [hunk 1]
```diff
@@ -169,6 +174,236 @@ for (int i = 0; i < searchResponse.hits().hits().size(); i++) {
...
+                "\"filter\": {\n" +
+                "\"shingle\": {\n" +
+                "\"type\": \"shingle\",\n" +
+                "\"min_shingle_size\": 2,\n" +
+                "\"max_shingle_size\": 3\n" +
+                "}\n" +
+                "}\n" +
+                "}\n" +
+                "}\n" +
+                "}";
```
dblock, line 336: This is pretty hard to read. Use a multiline Java string?

dblock:
Looks like a breaking change and will need to be in a major version increment?

USER_GUIDE.md
harshavamsi: Did one better in the actual tests to use resource files, but updated the example here with a text block.

harshavamsi:
> Looks like a breaking change and will need to be in a major version increment?

I don't think this would classify as a breaking change, this is a bug fix for broken feature, see #459. The refactoring of the core logic around the way functions are used is only because it was incorrect in the first place. But looks like there are no folks using it already because it was completely broken since fork anyway.

java-client/src/main/java/org/opensearch/client/opensearch/core/search/PhraseSuggestOption.java
```diff
@@ -83,6 +86,7 @@ public final String text() {
 	/**
 	 * Required - API name: {@code highlighted}
```
dblock, line 87: Now nullable.

java-client/src/main/java/org/opensearch/client/opensearch/core/search/TermSuggest.java
```diff
@@ -0,0 +1,102 @@
...
+import org.opensearch.client.json.ObjectBuilderDeserializer;
+import org.opensearch.client.json.ObjectDeserializer;
+import org.opensearch.client.opensearch.core.search.Suggest.Kind;
+import org.opensearch.client.util.ApiTypeHelper;
+import org.opensearch.client.util.ObjectBuilder;
+
+import jakarta.json.stream.JsonGenerator;
+
+@JsonpDeserializable
+public class TermSuggest extends SuggestBase implements SuggestVariant {
```
dblock, line 26: Should this be SuggestTerm?

dblock (commented):
Looks good. Found some nits. 

I think you should write unit tests for the new classes, even if it's very basic stuff.

java-client/src/main/java/org/opensearch/client/opensearch/core/search/TermSuggest.java
harshavamsi: I'm using the naming conventions from OpenSearch which uses `CompletionSuggest`, `TermSuggest`, and `PhraseSuggest`.

harshavamsi:
> 

For the purposes of escalating customer requests, I think that we should get this change in and release in 2.5.0. I think we have some confidence given that the integration tests are passing. I'm going to open up a new issue to add unit tests to the classes we've added(I don't believe there are unit tests for any API classes today).

java-client/src/main/java/org/opensearch/client/opensearch/core/search/CompletionSuggest.java
```diff
@@ -0,0 +1,137 @@
...
+import org.opensearch.client.json.ObjectBuilderDeserializer;
+import org.opensearch.client.json.ObjectDeserializer;
+import org.opensearch.client.opensearch.core.search.Suggest.Kind;
+import org.opensearch.client.util.ApiTypeHelper;
+import org.opensearch.client.util.ObjectBuilder;
+
+import jakarta.json.stream.JsonGenerator;
+
+@JsonpDeserializable
+public class CompletionSuggest<TDocument> extends SuggestBase implements SuggestVariant {
```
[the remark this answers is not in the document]
harshavamsi: @reta these are left over issues post fork from ES, do we need to include the license headers if we're re-adding classes and logic that was missing?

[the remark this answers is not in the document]
reta: @harshavamsi I mean this left overs - did you (we) wrote them from scratch (no license header updates needed)? did we inherited them from Elasticsearch codebase (we need to keep the original license header)? Did we copied them from Elasticsearch codebase (we need to keep the original license header)?

USER_GUIDE.md
```diff
@@ -169,6 +174,243 @@ for (int i = 0; i < searchResponse.hits().hits().size(); i++) {
...
+          "analyzer": "trigram"
+        }
+      }
+    }
+  }
+}
+""";
+
+
+    ObjectMapper mapper = new JsonMapper();
```
VachaShah, line 359: The formatting in this part is a little off. Can you fix that?

java-client/src/main/java/org/opensearch/client/opensearch/core/search/CompletionSuggest.java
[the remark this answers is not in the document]
VachaShah: +1.

.github/CODEOWNERS
```diff
@@ -1,2 +1,2 @@
 # This should match the team set up in https://github.com/orgs/opensearch-project/teams and include any additional contributors
-*   @reta @Bukhtawar @dblock @madhusudhankonda @saratvemulapalli @VachaShah
\ No newline at end of file
+*   @reta @Bukhtawar @dblock @szczepanczykd @madhusudhankonda @saratvemulapalli @VachaShah
```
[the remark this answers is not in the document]
harshavamsi: My bad, pulled in from main by accident. Fixed now.

java-client/src/main/java/org/opensearch/client/opensearch/core/search/CompletionSuggest.java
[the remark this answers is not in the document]
harshavamsi: @reta @VachaShah added in ES license for files that were copied over.

USER_GUIDE.md
```diff
@@ -169,6 +174,243 @@ for (int i = 0; i < searchResponse.hits().hits().size(); i++) {
...
+    }
+  }
+}
+""";
+
+
+ObjectMapper mapper = new JsonMapper();
+JsonpMapper jsonpMapper = new JacksonJsonpMapper(mapper);
+IndexSettings settings;
+try (JsonParser settingsParser = new JacksonJsonpParser(mapper.createParser(settingsJson))) {
```
reta, line 362: Do you think we could replace that with `IndexSettings.Builder` and `TypeMapping.Builder`? We should not encourage using raw `Xxx._DESERIALIZER`s in the official guides I think.

java-client/src/test/java/org/opensearch/client/opensearch/integTest/AbstractRequestIT.java
```diff
@@ -652,6 +671,210 @@ public void testPit() throws IOException {
...
+				JsonNode jsonNode = mapper.readValue(in, JsonNode.class);
+				mappingJson = mapper.writeValueAsString(jsonNode);
+			} catch (Exception e) {
+				throw new RuntimeException(e);
+			}
+
+			ObjectMapper mapper = new JsonMapper();
+			JsonpMapper jsonpMapper = new JacksonJsonpMapper(mapper);
+			IndexSettings settings;
+			try (JsonParser settingsParser = new JacksonJsonpParser(mapper.createParser(settingsJson))) {
```
reta, line 818: Same comment as for guides, would be great to see builders here instead

harshavamsi:
@reta I can fix the suggestions on another PR, for the time being can we merge this?

harshavamsi:
@dblock @VachaShah need approvals

reta:
> @reta I can fix the suggestions on another PR, for the time being can we merge this?

Why?

harshavamsi:
https://github.com/opensearch-project/opensearch-java/issues/459 -- seems like some folks really want this fix

reta:
> #459 -- seems like some folks really want this fix

I can help you address the comments, the documentation is the way we recommend our users to use the Java client, it should showcase the right APIs

harshavamsi:
@reta updated the example and tests to use builders instead

dblock:
> #459 -- seems like some folks really want this fix

I'm glad we're doing something users want! Let's make sure to address the must have's :) @reta all yours for CR!

java-client/src/test/resources/phraseIndexMappings.json
reta, line 1: Not needed anymore, please remove

java-client/src/test/resources/phraseIndexSettings.json
reta, line 1: Not needed anymore, please remove

java-client/src/test/resources/phraseIndexMappings.json
harshavamsi: Removed.

java-client/src/test/resources/phraseIndexSettings.json
harshavamsi: Removed.

reta (approved):

VachaShah (approved):

reta:
@harshavamsi could you please manually backport to `2.x`? thank you

iamdanhart:
Just as an anecdotal data point, this breaks one of my usages of the client library around completion suggestion (works as intended currently on version 2.4.0). It should be a relatively light lift for me to fix, but not something I expected in a minor release bump.

dblock:
> Just as an anecdotal data point, this breaks one of my usages of the client library around completion suggestion (works as intended currently on version 2.4.0). It should be a relatively light lift for me to fix, but not something I expected in a minor release bump.

Yep, shouldn't have per semver. What's your before/after example? You should open a bug for it.

Issue #184, opened 2022-07-20
[BUG] Java client: Jackson exception: Numeric value out of range of int


trtrmitya:
### What is the bug?
_types.StoreStats fields are ```int``` and are easily overflow.

### How can one reproduce the bug?
Similar problem was reported to Elastic team: https://discuss.elastic.co/t/java-client-jackson-exception-numeric-value-out-of-range-of-int/

```
ElasticsearchIndicesClient indicesClient = new ElasticsearchIndicesClient(transport);
indicesClient.stats(b -> b.index("name"));
```

result in 
```
Exception in thread "main" co.elastic.clients.json.JsonpMappingException: Error deserializing co.elastic.clients.elasticsearch._types.StoreStats: jakarta.json.stream.JsonParsingException: Jackson exception: Numeric value (5104343339) out of range of int
 at [Source: (org.apache.http.nio.entity.ContentInputStream); line: 1, column: 176] (JSON path: _all.primaries.store.size_in_bytes) (line no=1, column no=176, offset=-1)
	at co.elastic.clients.json.JsonpMappingException.from0(JsonpMappingException.java:134)
	at co.elastic.clients.json.JsonpMappingException.from(JsonpMappingException.java:121)
	at co.elastic.clients.json.ObjectDeserializer.deserialize(ObjectDeserializer.java:206)
	at co.elastic.clients.json.ObjectDeserializer.deserialize(ObjectDeserializer.java:136)
	at co.elastic.clients.json.JsonpDeserializer.deserialize(JsonpDeserializer.java:75)
	at co.elastic.clients.json.ObjectBuilderDeserializer.deserialize(ObjectBuilderDeserializer.java:79)
	at co.elastic.clients.json.DelegatingDeserializer$SameType.deserialize(DelegatingDeserializer.java:43)
	at co.elastic.clients.json.ObjectDeserializer$FieldObjectDeserializer.deserialize(ObjectDeserializer.java:71)
	at co.elastic.clients.json.ObjectDeserializer.deserialize(ObjectDeserializer.java:180)
	at co.elastic.clients.json.ObjectDeserializer.deserialize(ObjectDeserializer.java:136)
	at co.elastic.clients.json.JsonpDeserializer.deserialize(JsonpDeserializer.java:75)
	at co.elastic.clients.json.ObjectBuilderDeserializer.deserialize(ObjectBuilderDeserializer.java:79)
[...]

PR #489, opened 2023-05-20
fix: change int type to long

szczepanczykd:
### Description
The server returns long instead of int:
https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/store/StoreStats.java

### Issues Resolved
https://github.com/opensearch-project/opensearch-java/issues/184
https://github.com/opensearch-project/opensearch-java/issues/490

### DIFF
--- a/java-client/src/main/java/org/opensearch/client/opensearch/_types/StoreStats.java
+++ b/java-client/src/main/java/org/opensearch/client/opensearch/_types/StoreStats.java
@@ -68,7 +68,7 @@ public class StoreStats implements JsonpSerializable {
 	private final String totalDataSetSize;
 
 	@Nullable
-	private final Integer totalDataSetSizeInBytes;
+	private final Long totalDataSetSizeInBytes;
 
 	// ---------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


java-client/src/main/java/org/opensearch/client/opensearch/_types/StoreStats.java
```diff
@@ -57,18 +57,18 @@ public class StoreStats implements JsonpSerializable {
...
 
-	private final int reservedInBytes;
+	private final long reservedInBytes;
 
 	@Nullable
 	private final String totalDataSetSize;
 
 	@Nullable
-	private final Integer totalDataSetSizeInBytes;
+	private final Long totalDataSetSizeInBytes;
```
szczepanczykd, line 71: I don't see this field in the response from API in version 2.7.0
Seems like it was never produced by the server:
https://github.com/opensearch-project/OpenSearch/blob/1.0/server/src/main/java/org/opensearch/index/store/StoreStats.java

Should I remove it or mark it as @Deprecated?

dblock: Mark it as deprecated, will remove in the next major version.

dblock (approved):
Thanks! Make a separate PR for the deprecated field so we have a single link for that one.

VachaShah (approved):

java-client/src/main/java/org/opensearch/client/opensearch/_types/StoreStats.java
szczepanczykd: Done in: https://github.com/opensearch-project/opensearch-java/issues/497 :)

Issue #473, opened 2023-05-09
[BUG] Unhelpful error message when failing to authenticate with Amazon OpenSearch Service


chenqi0805:
### What is the bug?
This is related to our opensearch sink use case with Amazon OpenSearch backend sts role authentication: https://github.com/opensearch-project/data-prepper/issues/2655

### How can one reproduce the bug?
When sts role is not configured with the right permissions, one gets
```
[security_exception] authentication/authorization failure
```

### What is the expected behavior?
Ideally the [cause message](https://github.com/opensearch-project/opensearch-java/blob/af8d96fc34373542b2332efa3b258f754de33de6/java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java#L453) could be

1. more specific about the operation that fails authentication
2. print the sts role used in authentication

### What is your host/environment?
all

### Do you have any screenshots?
_If applicable, add screenshots to help explain your problem._

### Do you have any additional context?
_Add any other context about the problem._

dblock:
What's the actual HTTP response for this?

Looking at [the code](https://github.com/opensearch-project/opensearch-java/blob/e422eb0263847bd1ca867c8c109bbb15463fae5b/java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java#L452-L453) it does try to parse the response and pull out the complete message.

chenqi0805:
@dblock Sorry for the late response. I also had a hard time to reproduce it. But it is with getClusterSettings:
```
org.opensearch.client.opensearch._types.OpenSearchException: Request failed: [security_exception] authentication/authorization failure
	at org.opensearch.client.transport.aws.AwsSdk2Transport.parseResponse(AwsSdk2Transport.java:517) ~[opensearch-java-2.2.0.jar:?]
	at org.opensearch.client.transport.aws.AwsSdk2Transport.executeSync(AwsSdk2Transport.java:438) ~[opensearch-java-2.2.0.jar:?]
	at org.opensearch.client.transport.aws.AwsSdk2Transport.performRequest(AwsSdk2Transport.java:241) ~[opensearch-java-2.2.0.jar:?]
	at org.opensearch.client.opensearch.cluster.OpenSearchClusterClient.getSettings(OpenSearchClusterClient.java:282) ~[opensearch-java-2.2.0.jar:?]
[...]

PR #493, opened 2023-05-22
FIX: extract reason from error in AWSSdk2Transport

chenqi0805:
### Description
This is for resolving missing cause message when there is no proper permission to access Fine grained Access Control Amazon Opensearch domain.
```
Exception in thread "main" org.opensearch.client.opensearch._types.OpenSearchException: Request failed: [security_exception] authentication/authorization failure
	at org.opensearch.client.transport.aws.AwsSdk2Transport.parseResponse(AwsSdk2Transport.java:517)
	at org.opensearch.client.transport.aws.AwsSdk2Transport.executeSync(AwsSdk2Transport.java:438)
	at org.opensearch.client.transport.aws.AwsSdk2Transport.performRequest(AwsSdk2Transport.java:241)
	at org.opensearch.client.opensearch.OpenSearchClient.bulk(OpenSearchClient.java:211)
	at org.opensearch.dataprepper.plugins.sink.opensearch.index.TestClient.testBulk(TestClient.java:215)
	at org.opensearch.dataprepper.plugins.sink.opensearch.index.TestClient.main(TestClient.java:80)
```

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-java/issues/473


wbeckler:
Thank you for submitting this! Can you think of a scenario in which an application might be relying on the previous behavior? As in, could this possibly be a breaking change?

chenqi0805:
@wbeckler The only scenario I can think of is when both `reason` and `message` is present but so far it is hypothetical as I have not found such case. Do you have other scenario in your head?

wbeckler:
> @wbeckler The only scenario I can think of is when both `reason` and `message` is present but so far it is hypothetical as I have not found such case. Do you have other scenario in your head?

Not really. What you're saying makes sense, but I ask the question in case there's something I'm not thinking of.

VachaShah (commented):
LGTM! Just a minor comment.

VachaShah (approved):

dblock (changes requested):
Looks good, but needs tests please.

chenqi0805:
@dblock Thanks for the suggestion. I do not find a good strategy to unit test so I put a AwsSdk2SecurityIT that needs to be run against a FGAC without proper permission.

dblock:
Without looking at the code, do we not have a way to mock a response? If it's really hard, care to open an issue to add tests that exercise all these paths and we'll merge as is? Also @reta wdyt?

reta:
> Without looking at the code, do we not have a way to mock a response? If it's really hard, care to open an issue to add tests that exercise all these paths and we'll merge as is? Also @reta wdyt?

Sounds reasonable @dblock , we may need to come up with some mocks (like test fixtures we do in core) but it may take some efforts.

wbeckler:
@chenqi0805 opened a separate issue for test coverage of transport: https://github.com/opensearch-project/opensearch-java/issues/503

reta:
LGMT! @chenqi0805 there are conflicts, could you please resolve them? thank you

chenqi0805:
@reta resolved conflict

reta (approved):

reta:
@dblock LGTY? thanks!

VachaShah (approved):

chenqi0805:
@dblock Need your approval for unblocking merge

dblock (approved):

PR #518, opened 2023-06-06
fix: add missing translog sync interval option to index settings

### LINKED ISSUE #307: [BUG] Missing Index Setting Options
### What is the bug?
`org.opensearch.client.opensearch.indices.IndexSettings` does not support all configuration options offered by the server. For instance, there is no way to set `translog.sync_interval`.

If one tries to construct the index settings object via its builder, there is the setter `translogDurability(String)` for `translog.durability`, but no corresponding setter for `translog.sync_interval`. (When I check the settings of an existing index with request parameter `include_defaults=true`, then I see even more attributes not available in `IndexSettings`.)

Since many developers are likely to read the index settings and type mappings from a file, the following test seems relevant:
~~~java
    @Test
    public void createIndexSettingsFromJson() throws IOException {
        ObjectMapper mapper = new JsonMapper();
        JsonpMapper jsonpMapper = new JacksonJsonpMapper(mapper);

        String settingsJson = "{\"index\": {\"translog.durability\": \"async\", \"translog.sync_interval\": \"60s\"}}";
        IndexSettings settings;
        try (JsonParser settingsParser = new JacksonJsonpParser(mapper.createParser(settingsJson))) {
            settings = IndexSettings._DESERIALIZER.deserialize(settingsParser, jsonpMapper);
        }
        
        StringWriter writer = new StringWriter();
        try (JsonGenerator jsonGenerator = Json.createGenerator(writer)) {
            settings.serialize(jsonGenerator, jsonpMapper);
        }
        String roundTripJson = writer.toString();

        assertThat(roundTripJson, Matchers.containsString("durability"));
        assertThat(roundTripJson, Matchers.containsString("async"));
        assertThat(roundTripJson, Matchers.containsString("sync_interval"));
    }
~~~
The test fails with the message:
~~~text
Expected: a string containing "sync_interval"
     but: was "{"index":{"translog.durability":"async"}}"
~~~
The attribute `translog.sync_interval` has been lost.

As an aside: I am surprised the deserializer only recognizes the string-valued attribute `translog.durability`, not an object-valued attribute `translog` whose value then contains an attribute `durability`.

All index configuration options should be available through the Java client.

OpenSearch 2.4.1, Java client 2.1.0

I am migrating some code from Elasticsearch 7.10 + Rest high-level client to OpenSearch 2.x + Java client. The index in question will be filled in a long running bulk update. If the ingest of the document corpus fails for some reason, then the index is discarded anyway. Therefore, I prefer to set `translog.durability` to `async` and `translog.sync_interval` to a value in the minute range.

Mstrutov:
### Description
Now we can set translog sync interval option via client.

Additionally, server response contains separate `translog` object, which cannot be deserialized by present translog option deserializers.
It looks like this:
```
{
  "test-settings": {
    "settings": {
      "index": {
        "number_of_shards": "1",
        "translog": {
          "sync_interval": "10s",
          "durability": "async"
        },
        "provided_name": "test-settings",
        "creation_date": "1685967681952",
        "number_of_replicas": "1",
        "uuid": "q4R3eDKjT164eR_NNuY8pQ",
        "version": {
          "created": "136257827"
        }
      }
    }
  }
}
```

Therefore, a corresponding `Translog` class is introduced, similarly to how it is done for the `blocks` property.

### Issues Resolved
Closes #307 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Changed
 
 ### Deprecated
+- Deprecate translogDurability and translogFlushThresholdSize in IndexSettings in favor of a separate translog object ([#518](https://github.com/opensearch-project/opensearch-java/pull/518))
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Added
 - Add support for knn_vector field type ([#529](https://github.com/opensearch-project/opensearch-java/pull/524))
+- Add translog option object and missing translog sync interval option in index settings ([#518](https://github.com/opensearch-project/opensearch-java/pull/518))
 
 ### Dependencies
 - Bumps `jackson` from 2.14.2 to 2.15.2 ((#537)[https://github.com/opensearch-project/opensearch-java/pull/537])
--- a/java-client/src/main/java/org/opensearch/client/opensearch/indices/IndexSettings.java
+++ b/java-client/src/main/java/org/opensearch/client/opensearch/indices/IndexSettings.java
@@ -199,10 +199,16 @@ public class IndexSettings implements JsonpSerializable {
 	private final Integer maxSlicesPerScroll;
 
 	@Nullable
+	private final Translog translog;
+
+	@Nullable
 	private final String translogDurability;
 
 	@Nullable
 	private final String translogFlushThresholdSize;
+
+	@Nullable
+	private final Time translogSyncInterval;
 
 	@Nullable
 	private final Boolean queryStringLenient;
@@ -280,8 +286,10 @@ public class IndexSettings implements JsonpSerializable {
 		this.verifiedBeforeClose = builder.verifiedBeforeClose;
 		this.format = builder.format;
 		this.maxSlicesPerScroll = builder.maxSlicesPerScroll;
+		this.translog = builder.translog;
 		this.translogDurability = builder.translogDurability;
 		this.translogFlushThresholdSize = builder.translogFlushThresholdSize;
+		this.translogSyncInterval = builder.translogSyncInterval;
 		this.queryStringLenient = builder.queryStringLenient;
 		this.priority = builder.priority;
 		this.topMetricsMaxSize = builder.topMetricsMaxSize;
@@ -681,6 +689,14 @@ public class IndexSettings implements JsonpSerializable {
 	}
 
 	/**
+	 * API name: {@code translog}
+	 */
+	@Nullable
+	public final Translog translog() {
+		return this.translog;
+	}
+
+	/**
 	 * API name: {@code translog.durability}
 	 */
 	@Nullable
@@ -694,6 +710,14 @@ public class IndexSettings implements JsonpSerializable {
 	@Nullable
 	public final String translogFlushThresholdSize() {
 		return this.translogFlushThresholdSize;
+	}
+
+	/**
+	 * API name: {@code translog.sync_interval}
+	 */
+	@Nullable
+	public final Time translogSyncInterval() {
+		return this.translogSyncInterval;
 	}
 
 	/**
@@ -1016,6 +1040,11 @@ public class IndexSettings implements JsonpSerializable {
 			generator.write(this.maxSlicesPerScroll);
 
 		}
+		if (this.translog != null) {
+			generator.writeKey("translog");
+			this.translog.serialize(generator, mapper);
+
+		}
 		if (this.translogDurability != null) {
 			generator.writeKey("translog.durability");
 			generator.write(this.translogDurability);
@@ -1024,6 +1053,11 @@ public class IndexSettings implements JsonpSerializable {
 		if (this.translogFlushThresholdSize != null) {
 			generator.writeKey("translog.flush_threshold_size");
 			generator.write(this.translogFlushThresholdSize);
+
+		}
+		if (this.translogSyncInterval != null) {
+			generator.writeKey("translog.sync_interval");
+			this.translogSyncInterval.serialize(generator, mapper);
 
 		}
 		if (this.queryStringLenient != null) {
@@ -1221,10 +1255,16 @@ public class IndexSettings implements JsonpSerializable {
 		private Integer maxSlicesPerScroll;
 
 		@Nullable
+		private Translog translog;
+
+		@Nullable
 		private String translogDurability;
 
 		@Nullable
 		private String translogFlushThresholdSize;
+
+		@Nullable
+		private Time translogSyncInterval;
 
 		@Nullable
 		private Boolean queryStringLenient;
@@ -1717,6 +1757,14 @@ public class IndexSettings implements JsonpSerializable {
 		}
 
 		/**
+		 * API name: {@code translog}
+		 */
+		public final Builder translog(@Nullable Translog value) {
+			this.translog = value;
+			return this;
+		}
+
+		/**
 		 * API name: {@code translog.durability}
 		 */
 		public final Builder translogDurability(@Nullable String value) {
@@ -1729,6 +1777,14 @@ public class IndexSettings implements JsonpSerializable {
 		 */
 		public final Builder translogFlushThresholdSize(@Nullable String value) {
 			this.translogFlushThresholdSize = value;
+			return this;
+		}
+
+		/**
+		 * API name: {@code translog.sync_interval}
+		 */
+		public final Builder translogSyncInterval(@Nullable Time value) {
+			this.translogSyncInterval = value;
 			return this;
 		}
 
@@ -1920,10 +1976,13 @@ public class IndexSettings implements JsonpSerializable {
 		op.add(Builder::format, JsonpDeserializer.stringDeserializer(), "format", "index.format");
 		op.add(Builder::maxSlicesPerScroll, JsonpDeserializer.integerDeserializer(), "max_slices_per_scroll",
 				"index.max_slices_per_scroll");
+		op.add(Builder::translog, Translog._DESERIALIZER, "translog", "index.translog");
 		op.add(Builder::translogDurability, JsonpDeserializer.stringDeserializer(), "translog.durability",
 				"index.translog.durability");
 		op.add(Builder::translogFlushThresholdSize, JsonpDeserializer.stringDeserializer(),
 				"translog.flush_threshold_size", "index.translog.flush_threshold_size");
+		op.add(Builder::translogSyncInterval, Time._DESERIALIZER,
+				"translog.sync_interval", "index.translog.sync_interval");
 		op.add(Builder::queryStringLenient, JsonpDeserializer.booleanDeserializer(), "query_string.lenient",
 				"index.query_string.lenient");
 		op.add(Builder::priority, JsonpDeserializer.stringDeserializer(), "priority", "index.priority");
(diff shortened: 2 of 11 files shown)


java-client/src/main/java/org/opensearch/client/opensearch/indices/IndexSettings.java
```diff
@@ -198,12 +198,18 @@ public class IndexSettings implements JsonpSerializable {
 	@Nullable
 	private final Integer maxSlicesPerScroll;
 
+	@Nullable
+	private final Translog translog;
+
 	@Nullable
 	private final String translogDurability;
```
reta, line 205: It seems like the `translogDurability`, `translogFlushThresholdSize` and `translogSyncInterval` should be removed in favor of `translog`?

Mstrutov: Yeah, such a duplication seems ugly enough. Thank you for the feedback!

I have added a second commit, please check it out and verify that I have got the public API aspects right. I wrote the main ideas in the commit message.

Additionally, I think we will have to change serialization from
```
{
    "translog.durability": "async"
}
```
to
```
{
    "translog": {
        "durability": "async"
    }
}
```
(second commit supports both to ensure backward compatibility). As a result, only the latter form will be recognized by the deserializer. How should we mark the former one as deprecated?

java-client/src/main/java/org/opensearch/client/opensearch/indices/IndexSettings.java
```diff
@@ -992,15 +1001,20 @@ protected void serializeInternal(JsonGenerator generator, JsonpMapper mapper) {
...
-			generator.write(this.translogDurability);
+		if (this.translog != null) {
+			generator.writeKey("translog");
+			this.translog.serialize(generator, mapper);
 
-		}
-		if (this.translogFlushThresholdSize != null) {
-			generator.writeKey("translog.flush_threshold_size");
-			generator.write(this.translogFlushThresholdSize);
+			if (this.translog.durability() != null) {
```
reta, line 1008: Sorry why do you need to serialize translog twice? 

```
this.translog.serialize(generator, mapper);
```

should have all fields already included

java-client/src/test/java/org/opensearch/client/opensearch/experiments/ParsingTests.java
```diff
@@ -88,11 +80,27 @@ public void testFoo() throws Exception {
   }
 
   @Test
-  public void testIndexSettingsMappingParsing() {
+  public void testIndexSettingsSyncIntervalTimeParsing() {
```
reta, line 83: Could you please add test case to check the deprecated fields (`translog.flush_threshold_size`) are also serialized? Thank you.

reta (approved):

java-client/src/main/java/org/opensearch/client/opensearch/indices/IndexSettings.java
reta: Thank you!

java-client/src/main/java/org/opensearch/client/opensearch/indices/IndexSettings.java
reta: OK, we could clearly drop the old `translog.xxx` fields, this is the request send to server and it will properly decode the object into the right index settings, we don't need to send it twice

Mstrutov: I agree, we could drop the old `translog.xxx` fields if the serialized JSON is only used inside `opensearch-java:client`, i.e. in a request to server. 
But I am not sure about the client code. It could use some custom parser to deserialize `IndexSettings` JSON into some user-provided custom structure, and it could rely on the fact that translog settings are represented as `translog.xxx` fields. Therefore, getting rid of them in the serialized JSON would break such client code.
I seems unlikely that someone would write custom deserializer, given that each class has a built-in deserializer, and it would probably be easier to use the deserialized class for all the custom operations.

Nevertheless, I can't say with confidence that there is no such "someone", so just being extra careful :)

java-client/src/test/java/org/opensearch/client/opensearch/experiments/ParsingTests.java
Mstrutov: If we decide to keep serializing the `translog.xxx` fields, I will add the test cases.

java-client/src/main/java/org/opensearch/client/opensearch/indices/IndexSettings.java
reta: > But I am not sure about the client code. 

But this is the client code we are changing :) AFAIK since this is the request, out only, we never deserialize it. The only target of this request is OpenSearch server.

Mstrutov: OK I think I am reassured now, I am removing the redundant serialization.

reta (approved):

java-client/src/main/java/org/opensearch/client/opensearch/indices/IndexSettings.java
```diff
@@ -1863,8 +1888,9 @@ protected static void setupIndexSettingsDeserializer(ObjectDeserializer<IndexSet
 		op.add(Builder::format, JsonpDeserializer.stringDeserializer(), "format", "index.format");
 		op.add(Builder::maxSlicesPerScroll, JsonpDeserializer.integerDeserializer(), "max_slices_per_scroll",
 				"index.max_slices_per_scroll");
-		op.add(Builder::translogDurability, JsonpDeserializer.stringDeserializer(), "translog.durability",
-				"index.translog.durability");
+		op.add(Builder::translog, Translog._DESERIALIZER, "translog", "index.translog");
+//		op.add(Builder::translogDurability, JsonpDeserializer.stringDeserializer(), "translog.durability",
+//				"index.translog.durability");
```
dblock, line 1893: Let's not leave comments?

dblock (approved):

java-client/src/test/java/org/opensearch/client/opensearch/experiments/ParsingTests.java
```diff
@@ -88,11 +80,40 @@ public void testFoo() throws Exception {
...
+      assertEquals(indexSettings.translog().flushThresholdSize(), deserialized.translog().flushThresholdSize());
+
+      var deprecatedForm = "{\"translog\":{\"sync_interval\":\"10s\"},\"translog.durability\":\"async\",\"translog" +
+              ".flush_threshold_size\":\"256mb\"}";
+      IndexSettings deprecatedDeserialized = deserialize(deprecatedForm, IndexSettings._DESERIALIZER);
+      assertEquals(indexSettings.translog().syncInterval().time(), deprecatedDeserialized.translog().syncInterval().time());
+      assertEquals(indexSettings.translog().durability(), deprecatedDeserialized.translog().durability());
+      assertEquals(indexSettings.translog().flushThresholdSize(), deprecatedDeserialized.translog().flushThresholdSize());
+
+    } catch (JsonParsingException je) {
```
dblock, line 108: What's the difference between the caught `JsonParsingException` and the re-thrown one? Either way this is a test and it should never hit the `catch` part, so I suggest removing it.

java-client/src/test/java/org/opensearch/client/opensearch/experiments/ParsingTests.java
```diff
@@ -124,4 +136,22 @@ public void testIndexSettingsMappingParsing() {
...
+  private String serialize(JsonpSerializable object) {
+    var baos = new ByteArrayOutputStream();
+    var provider = JsonProvider.provider();
+
+    var generator = provider.createGenerator(baos);
+    object.serialize(generator, new JsonbJsonpMapper());
+    generator.close();
+
+    return baos.toString();
+  }
```
dblock, line 156: We should move this somewhere into a util class and reuse.

dblock (approved):
I have some nice to have's but this is good by me. Up to you if you want to address them.

dblock (approved):

java-client/src/test/java/org/opensearch/client/opensearch/experiments/ParsingTests.java
Mstrutov: Agreed, fixed and force-pushed.

java-client/src/test/java/org/opensearch/client/opensearch/experiments/ParsingTests.java
Mstrutov: Looks like it is `ModelTestCase`'s responsibility. Added a third commit which makes test code use its methods when dealing with jsons.

CHANGELOG.md
```diff
@@ -64,6 +64,8 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Fix failure when deserialing response for tasks API ([#463](https://github.com/opensearch-project/opensearch-java/pull/463))
 - Fix failure when deserializing boolean types for enums ([#463](https://github.com/opensearch-project/opensearch-java/pull/482))
 - Fix missing minScore, postFilter, searchAfter, sort, trackScores in the MultisearchBody ([#516](https://github.com/opensearch-project/opensearch-java/pull/516))
+- Fix missing translog sync interval option in index settings ([#518](https://github.com/opensearch-project/opensearch-java/pull/518))
```
dblock, line 67: I think we should move this and the one above "fix missing" into "added", these are new options that are added into the client not bug fixes.

dblock (approved):

dblock (dismissed):

dblock:
LGTM! @reta?

reta (approved):

CHANGELOG.md
```diff
@@ -72,6 +72,8 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Add buffered lookahead for Jackson ([#338](https://github.com/opensearch-project/opensearch-java/pull/338))
 - Add support for headers and sort parameters in cat requests ([#413](https://github.com/opensearch-project/opensearch-java/pull/413))
 - Add support for data stream operations ([#419](https://github.com/opensearch-project/opensearch-java/pull/419))
+- Add missing minScore, postFilter, searchAfter, sort, trackScores in the MultisearchBody ([#516](https://github.com/opensearch-project/opensearch-java/pull/516))
+- Add translog option object and missing translog sync interval option in index settings ([#518](https://github.com/opensearch-project/opensearch-java/pull/518))
```
reta, line 76: Ah ... apologies @Mstrutov , this line should go under `Unreleased 2.x` section

CHANGELOG.md
```diff
@@ -72,6 +72,8 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Add buffered lookahead for Jackson ([#338](https://github.com/opensearch-project/opensearch-java/pull/338))
 - Add support for headers and sort parameters in cat requests ([#413](https://github.com/opensearch-project/opensearch-java/pull/413))
 - Add support for data stream operations ([#419](https://github.com/opensearch-project/opensearch-java/pull/419))
+- Add missing minScore, postFilter, searchAfter, sort, trackScores in the MultisearchBody ([#516](https://github.com/opensearch-project/opensearch-java/pull/516))
```
reta, line 75: This changelog entry needs revert I believe

VachaShah (dismissed):

CHANGELOG.md
Mstrutov: I agree, it is probably useless to modify changelog of a released version

CHANGELOG.md
Mstrutov: Oh, I see, pardon me! I have also added a deprecation line --- please, see if that was done correctly.

reta (approved):

CHANGELOG.md
reta: All good, thanks @Mstrutov !

VachaShah (approved):
Thank you @Mstrutov!

Issue #437, opened 2023-04-07
[BUG] dense_vector instead of knn_vector


rogeriochaves:
### What is the bug?

The index builder has the dense_vector property (https://www.javadoc.io/doc/org.opensearch.client/opensearch-java/latest/org/opensearch/client/opensearch/_types/mapping/DenseVectorProperty.html), but dense_vector doesn't actually exists on OpenSearch, only on ES, if I try to use it, I get a 400.

On OpenSearch, it should actually be knn_vector, as it was decided on this RFC (https://github.com/opensearch-project/OpenSearch/issues/3545), but this option is not available

### How can one reproduce the bug?

Kotlin code:

```
        builder
            .properties(indexedProductSemanticVectorKey) { p ->
                p.denseVector(
                    DenseVectorProperty.of { d ->
                        d.dims(1536)
                    }
                )
            }
```

### What is the expected behavior?
To have a knn option available, or that the denseVector one just works

### What is your host/environment?
AWS

wbeckler:
good catch. Feel free to PR a fix.

wisocn:
Bumping this for visibility - got the same exact issue today

maltehedderich:
I am going to start working on this issue now.

rogeriochaves:
this worked well for us in the new version of the lib, thank you very much

PR #524, opened 2023-06-08
Add support for knn_vector property type

maltehedderich:
### Description
Adds support for the knn_vector property type and it's corresponding index settings.

### Issues Resolved
https://github.com/opensearch-project/opensearch-java/issues/437


java-client/src/main/java/org/opensearch/client/opensearch/_types/mapping/FieldType.java
```diff
@@ -124,6 +124,7 @@ public enum FieldType implements JsonEnum {
 
 	DenseVector("dense_vector"),
 
+	KnnVector("knn_vector"),
```
reta, line 127: So this is not OpenSearch core field type, this is provided by the plugin (https://github.com/opensearch-project/k-NN), afaik we do not (and probably should not) make it a part of the standard Java client but part of the k-NN plugin APIs, @VachaShah @dblock what do you think?

dblock: Right. This effectively creates a chicken/egg problem, which is why plugins are "awesome" :)

@maltehedderich Is there a mechanism that lets us add a k-nn high level REST client JAR to the k-nn plugin with this code? or maybe it's a waste of time and it makes more sense to add support for k-nn to opensearch-java's OpenSearch transport that doesn't have a dependency on core? See https://github.com/opensearch-project/opensearch-java/issues/326 and https://github.com/opensearch-project/opensearch-java/issues/262

maltehedderich: Thank you for the quick replies.

While namespaces probably work well for most plugins, this is difficult in the case of the KNN plugin due to its tight integration with core components.

The naming for both cluster and index-specific settings makes it difficult to identify them as plugin settings. As a user, if KNN settings remain under index settings, I would expect to be able to modify and read them through the Index Settings functionality in the Java client.

Moreover, I'm not aware of other plugins introducing custom field types. I think many users who use the OpenSearch KNN functionalities also use other field types (those who want a pure vector database have many alternatives). So it would be important to be able to define the mappings together with your other mappings. 

I believe that in order to solve this cleanly we would either have to incorporate the functionality of the KNN plugin into the OpenSearch core or fundamentally consider how to deal with extensions to field types and index settings.
Personally, I think integrating vector search capabilities into the OpenSearch core is the best solution. However, if this is not in line with your product vision I am also willing to spend time on alternative solutions, we should just clarify in advance where we really want to go.

As for the namespacing solution, it's achievable but would necessitate several class extensions and potentially affect usability. Regarding incorporating this into the new Transport Client, I'm unsure how this would look like and would need more information about your vision for that integration.

reta: > While namespaces probably work well for most plugins, this is difficult in the case of the KNN plugin due to its tight integration with core components.

Most if not all plugins have tight integration with core components

> Moreover, I'm not aware of other plugins introducing custom field types. 

There are many, even within OpenSearch core plugins, fe `mapper-murmur3` introduces `murmur3` custom field type.

> I believe that in order to solve this cleanly we would either have to incorporate the functionality of the KNN plugin into the OpenSearch core or fundamentally consider how to deal with extensions to field types and index settings.

I think more broadly we need an extensible mechanism to support plugin specific APIs in the Java client, we have talked about that in the past in scope of different issues (https://github.com/opensearch-project/opensearch-java/issues/377, https://github.com/opensearch-project/opensearch-java/issues/262) but have not conclude anything on the subject.

> As for the namespacing solution, it's achievable but would necessitate several class extensions and potentially affect usability. Regarding incorporating this into the new Transport Client, I'm unsure how this would look like and would need more information about your vision for that integration.

We need to get back to the extensibility part of OpenSearch Java client. The major drawback I see is that by adding `k-NN` plugin to the core client we put all other plugins in unfair and disadvantageous position, I think we need to work towards long term goal instead of taking the simple path of baking everything in.

maltehedderich: > There are many, even within OpenSearch core plugins, fe mapper-murmur3 introduces murmur3 custom field type.

How is "core plugin" defined? I am only aware of the distinction between "Bundled Plugins" and "Additional Plugins" (according to [installing plugins](https://opensearch.org/docs/latest/install-and-configure/plugins/)). Unlike murmur3, k-NN belongs to the Bundled Plugins and should therefore be present in the vast majority of OpenSearch deployments.

> I think more broadly we need an extensible mechanism to support plugin specific APIs in the Java client, we have talked about that in the past in scope of different issues (https://github.com/opensearch-project/opensearch-java/issues/377, https://github.com/opensearch-project/opensearch-java/issues/262) but have not conclude anything on the subject.

The discussions sound to me less like extensibility and more like the creation of a workaround through a kind of low level client. This could then of course be used as a basis for plugin specific libraries, but would itself not make the features of the Java client (typing, builders etc.) usable for e.g. users of the k-NN plugin.
According to my understanding of extensibility, there would need to be a standardized way to extend the components of the Java client with plugin specific types and functionalities. So that mappings, index settings and co can all be defined together via the builder without the need for custom HTTP requests.

> We need to get back to the extensibility part of OpenSearch Java client. The major drawback I see is that by adding k-NN plugin to the core client we put all other plugins in unfair and disadvantageous position, I think we need to work towards long term goal instead of taking the simple path of baking everything in.

The distinction between "bundled plugins" and "additional plugins" already puts all "additional plugins" at a disadvantage. In addition, I think it is questionable how much of the "plugin character" is left if they are shipped with all but the minimal distributions.

As you said, there are already several ideas and discussions about this topic, not only here in the Java Client repo. In my opinion it is important that we come to a holistic decision how to deal with "bundled" and "additional" plugins within the clients. How do you think we can reach such a decision most effectively?

The rough concepts I have seen for this so far are:
1. Integration into core client libraries with appropriate namespacing
2. Generic low level clients provided by the core client libraries
3. Development of plugin-specific clients which may build up on the core clients

reta: Thanks @maltehedderich 

> How is "core plugin" defined? I am only aware of the distinction between "Bundled Plugins" and "Additional Plugins" (according to [installing plugins](https://opensearch.org/docs/latest/install-and-configure/plugins/)).

Sorry, I should have been more clear, your distinction is very correct, by core plugins I meant the plugins included in OpenSearch repository and bundled with it. 

> The distinction between "bundled plugins" and "additional plugins" already puts all "additional plugins" at a disadvantage. 

There should be no distinction, I just provided an example of field type added in the plugin (since you mentioned you were not aware of), no emphasize on core - it just happened to be a plugin we bundle with OpenSearch. 

> The rough concepts I have seen for this so far are:
> 1.   Integration into core client libraries with appropriate namespacing
> 2.  Generic low level clients provided by the core client libraries
> 3.  Development of plugin-specific clients which may build up on the core clients


The solution (as I see) it is the combination of plugin specific APIs that could be plugged into OpenSearch Java client. The very rough idea of how it could look like (it is not formalized anywhere):

```
public class OpenSearchClient extends ApiClient<OpenSearchTransport, OpenSearchClient> {
	public <PluginClient> PluginClient plugin(Class<PluginClient> plugin) {
		....
	}
}
```


```
class OpenSearchKnnPluginClient {
  ....
}
```

But it does not include plugin specific mapping types, index settings, query builders, etc. , I think we could work it through by taking the `k-NN` plugin as the example and formalize the approach.

dblock:
Let's go back to our original problem @maltehedderich and @reta. I think we want support for `knn_vector` in opensearch-java. So what's the shortest path forward that we can all agree to (and pending open issue for the longer path)?

reta:
> So what's the shortest path forward that we can all agree to, and merge (and open issues for all the should/nice to have work)?

This pull request I think is the shortest path to support `knn_vector` field type

dblock:
> > So what's the shortest path forward that we can all agree to, and merge (and open issues for all the should/nice to have work)?
> 
> This pull request I think is the shortest path to support `knn_vector` field type

Shall we just merge it and open issues for the rest?

reta:
> Shall we just merge it and open issues for the rest, or do you think it's too much debt?

Taking all pros and cons, I would incline to merge it - it won't make things much worth then they are now, but kicking off the work to design the mechanism to plugin/extension  APIs would be great.

java-client/src/test/java/org/opensearch/client/opensearch/_types/mapping/KnnVectorPropertyTest.java
```diff
@@ -0,0 +1,69 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.client.opensearch._types.mapping;
+
+import static junit.framework.TestCase.assertEquals;
```
reta, line 11:
```suggestion
import static org.junit.Assert.assertEquals;

```

java-client/src/main/java/org/opensearch/client/opensearch/_types/mapping/KnnVectorProperty.java
```diff
@@ -0,0 +1,206 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+/*
```
reta, line 9: If it is not copied from Elasticsearch, please remove this header (for brand new files we need only OpenSearch license header)

java-client/src/main/java/org/opensearch/client/opensearch/_types/mapping/KnnVectorMethod.java
```diff
@@ -0,0 +1,229 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+/*
```
reta, line 9: If it is not copied from Elasticsearch, please remove this header (for brand new files we need only OpenSearch license header)

java-client/src/main/java/org/opensearch/client/opensearch/_types/mapping/FieldType.java
```diff
@@ -124,6 +124,7 @@ public enum FieldType implements JsonEnum {
 
 	DenseVector("dense_vector"),
```
reta, line 125: @maltehedderich please correct me if I am wrong, the `knn_vector` replaces `dense_vector`, right? Should we drop `dense_vector` if that is the case?

maltehedderich: The only (very weak) reason I would see is compatibility with elasticsearch (and maybe opendistro), where dense vector is a valid field type.

maltehedderich: Should I drop it within this PR or create another?

maltehedderich:
I forgot the --signoff flag when I committed the removal of the license header, but all my commits are signed with my SSH key. Is that enough or how should we solve that?

All code is written by myself, I just used one of the other Property Classes as template and kept the header because it was present in all other files.

java-client/src/main/java/org/opensearch/client/opensearch/_types/mapping/FieldType.java
reta: We could drop it in this pull request I think, thank you.

reta:
> I forgot the --signoff flag when I committed the removal of the license header, but all my commits are signed with my SSH key. Is that enough or how should we solve that?
> 
> I forgot the --signoff flag when I committed the removal of the license header, but all my commits are signed with my SSH key. Is that enough or how should we solve that?

As far as DCO checks are not failing, we should be good, thanks @maltehedderich !

dblock:
@reta But also, where's the DCO check?!

reta:
> @reta But also, where's the DCO check?!

It is no there! And also gone for OpenSearch, https://github.com/opensearch-project/OpenSearch/pull/8016

java-client/src/main/java/org/opensearch/client/opensearch/_types/mapping/FieldType.java
maltehedderich: From my current perspective, I think we definitely want something like that, but (as you already said) it only solves the problem with plugin specific APIs. However, in the case of the k-NN plugin, the standard APIs are used almost exclusively (except for training custom models?).

Nevertheless, I can understand your concerns about including plugin-specific code in the core library and will be happy to participate in possible solutions.

dblock:
> >

It's back!

maltehedderich:
I removed the dense_vector property and force pushed the signing to my previous commit. It's getting late here in Germany so I'm out, will fix it tomorrow if anything is still wrong.

dblock (approved):
Looks good. We will need updates to USER_GUIDE and such.

As a followup, let's talk about namespacing this with the word "plugin" better?

reta (approved):

maltehedderich:
> Looks good. We will need updates to USER_GUIDE and such.

Regarding the USER_GUIDES, I would prepare some small examples, but probably won't get to it until the next few days.  What is still unclear to me is the distinction to the [Java Client Docs](https://opensearch.org/docs/latest/clients/java/), do they pursue the same goal? Should we extend those as well?

> As a followup, let's talk about namespacing this with the word "plugin" better?

Yes. In general, we wanted to take another look at how we handle plugin-specific APIs and should also discuss there how we can separate plugin-specific extensions of the standard APIs better.

maltehedderich:
I have now added some examples for k-NN search to the USERGUIDE. 

Examples for the approximative k-NN search are still absent, because for this a KnnQuery class has to be added within our query_dsl first. I'm pretty busy the next two weeks, so I'll only be able to add this in early/mid July. But since exact k-NN search is already fully functional with the modifications in this PR, I would suggest to merge it and add the approximate k-NN search later.

I also changed the parameter order for the Apache HttpHost in the USERGUIDE, as it didn't seem to match the referenced [`org.apache.hc.core5.http`](https://hc.apache.org/httpcomponents-core-5.1.x/current/httpcore5/apidocs/org/apache/hc/core5/http/HttpHost.html) library before. However, the whole USERGUIDE seems still to be quite outdated and many examples do not work with the current library.

reta (approved):

dblock (approved):
Let's merge this!

maltehedderich:
> I think we do want to backport this to 2.x, do we? @maltehedderich would need to be done manually as the auto-backport failed.

Yes, that should be possible with reasonable effort. Shall we just create an issue for it and if it's not already done in the next few weeks, I'll take care of it in July?

Additionally, Mend seems to be unhappy with a dependency in the main branch, does dependabot take care of that automatically or do we have to adjust it manually?

PR #548, opened 2023-07-03
Add support for approximate k-NN queries

### LINKED ISSUE #539: [FEATURE] Add knn query support

Currently it is not possible to construct a query using opensearch-java with a knn query within it using the `Query` type:

https://opensearch.org/docs/latest/search-plugins/knn/filter-search-knn/#boolean-filter-with-ann-search
https://opensearch.org/docs/latest/search-plugins/knn/filter-search-knn/#lucene-k-nn-filter-implementation

opensearch-java does not contain KnnQuery as a variant it supports and due to the [Query.Kind](https://github.com/opensearch-project/opensearch-java/blob/main/java-client/src/main/java/org/opensearch/client/opensearch/_types/query_dsl/Query.java#L69) being a enum and the [QueryVariant](https://github.com/opensearch-project/opensearch-java/blob/main/java-client/src/main/java/org/opensearch/client/opensearch/_types/query_dsl/QueryVariant.java#L44) interface requiring a kind being specified it cannot be implemented as a separate library.

Given that opensearch-project/k-NN#524 adds precedent for adding KNN functionality into opensearch-java directly and has support for script scoring KNN integration, it seems reasonable that KNN query support be added directly into opensearch-java by adding a `KNN` variant to the `Query.Kind` enum and adding `KnnQuery` and `KnnQuery.Builder` classes to allow for the construction of these queries.

`Query.Kind` being an enum is restrictive in itself, preventing composability of client library support for non-bundled plugins. It could be that removing this restriction allows for the client to be built separately. But given opensearch-project/k-NN#524 was merged it seems like given if this restriction was removed, it still may be advantageous for consistency to add the KNN query support

Its not clear whether this request is the exact same thing as https://github.com/opensearch-project/opensearch-java/issues/547

Xtansia:
### Description
Adds support for the approximate k-NN query type to pair with the support for the `knn_vector` type added in #524. Along with a couple bug fixes for that support.

### Issues Resolved
#539 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,8 +28,9 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ## [Unreleased 2.x]
 
 ### Added
-- Add support for knn_vector field type ([#529](https://github.com/opensearch-project/opensearch-java/pull/524))
+- Add support for knn_vector field type ([#524](https://github.com/opensearch-project/opensearch-java/pull/524))
 - Add translog option object and missing translog sync interval option in index settings ([#518](https://github.com/opensearch-project/opensearch-java/pull/518))
+- Add support for approximate k-NN queries ([#548](https://github.com/opensearch-project/opensearch-java/pull/548))
 
 ### Dependencies
 - Bumps `jackson` from 2.14.2 to 2.15.2 ((#537)[https://github.com/opensearch-project/opensearch-java/pull/537])
--- a/java-client/src/main/java/org/opensearch/client/opensearch/indices/IndexSettings.java
+++ b/java-client/src/main/java/org/opensearch/client/opensearch/indices/IndexSettings.java
@@ -1066,7 +1066,7 @@ public class IndexSettings implements JsonpSerializable {
 
 		}
 		if (this.knnAlgoParamEfSearch != null) {
-			generator.writeKey("knn.algo_param_ef_search");
+			generator.writeKey("knn.algo_param.ef_search");
 			generator.write(this.knnAlgoParamEfSearch);
 
 		}
--- a/.github/workflows/test-integration-unreleased.yml
+++ b/.github/workflows/test-integration-unreleased.yml
@@ -3,8 +3,7 @@ name: Integration with Unreleased OpenSearch
 on:
   push:
     branches:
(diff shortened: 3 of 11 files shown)


java-client/src/main/java/org/opensearch/client/opensearch/indices/IndexSettings.java
```diff
@@ -1066,7 +1066,7 @@ protected void serializeInternal(JsonGenerator generator, JsonpMapper mapper) {
 
 		}
 		if (this.knnAlgoParamEfSearch != null) {
-			generator.writeKey("knn.algo_param_ef_search");
+			generator.writeKey("knn.algo_param.ef_search");
```
reta, line 1069: :+1:

java-client/src/main/java/org/opensearch/client/opensearch/indices/IndexSettings.java
```diff
@@ -1066,7 +1066,7 @@ protected void serializeInternal(JsonGenerator generator, JsonpMapper mapper) {
 
 		}
 		if (this.knnAlgoParamEfSearch != null) {
```
reta, line 1068: Not related to this change, but it seems like a bunch of KNN index settings are missed:

```
"settings" : {
        "index": {
            "knn": true,
            "knn.algo_param.m":90,
            "knn.algo_param.ef_construction":200,
            "knn.algo_param.ef_search":300
        }
    },
```

https://github.com/opensearch-project/k-NN/blob/main/RFC.md?plain=1

CHANGELOG.md
```diff
@@ -28,8 +28,9 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ## [Unreleased 2.x]
 
 ### Added
-- Add support for knn_vector field type ([#529](https://github.com/opensearch-project/opensearch-java/pull/524))
+- Add support for knn_vector field type ([#524](https://github.com/opensearch-project/opensearch-java/pull/524))
```
reta, line 31: :+1:

reta (dismissed):

reta (approved):

dblock (approved):
Do we need changes to USER_GUIDE? Can be followed up later.

Issue #534, opened 2023-06-21
[FEATURE] Support bucket size for MultiTermAggregation


fabiankopatschek:
### Is your feature request related to a problem?
_Using the MultiTermAggregation returns the default bucket size of 10. By setting the size parameter (see [Bucket aggregations](https://opensearch.org/docs/latest/aggregations/bucket-agg/#multi-terms)), you could get more buckets. In this library the possibilty is missing to set the "size" parameter. The [MultiTermAggregation](https://github.com/opensearch-project/opensearch-java/blob/2.x/java-client/src/main/java/org/opensearch/client/opensearch/_types/aggregations/MultiTermsAggregation.java#L54) Object has only the "terms" attribute._

### What solution would you like?
_Add the missing attribute to the MultiTermsAggregation type._

### What alternatives have you considered?
_An alternative could be using any HttpClient library instead._

### Do you have any additional context?
_In this forum post the issue was also addressed by someone else: [No Support to change the default bucket value in MultiTermAggregation](https://forum.opensearch.org/t/no-support-to-change-the-default-bucket-value-in-multitermaggregation/14360)_

_In all the type classes is the hint, that it is generated code. But i could not find any documentation on how or where to set the attribute and generate the type classes. Maybe then i would have done a Pull Request instead of this Feature Request._

wbeckler:
This sounds like a good idea. Feel free to propose an interface, or even better, a test case that would illustrate the correct approach.

fabiankopatschek:
I added the code in my fork. The forked version is used in our production system.
I created a pull request for the changes.
I noticed that the "generated code comment" is no longer applicable [#598](https://github.com/opensearch-project/opensearch-java/pull/598)
I couldn't find any exclusive testcases for this model classes, so there is no new testcase.

VachaShah:
Done via #627

PR #614, opened 2023-09-04
Added size attribute to MultiTermsAggregation

fabiankopatschek:
### Description

Added size attribute to MultiTermsAggregation to retreive other hit sizes than the default value of the attribute.
Code for the added attribute was taken from TermsAggregation class.

### Issues Resolved

- https://github.com/opensearch-project/opensearch-java/issues/534 


java-client/src/main/java/org/opensearch/client/opensearch/_types/aggregations/MultiTermsAggregation.java
```diff
@@ -37,25 +37,31 @@
 import org.opensearch.client.json.JsonpMapper;
 import org.opensearch.client.json.ObjectBuilderDeserializer;
 import org.opensearch.client.json.ObjectDeserializer;
+import org.opensearch.client.opensearch._types.aggregations.BucketSortAggregation.Builder;
```
reta, line 40: Does not look like the right import

reta:
Thanks for contribution @fabiankopatschek , could you please add a test cases under `AbstractAggregationRequestIT`? Thank you.

dblock:
Also this should go against `main`, and we can backport it after we're good there.

java-client/src/main/java/org/opensearch/client/opensearch/_types/aggregations/MultiTermsAggregation.java
fabiankopatschek: Thanks for the finding. I removed this line.

fabiankopatschek:
> Thanks for contribution @fabiankopatschek , could you please add a test cases under `AbstractAggregationRequestIT`? Thank you.

I really would like to add a test case. But unfortunately i'm not able to run the integration tests. 
`31 tests completed, 24 failed, 7 skipped`

The AggregationRequestIT for example failes with this message:

> java.lang.RuntimeException: found jar hell in test classpath
	at org.opensearch.bootstrap.BootstrapForTesting.<clinit>(BootstrapForTesting.java:118)
	...

Any tips on how to fix this?

reta:
> Any tips on how to fix this?

Are you running with Eclipse? Or gradle?

fabiankopatschek:
> > Any tips on how to fix this?
> 
> Are you running with Eclipse? Or gradle?

I'm running the tests in the shell with gradle.
I've noticed that the integration test for this MR is having the same trouble.
I encounter the same problems with the main branch as well without any changes from me.

reta:
> I've noticed that the integration test for this MR is having the same trouble.

You are very right, the issue seems to be caused by latest software.amazon.awssdk 2.20.141 (which is imported using version range). Created an issue for it here https://github.com/opensearch-project/opensearch-java/issues/615

fabiankopatschek:
I migrated the code changes to the main branch as [dblock](https://github.com/dblock) suggested.
I also added an Integration Test as [reta](https://github.com/reta) suggested.
There is a new pull request #627 for the changes.

Issue #631, opened 2023-09-21
[BUG] RestClientTransport throws NoSuchMethodError


andrewparmet:
### What is the bug?
RestClientTransport is unusable:

```
java.lang.NoSuchMethodError: 'void org.opensearch.client.Request.setEntity(org.apache.hc.core5.http.HttpEntity)'
  at org.opensearch.client.transport.rest_client.RestClientTransport.prepareLowLevelRequest(RestClientTransport.java:221)
  at org.opensearch.client.transport.rest_client.RestClientTransport.performRequest(RestClientTransport.java:148)
  at org.opensearch.client.opensearch.indices.OpenSearchIndicesClient.create(OpenSearchIndicesClient.java:244)
  at org.opensearch.client.opensearch.indices.OpenSearchIndicesClient.create(OpenSearchIndicesClient.java:258)
```

Using latest HEAD - `d09bb4e`.

### How can one reproduce the bug?
Use the RestClientTransport.

### What is the expected behavior?
Requests should work.

### What is your host/environment?
Mac OS.

### Do you have any screenshots?
Nope.

### Do you have any additional context?
Nope.

reta:
Please check https://github.com/opensearch-project/opensearch-java/blob/main/USER_GUIDE.md#creating-a-client, the `RestClientTransport` is deprecated in 3.0.0.

PR #656, opened 2023-10-06
Recreate or wrap exceptions thrown by async transport implementations to preserve caller stack traces

### LINKED ISSUE #653: [BUG] Synchronous calls with async transport should preserve caller stack trace
### What is the bug?

When an OpenSearchClient built with async transport fails, the transports unwrap causes and propagate them if they're RuntimeExceptions. We lose the stack trace since the exception originated in some pooled thread. Can these exceptions instead be re-wrapped in an IOException and propagated as the interface allows?

As it stands I just wrap the existing transport with one that catches and wraps this way, so it's not the end of the world, but it was confusing that we got random stack traces from a "synchronous" call.

andrewparmet:
### Description
Ensures that callers of synchronous methods will see their method name in the stracktrace of an exception thrown by an asynchronous transport.

### Issues Resolved
Fixes #653.


java-client/src/test/java/org/opensearch/client/opensearch/integTest/AbstractClusterClientIT.java
```diff
@@ -93,10 +93,17 @@ public void testClusterUpdateSettingNonExistent() throws IOException {
         try {
             openSearchClient.cluster().putSettings(request);
             fail();
-        } catch (OpenSearchException e) {
-            assertNotNull(e);
-            assertEquals(e.response().status(), 400);
-            assertTrue(e.getMessage().contains("transient setting [no_idea_what_you_are_talking_about], not recognized"));
+        } catch (Exception e) {
+            OpenSearchException openSearchException;
+            if (e instanceof OpenSearchException) {
```
dblock, line 98: I wouldn't expect conditionals in tests, is this not predictable?

dblock (commented):
Very cursory review for now ...

1. What is the the user experience change with this?
2. Anything breaking in a semver sense?
3. Let's also update some user guides on exceptions/error handling as part of this PR?

java-client/src/test/java/org/opensearch/client/opensearch/integTest/AbstractAsyncStracktraceIT.java
```diff
@@ -0,0 +1,25 @@
...
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.client.opensearch.integTest;
+
+import org.apache.logging.log4j.core.util.Throwables;
+import org.junit.Test;
+
+public abstract class AbstractAsyncStracktraceIT extends OpenSearchJavaClientTestCase {
```
andrewparmet, line 14: I'd like to be able to mimic the existing pattern for RestClient and Apache HTTP5 integration tests and write one abstract test class that both Apache HTTP5 and AWS SDK V2 clients can use. I don't have it configured correctly here. When the CI suite runs these tests the AWS test is executing when it shouldn't be.

Altogether it looks like there's no great universal parent class that achieves exactly what I'm looking for but `OpenSearchJavaClientTestCase` seems to be the closest thing to what I'm looking for.

java-client/src/test/java/org/opensearch/client/opensearch/integTest/AbstractClusterClientIT.java
andrewparmet: It's predictable but now varies per implementation.

This test is run against the RestClient transport and the Apache HTTP 5 transports. The RestClient transport throws OpenSearchException directly and since its transport is synchronous, this exception's stack trace contains a reference to the caller. The HC5 transport is _always_ asynchronous and while the underlying logic will throw an OpenSearchException, that logic is executing on a pooled thread. That OpenSearchException then has to be wrapped in order to preserve a reference to the caller in the thrown exception's stack trace. I cannot construct an OpenSearchException wrapping an OpenSearchException (which would be suspicious anyways), so this now has to catch any exception.

The correct way to model this is to parameterize each _implementation_ of this abstract class by the type of exception it should be expected to throw and use the `assertThrows` pattern here instead of `try-catch`.

This modeling issue combined with https://github.com/opensearch-project/opensearch-java/pull/656#discussion_r1349301994 are two overarching pieces of test tech debt I think should be handled separately from this PR, but would enable more comprehensive transport-agnostic integration tests.

dblock: Thanks for the detailed explanation. Open GitHub issues for these leftovers so we don't forget / someone else can help.

dblock (approved):

reta:
@dblock I don't think we should be doing this change for 2 reasons:
 - there will be runtime exception  in the code (wrapped in the `CompletionException` or not) and it is normal to let them pop
 - this could be considered a breaking change, since some exceptions that was propagated before as they should become now wrapped into `IOException`s

reta:
> While it's generally acceptable to let RuntimeExceptions propagate, it's expected that the stacktrace will include your caller. Would it be acceptable to re-wrap with a different runtime exception here? Perhaps OpenSearchRuntimeException?

Let's take an example of the `CompletionException` here: as we know, this is wrapper. Ideally, what we would like to have here is the exact exception that the code being called has raised (and currently, implementation tries to do that). With respect to the caller stacktrace: async code is notoriously difficult to troubleshoot but we have only 1 single place in that does that (`performRequestAsync`). How it is helpful? it could be convenient but could you please provide an example that it is necessary?

> Is this not permitted in 3.0.0? Since it's unreleased, such a breaking change would be permitted by semver.

Sure, for 3.x we are fine, we won't be backporting it which could be a maintenance issue (we will have to support 2 different behaviours across 2.x and 3.x branches).

andrewparmet:
> async code is notoriously difficult to troubleshoot but we have only 1 single place that does that (performRequestAsync). How it is helpful to have it included?

That one single place is on the code path of every call the client makes :).

A `logger.error` call records the stack trace of the caught exception, but we have no way of correlating it with a call to the OpenSearch client. We can't answer the question: "Which of our calls to OpenSearch during this request failed?" A caller stack trace is the idiomatic way to answer that question.

In some cases the OpenSearchClient uses async transport for synchronous calls: for example, when using the Apache HC5 transport. This also happens when making synchronous calls using the AWS async client which we reuse across synchronous and asynchronous calls to enable connection reuse. Notably we have no synchronous transport we can use in tests (see https://github.com/opensearch-project/opensearch-java/issues/631) and this makes it much harder to debug new code.

I think the root of the problem is really that OpenSearchClient can be built with async transport when it only offers synchronous APIs. In contrast AWS clients offered by the v2 Java SDK are completely split between sync and async variants and you cannot construct one with a transport of the other kind. 

Clients built with the other kind of transport compromise their behavior: an async client built with sync transport must wrap synchronous calls with Futures or in Executors (RestClientTransport does the first). A sync client built with async transport cannot offer the same API for thrown exceptions since it must rewrap exceptions or throw their wrapping completion/execution exceptions (HC5/AwsSdk do this).

The solution is to either accept a compromise or to disable this kind of "square peg round hole" construction. I think that obfuscating caller traces is a bad compromise so I filed an issue to achieve consensus before attempting to change the behavior. I didn't dump too many hours into this, so if the consensus isn't there, I do have the functioning workaround I mentioned. I don't think it's a good pattern to require from clients of a library, especially if you're free to change behavior in a new major version.

> for 3.x we are fine

Good to hear. We're upgrading from the RestHighLevelClient (1.3.2) and we won't need a backport.

dblock:
Thanks @andrewparmet @reta. I think my questions remain, we should work backwards from the developer experience that we want. Keeping an open mind about the solution!

reta:
> Thanks @andrewparmet @reta. I think my questions remain, we should work backwards from the developer experience that we want. Keeping an open mind about the solution!

@andrewparmet my apologies for the delay, so I have looked into 2 options here:
1. Suggested by you, the approach to rethrow exception: it will help to preserve the stacktrace but the risk of breaking things here is very high: -
   - we cannot preserve the exact exception type
   - we diverge the behaviour of the RestClientTransport and ApacheHttpClient5Transport (not only confusing but will make migration more difficult)
2. Adding dedicated sync client. It will work but sadly requires the API changes (since configuration callbacks are all bound to `AsyncHttpClient`, not ideal either.

I believe we should keep the current implementation unchanged and recommend people to use the `OpenSearchAsyncClient` directly to dial with exception management the way it fits the need.

andrewparmet:
> we diverge the behaviour of the RestClientTransport and ApacheHttpClient5Transport

They are already divergent in the sense that one preserves the trace of its caller and the other does not. Option 2 is the only way to be fully consistent. If HC5 offers synchronous transport then perhaps it shares a configuration type with the async transport.

reta:
> They are already divergent in the sense that one preserves the trace of its caller and the other does not even given equivalent server-side errors.

Oh that I missed out, my apologies for that, that is an argument towards making it right, looking into divergence now, thank you.

> If HC5 offers synchronous transport then perhaps it shares a configuration type with the async transport.

It offers but does not share anything, those are separate implementations

andrewparmet:
>  those are separate implementations

Boo, but that's probably the right way to do it from their side.

My ultimate position here then is to match the AWS SDK - async clients should take async transports and sync clients should take sync transports. Then, if you want to do what I'm doing and share a connection pool, do the wrapping and unwrapping of async calls yourself. Everybody gets exactly the API they want - all the way down to the way stacktraces look.

reta:
> Boo, but that's probably the right way to do it from their side.

Totally :)

> My ultimate position here then is to match the AWS SDK - async clients should take async transports and sync clients should take sync transports. T

I agree, the part which is difficult to translate seamless is `HttpClientConfigCallback` that requires `HttpAsyncClientBuilder` explicitly. 

@andrewparmet mind if I push a change to your pull request? TLDR; replicating the RestClient approach, essentially exactly the way you are trying to fix it here but with a bit more complicated logic. Thank you.

andrewparmet:
> HttpClientConfigCallback that requires HttpAsyncClientBuilder explicitly

Yeah this API would probably need to be renamed to distinguish between sync and async APIs, e.g. ApacheHttpClient5AsyncTransportBuilder.

Feel free to hijack this PR - the magic button should be enabled.

java-client/src/main/java/org/opensearch/client/transport/httpclient5/ApacheHttpClient5Transport.java
```diff
@@ -1013,4 +1020,72 @@ public InputStream asInput() {
             return new ByteArrayInputStream(this.buf, 0, this.count);
         }
     }
+
+    /**
+     * Wrap the exception so the caller's signature shows up in the stack trace, taking care to copy the original type and message
+     * where possible so async and sync code don't have to check different exceptions.
+     */
+    private static Exception extractAndWrapCause(Exception exception) {
```
reta, line 1028: @andrewparmet I am not very proud of this code, but it seems to be doing the job for now, regarding AwsSdk2Transport, I am not very familiar with it (@dblock is) but we could split this change in tow minimize any potential risks.

andrewparmet:
@dblock What is the current state of the exception API? Are you all happy with the implementation as it stands? In that case is the remainder of the work just getting the tests to pass?

dblock:
> @dblock What is the current state of the exception API? Are you all happy with the implementation as it stands? In that case is the remainder of the work just getting the tests to pass?

I don't think I know this code well enough to say. Maybe @reta ?

reta:
> I don't think I know this code well enough to say. Maybe @reta ?

@andrewparmet @dblock I basically replicated RestClient exception handling to Apache HttpClient 5 client, but not for `AwsSdk2Transport`, mind to finish this part? Thank you.

andrewparmet:
@dblock @reta Sorry for the delay here. This PR is now up to date with the latest idea on exception extraction implemented in the AWS transport.

It includes two tests - one for the AWS transport and one for the HC5 transport - that lightly verify that stack trace propagation works. Ideally there would be one abstract test that can cover both transports, but I didn't see an easy way to do that given the current setup of the test suite. Let me know if there's another approach you'd prefer. It does look like this may be easier to do once Java 8 support is dropped and AWS integration tests are moved into the Java 11 test source set so code can be shared.

Due to the new logic introduced in 03b214f1811d41d6b3f43f95f46cd36c507d8661, it does not completely cover all potential cases. In particular I think this strategy is slightly brittle as new exceptions may (will) be introduced that are not correctly captured. For example, this happened with OpenSearchClientException between the last time we touched this PR and now.

java-client/build.gradle.kts
```diff
@@ -205,6 +205,7 @@ dependencies {
     // For AwsSdk2Transport
     "awsSdk2SupportCompileOnly"("software.amazon.awssdk","sdk-core","[2.15,3.0)")
     "awsSdk2SupportCompileOnly"("software.amazon.awssdk","auth","[2.15,3.0)")
+    "awsSdk2SupportCompileOnly"("io.netty","netty-handler","4.1.108.Final")
```
reta, line 208: This dependency should not be needed, could you clarify why it is there?

andrewparmet: We see in our service that the AWS transport directly rethrows Netty exceptions, so in order to follow the catch-create-throw pattern from your implementation of the HC5 handling, we need access to the Netty dependency.

java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
```diff
@@ -8,12 +8,18 @@
 
 package org.opensearch.client.transport.aws;
 
+import io.netty.channel.ChannelPipelineException;
```
reta, line 11: I think we should not catch anything specific to `netty` or alike: AFAIK the client transport is pluggable in AWS SDK, we cannot cover all possible client impl

andrewparmet: Ok. This means that Netty exceptions will all be rethrown wrapped in IOException. I can revert this.

java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
```diff
@@ -615,4 +622,86 @@ private static <T> Optional<T> or(Optional<T> opt, Supplier<? extends Optional<?
...
+        }
+        if (exception instanceof ExecutionException) {
+            ExecutionException executionException = (ExecutionException) exception;
+            Throwable t = executionException.getCause() == null ? executionException : executionException.getCause();
+            if (t instanceof Error) {
+                throw (Error) t;
+            }
+            exception = (Exception) t;
+        }
+        if (exception instanceof ConnectTimeoutException) {
```
reta, line 638: AFAIK AWS SDK does not support ACH 5 (only ACH 4 as per https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/http-configuration.html), how come we're seeing this exceptions?

reta: Thank you

java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
reta: Thank you

reta:
> @dblock @reta Sorry for the delay here. This PR is now up to date with the latest idea on exception extraction implemented in the AWS transport.

Thanks a lot @andrewparmet , LGTM, does it solve your problem in an implementation shape like that? Thank you

andrewparmet:
It does - we'll be able to trace our calls at the very least to the rethrown variations at the bottom of the extraction functions.

reta (approved):

reta:
@dblock do you mind to check it out? thank you!

dblock:
> @dblock do you mind to check it out? thank you!

Looks great, merging.

@andrewparmet Thanks for hanging in here with us. I think the project could use some documentation along the "handling errors" lines that could explain the various scenarios that you've fixed and how one should deal with them. Maybe you want to write something simple to start?

PR #677, opened 2023-10-19
Add scroll to search request serialization.

### LINKED ISSUE #676: [BUG] SearchRequest is not serializing the Scroll time parameter
### What is the bug?
The Scroll API is not working properly because the ScrollId returned in the SearchResponse has an ill-defined validity.
The root cause appears to be that the SearchRequest `Scroll` parameter (or type `Time`) is not serialized, and thus the search validity cannot be specified.

Create a SearchRequest, and specify a `Scroll` with value `"5m"`, which is 5 minutes.
With the search response, use the scroll id to do a ScrollRequest;  
most of the time the server replies with "all shards failed".

The search should success with 0 or more results.

Observed on Alpine Linux (hosted on AWS) and Windows 11 (developer laptop).

No, but I have application logs (from our in-house application)

no.

frivard-coveo:
### Description
Add the `scroll` parameter when serializing the SearchRequest.

### Issues Resolved
#676 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 ### Removed
 
 ### Fixed
-- Fix for 'scroll' parameter not serialized to SearchRequest.
+- Fix for 'scroll' parameter not serialized to SearchRequest ([#677](https://github.com/opensearch-project/opensearch-java/pull/677))
 
 ### Security
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 ### Removed
 
 ### Fixed
+- Fix for 'scroll' parameter not serialized to SearchRequest.
 
 ### Security
 
--- a/java-client/src/main/java/org/opensearch/client/opensearch/core/SearchRequest.java
+++ b/java-client/src/main/java/org/opensearch/client/opensearch/core/SearchRequest.java
@@ -963,6 +963,10 @@ public class SearchRequest extends RequestBase implements JsonpSerializable {
             generator.writeEnd();
 
         }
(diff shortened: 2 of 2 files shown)


reta:
@frivard-coveo could you please add an entry to CHANGELOG.md? thank you

VachaShah:
> @VachaShah @dblock it seems like we don't have test cases for regular search functionality, I will add one shortly

Thank you @reta! We have some search scenarios in samples as well: https://github.com/opensearch-project/opensearch-java/blob/main/samples/src/main/java/org/opensearch/client/samples/Search.java

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 ### Removed
 
 ### Fixed
+- Fix for 'scroll' parameter not serialized to SearchRequest.
```
reta, line 41:
```suggestion
- Fix for 'scroll' parameter not serialized to SearchRequest ([#677](https://github.com/opensearch-project/opensearch-java/pull/677))
```

frivard-coveo: oh my god...   I had exactly that in my editor, but I somehow did not click "save" before committing (which IDE isn't auto-save nowadays?)

dblock:
This does need a test, let's add those as part of (merging) this PR.

reta:
> This does need a test, let's add those as part of (merging) this PR.

@dblock I've added basic Search API test (https://github.com/opensearch-project/opensearch-java/pull/678) so we could add more tests on top

frivard-coveo:
Oh no!

I just realized that this PR is all wrong. The `scroll` parameter is supposed to be part of the query path, not the request body.
See this doc : https://opensearch.org/docs/latest/api-reference/search/#url-parameters
Unfortunately this means I'm back to square one trying to understand why my scroll requests are failing with "all shards failed".

Issue #588, opened 2023-07-31
[FEATURE] serialize requests to JSON


dblock:
### Is your feature request related to a problem?

Add a way to easily serialize queries to JSON for logging.

### What solution would you like?

```java
            IndexRequest<Movie> indexRequest = new IndexRequest.Builder<Movie>()
                    .index(index)
                    .id("1")
                    .document(movie)
                    .build();

            logger.info("JSON: {}", indexRequest.toJson());
```

Working serialization code [here](https://github.com/dblock/opensearch-java-client-demo/blob/to-json/src/main/java/Example.java#L34).

### Do you have any additional context?

From a Slack discussion in https://opensearch.slack.com/archives/C051XNH62KB/p1690396113772759

wbeckler:
Sounds cool. I've seen a few people request this, and it's cleaner than using log4j.

PR #694, opened 2023-10-29
Serialize requests to json

Tamir-Schwarz:
link to issue
https://github.com/opensearch-project/opensearch-java/issues/588

actually used the code of @dblock  - thank you very much!

original example
https://github.com/dblock/opensearch-java-client-demo/blob/to-json/src/main/java/Example.java#L34

=========
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.

Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.

### DIFF
--- a/java-client/src/test/java/org/opensearch/client/opensearch/json/JsonDataTest.java
+++ b/java-client/src/test/java/org/opensearch/client/opensearch/json/JsonDataTest.java
@@ -40,7 +40,9 @@ import org.junit.Assert;
 import org.junit.Test;
 import org.opensearch.client.json.JsonData;
 import org.opensearch.client.json.JsonpMapper;
+import org.opensearch.client.json.JsonpUtils;
 import org.opensearch.client.json.jsonb.JsonbJsonpMapper;
+import org.opensearch.client.opensearch.core.IndexRequest;
 import org.opensearch.client.opensearch.model.ModelTestCase;
 
 public class JsonDataTest extends Assert {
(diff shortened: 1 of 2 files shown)


java-client/src/test/java/org/opensearch/client/opensearch/json/JsonDataTest.java
```diff
@@ -89,4 +96,65 @@ public void testConvert() {
...
+        String expectedJson = "{\"aga\":42,\"name\":\"koko\"}";
+        Person person = new Person(42, "koko");
+        IndexRequest<Person> indexRequest = new IndexRequest.Builder<Person>()
+                .index("sample-index")
+                .id("1")
+                .document(person)
+                .build();
+
+        String json = JsonpUtils.toJson(indexRequest);
+        System.out.println(json);
```
dblock, line 132: Get rid of print outs.

dblock (changes requested):
I think with some documentation in USER_GUIDE or related this could be good enough.

VachaShah:
> I think with some documentation in USER_GUIDE or related this could be good enough.

+1. Some documentation on how to use `toJson()` for a class would be great!

dblock:
@Tamir-Schwarz Want to finish this?

felipetoffolo-toast:
I don't think that this work. I can't find toJson function

Jai2305:
Hi @Tamir-Schwarz and @dblock , first of thanks for the contribution, all services of my organization are migrating from the High Level Rest Client to Java Client and very much like these kind of enhancements to be implemented, as they provide a clean code convenience to new clients.
Although I have some suggestions and I would like your opinion on them - 
* Can we not implement this method in JsonpSerializable interface as a default method, I think it will be a better design as it will ensure that only classes implementing the interface are allowed to utilize the default method, you could internally call serialize method of the interface, this will avoid implementation and call to a utils class.
* Also you can try implementing `JsonGenerator generator = mapper.jsonProvider().createGenerator(stringWriter);` with try with resource block to ensure that  the generator is closed automatically even if an exception occurs.
* Finally please include test cases for classes which are implementing JsonpSerializable indirectly, just to ensure the robustness of code , for eg. IndexResponse extends WriteResponseBase implements JsonpSerializable .

dblock:
@Jai2305 These look very reasonable. Want to open another PR with your suggestions?

Jai2305:
@dblock thanks for a quick response, I am on it.

Xtansia:
A version of this utility was added via #588, documented here: https://github.com/opensearch-project/opensearch-java/blob/main/guides/json.md#using-tojsonstring

Issue #721, opened 2023-11-09
[BUG] Unsupported ICU collation keyword field type


paverman:
### What is the bug?

Using of custom plugin-defined keyword field type [icu_collation_keyword](https://www.elastic.co/guide/en/elasticsearch/plugins/current/analysis-icu-collation-keyword-field.html#_parameters_for_icu_collation_keyword_fields) in the index mapping will lead to an error during response deserialization:

`Missing Required Property Builder.<variant kind>`

Can be reproduced by the following code:
```
index = "index_with_icu_collation_keyword"
GetMappingRequest getMappingsRequest = new GetMappingRequest.Builder().index(index).build();
response = client.indices().getMapping(getMappingsRequest);
```

Index mapping example:

```json
{
    "mappings": {
        "_doc": {
            "dynamic": "strict",
            "properties": {
                "FIELD_NAME": {
                    "copy_to": [
                        "all"
                    ],
                    "store": true,
                    "type": "text",
                    "fields": {
                        "sort": {
                            "type": "icu_collation_keyword"
                        },
                        "keyword": {
                            "type": "keyword"
                        }
                    }
                }
            }
        }
    }
}
```

Seems like this is [known issue ](https://github.com/elastic/elasticsearch-java/issues/249#issuecomment-1150082726)and was fixed in elasticsearch-java sources as a part of [this ](https://github.com/elastic/elasticsearch-java/pull/370)

Perhaps, this is a duplicate of https://github.com/opensearch-project/opensearch-java/issues/611 but with more specific details.

There is a possible [workaround ](https://www.elastic.co/guide/en/elasticsearch/client/java-api-client/7.17/missing-required-property.html) but it does not work with [Apache HttpClient 5 Transport](https://opensearch.org/docs/latest/clients/java/) due to a different threads issue.

### How can one reproduce the bug?

Use icu_collation_keyword in your index mapping.

### What is the expected behavior?

The icu_collation field type should be deserialized without errors.

### What is your host/environment?
Windows
OpenSearch Java Client 2.6.0
Java 11

Bfindlay:
You can assgn this to me. I have a fix thats working locally, just need to spend some time writing more tests.

paverman:
@Bfindlay Great, thank you. 
Will this fix be included in next release 2.9.0? What is the approximate date for the next release? Can I find somewhere a build package from the main branch to test this fix without waiting for the release?

Bfindlay:
[...]

PR #725, opened 2023-11-14
Add support for icu_collation_keyword field type

Bfindlay:
### Description
This PR adds support for the icu_collation_keyword field type.
Currently, if you have this common plugin enabled, and on an index mapping it will cause an error `Missing Required Property Builder.<variant kind>` when using getting templates or mappings.

### Issues Resolved
https://github.com/opensearch-project/opensearch-java/issues/721

### DIFF
--- a/java-client/src/main/java/org/opensearch/client/opensearch/_types/mapping/IcuCollationKeywordProperty.java
+++ b/java-client/src/main/java/org/opensearch/client/opensearch/_types/mapping/IcuCollationKeywordProperty.java
@@ -1,3 +1,11 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
 package org.opensearch.client.opensearch._types.mapping;
 
 import jakarta.json.stream.JsonGenerator;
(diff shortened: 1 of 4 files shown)


java-client/src/main/java/org/opensearch/client/opensearch/_types/mapping/IcuCollationKeywordProperty.java
```diff
@@ -0,0 +1,138 @@
...
+import jakarta.json.stream.JsonGenerator;
+import java.util.function.Function;
+import javax.annotation.Nullable;
+import org.opensearch.client.json.*;
+import org.opensearch.client.util.ObjectBuilder;
+
+// typedef: _types.mapping.ICUCollationKeywordProperty
+
+@JsonpDeserializable
+public class IcuCollationKeywordProperty extends DocValuesPropertyBase implements PropertyVariant {
```
reta, line 20: We probably should also support collation options [1] for these fields, right? 

[1] https://www.elastic.co/guide/en/elasticsearch/plugins/current/analysis-icu-collation-keyword-field.html#_collation_options

Bfindlay: @reta good callout. I have added those now.

reta (approved):

reta:
Thank you @Bfindlay !

dblock (approved):

Issue #666, opened 2023-10-13
[BUG] UnexpectedJsonEventException reading task that failed


uriofferup:
### What is the bug?
Failed to serialize JSON response when reading a task that failed to execute. 

### How can one reproduce the bug?
Run any operation using `waitForCompletiong(false)` that fail to execute (for example due to versioning issue).
Then read the taskId returned:
```
GetTasksRequest getTasksRequest = GetTasksRequest.of(t -> t.taskId(osTask.getTaskId()));
GetTasksResponse getTasksResponse = openSearchClient.tasks().get(getTasksRequest);
```

### What is the expected behavior?
Deserialize the response into the GetTasksResponse object.

### What is your host/environment?
AWS Lambda
ElasticSearch 7.10

### Do you have any screenshots?
Executing /_tasks/<task-id>
Gets this:
```
{
  "completed": true,
  "task": {
    "node": "xjAO-PN5TKarvcLhJmHfcA",
    "id": 47102523527,
    "type": "transport",
    "action": "indices:data/write/update/byquery",
    "status": {
      "total": 152,
      "updated": 151,
      "created": 0,
      "deleted": 0,
      "batches": 6,
      "version_conflicts": 1,
      "noops": 0,
      "retries": {
        "bulk": 0,
        "search": 0
      },
      "throttled_millis": 0,
      "requests_per_second": -1,
      "throttled_until_millis": 0,
      "slices": [
        {
          "slice_id": 0,
          "total": 26,
          "updated": 26,
          "created": 0,
          "deleted": 0,
          "batches": 1,
          "version_conflicts": 0,
          "noops": 0,
          "retries": {
            "bulk": 0,
            "search": 0
          },
          "throttled_millis": 0,
          "requests_per_second": -1,
          "throttled_until_millis": 0
        },
        {
          "slice_id": 1,
          "total": 23,
          "updated": 23,
          "created": 0,
          "deleted": 0,
          "batches": 1,
          "version_conflicts": 0,
          "noops": 0,
          "retries": {
            "bulk": 0,
            "search": 0
          },
          "throttled_millis": 0,
          "requests_per_second": -1,
          "throttled_until_millis": 0
        },
        {
          "slice_id": 2,
          "total": 29,
          "updated": 29,
          "created": 0,
          "deleted": 0,
          "batches": 1,
          "version_conflicts": 0,
          "noops": 0,
          "retries": {
            "bulk": 0,
            "search": 0
          },
[...]

PR #727, opened 2023-11-15
Fix GetTasksResponse TaskFailure deserialization 

Bfindlay:
### Description
The GetTasksResponse failures is currently trying to desierialize a list of String errors, whereas in reality the response is a map which results in a deserialization error.

There was also a missing property "cancelled" on the Status object, that I have added in this PR.

### Issues Resolved
https://github.com/opensearch-project/opensearch-java/issues/666

### DIFF
--- a/java-client/src/test/java/org/opensearch/client/opensearch/integTest/AbstractIndicesClientIT.java
+++ b/java-client/src/test/java/org/opensearch/client/opensearch/integTest/AbstractIndicesClientIT.java
@@ -38,6 +38,8 @@ import org.opensearch.client.opensearch.indices.PutIndexTemplateResponse;
 import org.opensearch.common.settings.Settings;
 
 public abstract class AbstractIndicesClientIT extends OpenSearchJavaClientTestCase {
+
+    @Test
     public void testIndicesExists() throws IOException {
 
         // Index present
@@ -85,6 +87,7 @@ public abstract class AbstractIndicesClientIT extends OpenSearchJavaClientTestCa
         }
     }
 
+    @Test
     public void testCreateIndex() throws Exception {
         OpenSearchAsyncClient asyncClient = new OpenSearchAsyncClient(javaClient()._transport());
         CreateIndexResponse createResponse = javaClient().indices().create(b -> b.index("my-index"));
@@ -101,6 +104,7 @@ public abstract class AbstractIndicesClientIT extends OpenSearchJavaClientTestCa
         assertNotNull(indices.get("my-index"));
     }
 
+    @Test
     public void testGetSettingsNonExistentIndex() throws IOException {
 
         String nonExistentIndex = "index_that_doesnt_exist";
@@ -117,6 +121,7 @@ public abstract class AbstractIndicesClientIT extends OpenSearchJavaClientTestCa
         }
     }
 
+    @Test
     public void testDataStream() throws IOException {
         String dataStreamIndexTemplateName = "test-data-stream-template";
         String timestampFieldName = "my_timestamp_field";
@@ -127,6 +132,7 @@ public abstract class AbstractIndicesClientIT extends OpenSearchJavaClientTestCa
         PutIndexTemplateResponse putIndexTemplateResponse = javaClient().indices()
             .putIndexTemplate(
                 b -> b.name(dataStreamIndexTemplateName)
+                    .indexPatterns("test*")
                     .dataStream(new DataStream.Builder().timestampField(bd -> bd.name(timestampFieldName)).build())
                     .indexPatterns(namePattern)
             );
@@ -188,6 +194,7 @@ public abstract class AbstractIndicesClientIT extends OpenSearchJavaClientTestCa
         }
     }
 
+    @Test
     public void testGetNotExistingIndexAlias() throws Exception {
         String notExistingIndexAlias = "alias_not_exists";
         GetAliasRequest aliasRequest = new GetAliasRequest.Builder().name(notExistingIndexAlias).build();
--- a/java-client/src/test/java/org/opensearch/client/opensearch/integTest/AbstractTasksIT.java
+++ b/java-client/src/test/java/org/opensearch/client/opensearch/integTest/AbstractTasksIT.java
@@ -1,0 +1,62 @@
+package org.opensearch.client.opensearch.integTest;
+
+import java.io.IOException;
+import org.junit.Test;
+import org.opensearch.client.opensearch.tasks.Status;
+
+public abstract class AbstractTasksIT extends OpenSearchJavaClientTestCase {
+
+    @Test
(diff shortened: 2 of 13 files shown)


dblock (dismissed):

reta:
@dblock  @Bfindlay the `Info` changes look good, but `Status` is not , I have been looking what OpenSearch server returns and it is indeed just opaque bytes for error or response, we may follow the same fix as Elastic folks [1] (this repository is ASF v2.0 licensed, for a reference).

[1] https://github.com/elastic/elasticsearch-specification/pull/2250

dblock:
> @dblock @Bfindlay the `Info` changes look good, but `Status` is not , I have been looking what OpenSearch server returns and it is indeed just opaque bytes for error or response, we may follow the same fix as Elastic folks [1] (this repository is ASF v2.0 licensed, for a reference).
> 
> [1] [elastic/elasticsearch-specification#2250](https://github.com/elastic/elasticsearch-specification/pull/2250)

Good eyes. @Bfindlay

Bfindlay:
@reta Interesting 👀 let me take a look.

Bfindlay:
Ok I see whats happening here, I will update the PR shortly.
Regarding the changes to the missing property on `Info` I'm going to split it out into its own PR for visibility so its not mixed in with this larger change.

Bfindlay:
@reta I have updated the GetTasksStatus.response to now return JsonData. This however raises the question, on [Info.status ](https://github.com/opensearch-project/opensearch-java/blob/1d99e45ead2aa86204fceb56c0e64f121d2d784b/java-client/src/main/java/org/opensearch/client/opensearch/tasks/Info.java#L74) and if it should be following the same pattern?

reta:
> This however raises the question, on [Info.status ](https://github.com/opensearch-project/opensearch-java/blob/1d99e45ead2aa86204fceb56c0e64f121d2d784b/java-client/src/main/java/org/opensearch/client/opensearch/tasks/Info.java#L74) and if it should be following the same pattern?

Thanks @Bfindlay , I think it may need to return JsonData as well

dblock (approved):

Bfindlay:
@reta PR updated to return JsonData for both properties 👍

reta:
> @reta PR updated to return JsonData for both properties 👍

@Bfindlay it looks great, thank you, I think the test case we have is a bit isolated from the real flow, could we add the integration test for it? I could certainly help you here

Bfindlay:
> > @reta PR updated to return JsonData for both properties 👍
> 
> @Bfindlay it looks great, thank you, I think the test case we have is a bit isolated from the real flow, could we add the integration test for it? I could certainly help you here

I can take a look at some iTests. If you had any specific scenario ideas let me know.

reta:
> I can take a look at some iTests. If you had any specific scenario ideas let me know.

We have some very limited coverage in `org.opensearch.client.opensearch.integTest.AbstractNodesIT` (only tasks list), would be great to add more test cases there to cover the get APIs

Bfindlay:
> > I can take a look at some iTests. If you had any specific scenario ideas let me know.
> 
> We have some very limited coverage in `org.opensearch.client.opensearch.integTest.AbstractNodesIT` (only tasks list), would be great to add more test cases there to cover the get APIs, thank you

@reta I've added an iTest for JsonData response however facing an iTest issue in the pipeline thats not happening locally. I'll probably have time to get to this sometime over the week.

java-client/src/test/java/org/opensearch/client/opensearch/integTest/OpenSearchJavaClientTestCase.java
```diff
@@ -45,7 +45,12 @@
 import org.opensearch.test.rest.OpenSearchRestTestCase;
 
 public abstract class OpenSearchJavaClientTestCase extends OpenSearchRestTestCase implements OpenSearchTransportSupport {
-    private static final List<String> systemIndices = List.of(".opensearch-observability", ".opendistro_security", ".plugins-ml-config");
+    private static final List<String> systemIndices = List.of(
+        ".opensearch-observability",
+        ".opendistro_security",
+        ".plugins-ml-config",
+        ".tasks"
```
Bfindlay, line 52: This was causing the pipeline failure 🤦 so I have added the .tasks system index here.

java-client/src/test/java/org/opensearch/client/opensearch/integTest/AbstractIndicesClientIT.java
```diff
@@ -38,6 +38,8 @@
 import org.opensearch.common.settings.Settings;
 
 public abstract class AbstractIndicesClientIT extends OpenSearchJavaClientTestCase {
+
+    @Test
```
reta, line 42: Please remove `@Test`, it should not be needed

java-client/src/test/java/org/opensearch/client/opensearch/integTest/AbstractTasksIT.java
```diff
@@ -0,0 +1,62 @@
+package org.opensearch.client.opensearch.integTest;
+
+import java.io.IOException;
+import org.junit.Test;
+import org.opensearch.client.opensearch.tasks.Status;
+
+public abstract class AbstractTasksIT extends OpenSearchJavaClientTestCase {
+
+    @Test
```
reta, line 9: Please remove `@Test`

Bfindlay:
@reta I have put this back into draft, sorry for the commit spam on this one.

After adding the iTest for tasks, it has caused side effects in many other iTests due to the `'strict'` configuration of the underlying http transport in regards to the OpenSearch warnings.
As it creates a `.tasks` system index, a lot of existing iTests are requesting all indices, or manipulating all indices which now includes the `.tasks` index.
Example: `[this request accesses system indices: [.tasks], but in a future major version, direct access to system indices will be prevented by default]` and therefore throws a `WarningFailureException`

I'll probably put this one on hold for a bit and can get back to it next week as this has expanded the scope quite a bit.
Let me know if you have any thoughts.

Bfindlay:
@reta I might not be able to finish this one anytime soon just low on time at the moment. Will see how I go but the additional work added as side effect of the tasks iTest just added more time.

reta:
> @reta I might not be able to finish this one anytime soon just low on time at the moment. Will see how I go but the additional work added as side effect of the tasks iTest just added more time.

@Bfindlay please take your time and thank you for working on this, there should be no pressure on you to finish the change as soon as possible, in any case, if you need a hand - happy to help there. Thanks again!

uriofferup:
Hello @reta  and @Bfindlay , where are we with this ? can I help to solve the issue ?.
Thank you

reta:
> Hello @reta and @Bfindlay , where are we with this ? can I help to solve the issue ?.

@uriofferup the issue seems to be stuck at the moment, if you have time to pick it up, would be highly appreciated, thank you

Xtansia:
This should be resolved in v3.0.0 of the client

Issue #749, opened 2023-11-29
[FEATURE] possibility to access HTTP status code of response


rursprung:
### Is your feature request related to a problem?
note: i'm using the `ApacheHttpClient5Transport` APIs.

i'm doing a request through `OpenSearchClient#create` because the API allows me to ensure that a document is only indexed once, as per the javadoc:
https://github.com/opensearch-project/opensearch-java/blob/c0f51d038c5d0d9b2b13235ee18c1bc555039595/java-client/src/main/java/org/opensearch/client/opensearch/OpenSearchClient.java#L329-L338

however, if the document with this ID already exists then the API throws an exception instead of returning a response. and while the `ResponseException` would offer access to the actual `Response`:
https://github.com/opensearch-project/opensearch-java/blob/c0f51d038c5d0d9b2b13235ee18c1bc555039595/java-client/src/main/java/org/opensearch/client/transport/httpclient5/ResponseException.java#L92-L94
which would also offer access to the HTTP status via `StatusLine`:
https://github.com/opensearch-project/opensearch-java/blob/c0f51d038c5d0d9b2b13235ee18c1bc555039595/java-client/src/main/java/org/opensearch/client/transport/httpclient5/Response.java#L84-L86
the problem is that `Response` is package-private, so i cannot access any of its methods (even though they're marked as `public`):
https://github.com/opensearch-project/opensearch-java/blob/c0f51d038c5d0d9b2b13235ee18c1bc555039595/java-client/src/main/java/org/opensearch/client/transport/httpclient5/Response.java#L52

### What solution would you like?
there should be a clear, canonical (and documented) way of accessing the HTTP status if such a request fails.

one possibility could be to make `Response` public (i don't understand why it isn't).

### What alternatives have you considered?
parsing the exception as a string to see if it contains "[HTTP/1.1 409 Conflict]" is at best an ugly hack as the string is not a guaranteed API

### Do you have any additional context?
i wasn't sure whether to categorise this as a feature request (because it needs a new API to be exposed), a bug report (because i consider it wrong that i can't access this information) or a question (because i might just have missed a way of getting to that information; in that case it'd be good if it could be documented as i didn't find it in this repo!).

dblock:
+1, it's pretty common to want the entire underlying HTTP request/response to be accessible. The alternative of exposing methods is also fine if we want to support a generic way to retrieve various non-transport-specific things like HTTP status codes.

rursprung:
[...]

PR #756, opened 2023-12-04
add `ResponseException#status`

rursprung:
this allows accessing the HTTP status code of the response when an API returns a `ResponseException`. this was not possible through `getResponse` since `Response` itself is package-private and thus its members (even though they are marked as `public`) are not accessible to external consumers.

as there's no test coverage for `ResponseException` i haven't added any for the new API either, as it is a trivial API and it was unclear how adding test coverage for it would fit into the existing test coverage.

solves #749

unrelated fun fact: 756 (the PR ID) is the ISO country code for my country 😄🇨🇭

### DIFF
--- a/java-client/src/main/java/org/opensearch/client/transport/httpclient5/ResponseException.java
+++ b/java-client/src/main/java/org/opensearch/client/transport/httpclient5/ResponseException.java
@@ -64,7 +64,7 @@ public final class ResponseException extends IOException {
             response.getRequestLine().getMethod(),
             response.getHost(),
             response.getRequestLine().getUri(),
-            response.getStatusLine().toString()
+            response.getStatusLine()
         );
 
         if (response.hasWarnings()) {
@@ -91,5 +91,12 @@ public final class ResponseException extends IOException {
      */
     public Response getResponse() {
         return response;
+    }
+
+    /**
+     * HTTP status code returned by OpenSearch.
+     */
+    public int status() {
+        return this.response.getStatusLine().getStatusCode();
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 
 ### Added
 - Document HTTP/2 support ([#330](https://github.com/opensearch-project/opensearch-java/pull/330))
+- Expose HTTP status code through `ResponseException#status` ([#756](https://github.com/opensearch-project/opensearch-java/pull/756))
 
 ### Dependencies
 
--- a/java-client/src/test/java/org/opensearch/client/transport/httpclient5/ResponseExceptionTest.java
+++ b/java-client/src/test/java/org/opensearch/client/transport/httpclient5/ResponseExceptionTest.java
@@ -1,0 +1,34 @@
+package org.opensearch.client.transport.httpclient5;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+
+import java.io.IOException;
+import org.apache.hc.core5.http.HttpHost;
+import org.apache.hc.core5.http.HttpStatus;
+import org.apache.hc.core5.http.HttpVersion;
+import org.apache.hc.core5.http.message.BasicClassicHttpResponse;
+import org.apache.hc.core5.http.message.RequestLine;
+import org.junit.Test;
+
+public class ResponseExceptionTest {
+
+    @Test
+    public void testStatus() throws IOException {
+        final var response = this.buildResponseException(HttpStatus.SC_BAD_REQUEST);
+        assertThat(response.status(), equalTo(HttpStatus.SC_BAD_REQUEST));
+    }
+
+    private ResponseException buildResponseException(final int statusCode) throws IOException {
+        return new ResponseException(this.buildTestResponse(statusCode));
+    }
+
+    private Response buildTestResponse(final int statusCode) {
+        return new Response(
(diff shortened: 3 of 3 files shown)


rursprung:
CC @reta & @dblock

java-client/src/main/java/org/opensearch/client/transport/httpclient5/ResponseException.java
```diff
@@ -73,7 +73,7 @@ static String buildMessage(Response response) throws IOException {
 
         HttpEntity entity = response.getEntity();
         if (entity != null) {
-            if (entity.isRepeatable() == false) {
+            if (!entity.isRepeatable()) {
```
reta, line 76: Could you please revert this change? (this is the convention we use in OpenSearch for boolean checks, for better or worth)

java-client/src/main/java/org/opensearch/client/transport/httpclient5/ResponseException.java
```diff
@@ -64,7 +64,7 @@ static String buildMessage(Response response) throws IOException {
             response.getRequestLine().getMethod(),
             response.getHost(),
             response.getRequestLine().getUri(),
-            response.getStatusLine().toString()
+            response.getStatusLine()
```
reta, line 67: Let's pass `this.response.getStatusLine().getStatusCode()`? We don't need the full `StatusLine` here ...

rursprung: hm, that was just an unrelated cleanup (the `toString` happens automatically) which my IDE nicely suggested - i can change it to `getStatusCode()`, but that'd change the behaviour - if you're ok with that i can do it (though i'll split it out into a separate commit so that it's more obvious)

java-client/src/main/java/org/opensearch/client/transport/httpclient5/ResponseException.java
```diff
@@ -92,4 +92,12 @@ static String buildMessage(Response response) throws IOException {
     public Response getResponse() {
         return response;
     }
+
+    /**
+     * Status code returned by Elasticsearch. Shortcut for
```
reta, line 97:
```suggestion
     * Status code returned by Opensearch. Shortcut for
```

java-client/src/main/java/org/opensearch/client/transport/httpclient5/ResponseException.java
reta: > hm, that was just an unrelated cleanup

Oh I see, sure, let's use the `response.getStatusLine()`, apologies for confusion

java-client/src/main/java/org/opensearch/client/transport/httpclient5/ResponseException.java
dblock: capitalize OpenSearch accordingly.=

dblock (changes requested):
Code looks good! Does need a test even if it's just to exercise the code to make sure it really works and we don't accidentally remove it.

dblock:
> unrelated fun fact: 756 (the PR ID) is the ISO country code for my country 😄🇨🇭

Hopp Schwiiz!!

java-client/src/main/java/org/opensearch/client/transport/httpclient5/ResponseException.java
```diff
@@ -92,4 +92,12 @@ static String buildMessage(Response response) throws IOException {
...
+     * {@code response().status()}.
+     */
+    public int status() {
```
dblock, line 100: Q: Should it be `status` or `getStatus`? What's the convention?

java-client/src/main/java/org/opensearch/client/transport/httpclient5/ResponseException.java
rursprung: that was a copy & paste thing (it still says `Elasticsearch` on `OpenSearchException` where i copied this from and i hadn't seen it). i've checked and there are more such places, thus i've now raised #757 to get them all fixed (i'll fix it here as it's new code but won't touch `OpenSearchException` as part of this PR)

java-client/src/main/java/org/opensearch/client/transport/httpclient5/ResponseException.java
rursprung: i took this 1:1 from `OpenSearchException`, presuming that this was the accepted convention. happy to change it to `getStatus`

rursprung: i've left it at `status` for now pending your feedback

rursprung:
> Does need a test even if it's just to exercise the code to make sure it really works and we don't accidentally remove it.

ok, i've added a small test for it (though i skipped on testing the existing things as `buildMessage` is just creating a human-readable string, so parsing it in a test is IMHO adding little benefit and `getResponse` is package-private)

reta (approved):

rursprung:
the test failure is a false reject due to infrastructure issues:
```
Downloading https://services.gradle.org/distributions/gradle-8.4-bin.zip

Error: Exception in thread "main" java.net.SocketException: Software caused connection abort: socket write error
```

can you re-trigger it? (i don't have the rights for that)

dblock:
Thanks for doing this @rursprung!

dblock:
@rursprung can you please manually backport it to 2.x so we can ship it sooner than later?

rursprung:
> @rursprung can you please manually backport it to 2.x so we can ship it sooner than later?

done. thanks!

Issue #716, opened 2023-11-07
use _toQuery() when combining queries


channel-dante:
### Is your feature request related to a problem?
The _toQuery() method is often used to combine multiple queries as below.
```
    builder.query(BoolQuery.of(boolQueryBuilder ->
        boolQueryBuilder
            .must(must.build()._toQuery())
            .filter(should.build()._toQuery())
    )._toQuery());
```

### What solution would you like?
Methods starting with an underscore seem like naming conventions that should only be used internally. Can I continue to use this method as above? If that's the case, could you consider removing the underscore from the method naming like (.toQuery())?

If I am using the method in the wrong way, what is the correct way to combine queries like above?

PR #760, opened 2023-12-07
rename _toQuery() to toQuery()

channel-dante:
### Description
- Add toQuery() method with the same logic as _toQuery() in Query and QueryVariant
- add deprecated annotation on _toQuery() method

### Issues Resolved
- https://github.com/opensearch-project/opensearch-java/issues/716

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,13 +36,13 @@ This section is for maintaining a changelog for all breaking changes for the cli
 ### Added
 - Added support for icu_collation_keyword type ([#725](https://github.com/opensearch-project/opensearch-java/pull/725))
 - Added support for flat_object field property ([#735](https://github.com/opensearch-project/opensearch-java/pull/735))
-
+- Added toQuery method in Query and QueryVariant ([#760](https://github.com/opensearch-project/opensearch-java/pull/760)
 ### Dependencies
 
 ### Changed
 
 ### Deprecated
-
+- deprecated _toQuery() in Query and QueryVariant ([#760](https://github.com/opensearch-project/opensearch-java/pull/760)
 ### Removed
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


channel-dante:
Are there any plans to support version 1.x once this code is merged? If so, please tell me the appropriate branch and I will do the same work there.

CHANGELOG.md
```diff
@@ -36,13 +36,13 @@ This section is for maintaining a changelog for all breaking changes for the cli
...
 - Added support for flat_object field property ([#735](https://github.com/opensearch-project/opensearch-java/pull/735))
-
+- Added toQuery method in Query and QueryVariant ([#760](https://github.com/opensearch-project/opensearch-java/pull/760)
 ### Dependencies
 
 ### Changed
 
 ### Deprecated
-
+- deprecated _toQuery() in Query and QueryVariant ([#760](https://github.com/opensearch-project/opensearch-java/pull/760)
```
dblock, line 45: Capitalize `Deprecated` and maybe quote `_toQuery` and other types (feels like we should be doing it, nbd).

java-client/src/main/java/org/opensearch/client/opensearch/_types/query_dsl/QueryVariant.java
```diff
@@ -39,8 +39,12 @@ public interface QueryVariant {
 
     Query.Kind _queryKind();
 
+    @Deprecated
```
dblock, line 42: Should we implement this one in terms of the other so we don't duplicate code? The JIT will optimize it away anyway.

channel-dante: I've seen it being used elsewhere in code from this repository. The intention was to provide reference for internal code writers.

PR #767, opened 2023-12-12
Restore support for Java 8

reta:
### Description
Restore support for Java 8

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-java/issues/156


dblock:
YES! Let's add CI.

reta:
> YES! Let's add CI.

Thanks @dblock ! Working on it :)

reta:
@dblock @VachaShah ended up with the following change:
 - the `opensearch-java` baseline is back to 1.8 (since RestClient is also back to 1.8)
 - the integration tests heavily rely on OpenSearch core test framework (it could not be switched to JDK 8), so this tests need JDK-11 baseline, the solution here (one of) is to split tests into `src/test/java` and `src/test/java11`
 - the Gradle build uses `grgit-gradle` 5.x that also needs JDK-11 baseline so we cannot build with JDK 8 sadly (it explains why GA/CI has no JDK 8 in matrix)

dblock:
> * the Gradle build uses `grgit-gradle` 5.x that also needs JDK-11 baseline so we cannot build with JDK 8 sadly (it explains why GA/CI has no JDK 8 in matrix)

I think this should be a blocker. Can we downgrade? We need at least _some_ CI on JDK8, anything even a basic integration test would be enough.

dblock (dismissed):
I would be ok we merge it with an issue on re-adding JDK8 to CI, unless it's doable with the downgrade of the dependent library.

reta:
> I think this should be a blocker. Can we downgrade? We need at least _some_ CI on JDK8, anything even a basic integration test would be enough.

We could downgrade but to unsupported version (which is not good), alliteratively we could try to use other way to get the git details needed, I will look further

dblock:
@VachaShah or @Bukhtawar ?

VachaShah (approved):
I am fine with merging this, thank you @reta for adding this! A few TODOs:
1. We can add the restored support details to the client documentation. For example, here https://github.com/opensearch-project/opensearch-java/blob/main/DEVELOPER_GUIDE.md.
2. As @dblock mentioned, CI with JDK-8.

reta:
> As @dblock mentioned, CI with JDK-8.

Was trying to pull it off, but there are 3 blockers (previously I thought there is only 1 but it turns out Gradle reports issues one at a time):
 1. `grgit-gradle` - this one could be rewritten
 2. `spotless` - we need at least 6.23.x for JDK-21 but it seems like 6.11.x was the last one that supported JDK-8
 3. OpenSearch build tools (OpenSearch Gradle plugins) - needs JDK-11

So I am not sure what could we do here besides enforcing the target / source level

reta:
> 1. . For example, here https://github.com/opensearch-project/opensearch-java/blob/main/DEVELOPER_GUIDE.md.

Thanks @VachaShah, we have this documented in the COMPATIBILITY.md, so updated the versions there

dblock:
> So I am not sure what could we do here besides enforcing the target / source level

Open a GitHub issue?

dblock (approved):

reta:
> Open a GitHub issue?

For `grgit` and `spotless` decisions to move baselines to JDK-11 were deliberate (I could link the related issues if you are curious), so I don't think we should be opening an issue here. For OpenSearch build tools we probably could try to go with JDK 8 baseline but that would not solve the problem at large

VachaShah (approved):

dblock:
Ok, I merged this, will need a manual backport - let's try to release a version that restored JDK8 support?

Issue #755, opened 2023-12-04
[BUG] SearchAfter only allows List<Strings> as type.


ahmadSibai:
### What is the bug?
_OpenSearch search request `search_after` must accept any type of variables, in order to be able to paginate using the search_after feature. 
The type defined in `SearchRequest.java` only accepts `List<String>`. This prevents any other non-String sort value from a previous query to be used in the proper type, causing the search query to fail.

The SearchResponse also casts any value from the hits.sort() to String.

### How can one reproduce the bug?
1. Create a query with a `sort` value on a field that is not of type String.
2. Collect the `sort` value of your last hit
3. Create a copy of your first query, but add a `searchAfter` key with value of the List collected in step 2.
4. Notice how these values would have to be of type `String`. Try calling the query and notice how the shards will fail, returning 0 hits.

### What is the expected behavior?
Expected behaviour is for the `sort` value to be in the original type, and for `SearchAfter` to be capable of being of that Type.

### What is your host/environment?
AOSS

### Do you have any screenshots?
No, but you can see [here](https://github.com/opensearch-project/opensearch-java/blob/261cab4cf79e176b0892cfb44fad7c8cfaf44204/java-client/src/main/java/org/opensearch/client/opensearch/core/SearchRequest.java?t=7#L1175C10-L1176C42)

### Do you have any additional context?
_Add any other context about the problem._

ahmadSibai:
Opensearch documentation states that the `search_after` field should have the same values and types of the `sort` field, which can be a non-String. Please see reference [here](https://opensearch.org/docs/latest/search-plugins/searching-data/paginate/#the-search_after-parameter). The examples clearly show how it can be a list of an integer and String.
```
"search_after": [ 1, "32635"],
```

dblock:
Thanks. If you have time, write a test that doesn't compile and maybe see if it's easy to fix?

channel-dante:
close this issue? This issue has been resolved. However, since it is not backward compatible, it will be applied to version 3.x.

PR #769, opened 2023-12-13
Changed SearchAfter field type to FieldValue instead of String

channel-dante:
### Description
- Changed SearchAfter of SearchRequest type to FieldValue instead of String
- Written in the same way as other models receive various types

### Issues Resolved
https://github.com/opensearch-project/opensearch-java/issues/755

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 
 ### Changed
 - Migrate client transports to Apache HttpClient / Core 5.x ([#246](https://github.com/opensearch-project/opensearch-java/pull/246))
+- Changed SearchAfter of SearchRequest type to FieldValue instead of String ([#769](https://github.com/opensearch-project/opensearch-java/pull/769))
 
 ### Deprecated
 - Deprecate RestClientTransport ([#536](https://github.com/opensearch-project/opensearch-java/pull/536))
@@ -42,7 +43,6 @@ This section is for maintaining a changelog for all breaking changes for the cli
 ### Dependencies
 
 ### Changed
-- Changed SearchAfter of SearchRequest type to FieldValue instead of String
 
 ### Deprecated
 - Deprecated "_toQuery()" in Query and QueryVariant ([#760](https://github.com/opensearch-project/opensearch-java/pull/760)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 ### Dependencies
 
 ### Changed
+- Changed SearchAfter of SearchRequest type to FieldValue instead of String
 
 ### Deprecated
 - Deprecated "_toQuery()" in Query and QueryVariant ([#760](https://github.com/opensearch-project/opensearch-java/pull/760)
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -2,5 +2,19 @@
 
 ## [UPGRADING 2.x to 3.0]
 ### SearchAfter of SearchRequest type
-- Changed SearchAfter of SearchRequest type to FieldValue instead of String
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 ### Dependencies
 
 ### Changed
+- Changed SearchAfter of SearchRequest type to FieldValue instead of String
```
dblock, line 45: Add a link to the PR like other line-items.

dblock (changes requested):
So, existing code that was `.searchAfter("string")` worked, right? Therefore after this change that code would no longer build? 

This would make the change backwards incompatible, which would require a major version bump. Can we implement it in a way that doesn't break backwards compat?

CHANGELOG.md
channel-dante: Since overload cannot distinguish between List<FieldValue> and List<String>, I think you have no choice but to make backwards incompatible. It can be implemented as shown below, but it seems that there will be a side effect of allowing unintended Objects.

And by adding this overloading, the issue of not being able to find the targeted method in the deserializer also occurs.

image

image

dblock: @reta any ideas what's best here?

channel-dante: @reta Can you give me any comment here? 🥲

reta: Sorry folks, overlooked this pull request, we could add the varargs overload:
```
public final Builder searchAfter(String value, String... values) {
```

but indeed due to erasure, we may only keep  `List<String>` for backward compatibility in 2.x (and convert it to `FielValue` internally) but use  `List<FieldValue>` in 3.x, what do you think?

channel-dante: @reta Converting to FieldValue only internally doesn't seem to make sense. So, how about applying this PR to 3.x? If I have to do that, should I move the branch to 3.x?

dblock:
I think this PR looks good, it's a breaking change for 3.0. For a backport to 2.x your plan LGTM.

How about we start introducing an UPGRADING.md that describes all things that change in a way that is backwards incompatible? @channel-dante Care to add an UPGRADING.md in the same format as https://github.com/opensearch-project/opensearch-py/blob/main/UPGRADING.md? And rebase. 

Thank you for hanging in here with us!

CHANGELOG.md
reta: Sure @channel-dante , the `main` is 3.x so no need to change anything

UPGRADING.md
```diff
@@ -0,0 +1,6 @@
+# UPGRADING
+
+## [UPGRADING 2.x to 3.0]
+### SearchAfter of SearchRequest type
+- Changed SearchAfter of SearchRequest type to FieldValue instead of String
+- Consider using `FieldValue.of("")` to make string type values compatible.
```
reta, line 6:
```suggestion
- Consider using `FieldValue.of` to make string type values compatible.
```
```suggestion
- Consider using `FieldValue.of("")` to make string type values compatible.
```

reta (dismissed):

UPGRADING.md
dblock, line 6: Since this is the first time we're doing this, can we try to be extra helpful to the users?

I think we need something like https://github.com/ruby-grape/grape/blob/master/UPGRADING.md#recognizing-path, which has 1) Text that explains the change, a link to the PR (this PR), a before and after.

reta (approved):

channel-dante:
@dblock Can you please confirm about this pr?

dblock (approved):

dblock:
Merged, thanks @channel-dante. Want to do a manual backport for a backwards compatible version?

channel-dante:
@dblock I didn't understand what kind of work it was. Can you give a little explanation?

dblock:
Nevermind, I re-read the comments on bcw above. Thanks.

Issue #684, opened 2023-10-24
Analyzer.Kind  to support all languages for analyzers


brentam:
### Is your feature request related to a problem?
When trying to build a client using json files, when deserialising the settings, I get the error 
**`Missing required property 'Builder.<variant kind>'`**.
This is an example of a settings that will fail:
`{
  "index": {
    "analysis": {
      "analyzer": {
        "english_analyzer": {
          "type": "english",
          "char_filter": [
            "html_strip"
          ],
          "tokenizer": "standard"
        }
      }
    }
  }
}`

The problem is that Analyzer.Kind does not support most languages. Actually only Dutch is listed in the enum.
[Analyzer source code](
https://github.com/opensearch-project/opensearch-java/blob/main/java-client/src/main/java/org/opensearch/client/opensearch/_types/analysis/Analyzer.java)

### The problem is that
Not able to build the client using analyzers type for languages other than Dutch

### What solution would you like?
Add all other languages listed  [here](https://opensearch.org/docs/latest/analyzers/language-analyzers)

### What alternatives have you considered?
Using a different client.

### Do you have any additional context?
All the necessary info is already mentioned above.
here is a sample code I am using to build the index

`    fun createIndex(indexName: String) {
        val mapper: JsonpMapper = client._transport().jsonpMapper()
        val mappingsParser = parserForFile(mapper, "index/mappings.json")
        val settingsParser = parserForFile(mapper, "index/settings.json")

        val createIndexRequest = CreateIndexRequest.Builder().index(indexName)
            .mappings(TypeMapping._DESERIALIZER.deserialize(mappingsParser, mapper))
            .settings(IndexSettings._DESERIALIZER.deserialize(settingsParser, mapper))
            .build()

        client.indices().create(createIndexRequest)
    }`

dblock:
@brentam Want to try to contribute a fix for it? Feels like at least you should be able to specify an arbitrary analyzer without constantly having to expand the enumeration.

brentam:
@dblock hi, I will try to contribute when I have some time, I probably will add all the missing languages. That would be  enough for this case.

brentam:
Created PR: https://github.com/opensearch-project/opensearch-java/pull/779

LiquidTheDangerous:
Still active?

LiquidTheDangerous:
Is there any workaround?

dblock:
[...]

PR #779, opened 2023-12-21
Added support for 33 language analyzers

brentam:
### Description
When trying to build `IndexSettings`. There is no way to build default analyzers for languages supported in opensearch.
The only supported language is Dutch at the moment.
This is also a problem when deserializing `IndexSettings` from json. It will fail for any language other than Dutch.
See this issue [684](https://github.com/opensearch-project/opensearch-java/issues/684)

In this PR, support for creating language Analyzers was added for these languages:
_arabic, armenian, basque, bengali, brazilian, bulgarian, catalan, czech, danish, english, estonian, finnish, french, galician, german, greek, hindi, hungarian, indonesian, irish, italian, latvian, lithuanian, norwegian, persian, portuguese, romanian, russian, sorani, spanish, swedish, turkish, and thai_

### Issues Resolved
Will close this issue: [684](https://github.com/opensearch-project/opensearch-java/issues/684)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 ### Added
 - Document HTTP/2 support ([#330](https://github.com/opensearch-project/opensearch-java/pull/330))
 - Expose HTTP status code through `ResponseException#status` ([#756](https://github.com/opensearch-project/opensearch-java/pull/756))
+- Added support for 33 new language analyzers (only Dutch existed previously) ([#000](https://github.com/opensearch-project/opensearch-java/pull/ooo))
 
 ### Dependencies
 
--- a/java-client/src/main/java/org/opensearch/client/opensearch/_types/analysis/Analyzer.java
+++ b/java-client/src/main/java/org/opensearch/client/opensearch/_types/analysis/Analyzer.java
@@ -62,8 +62,6 @@ public class Analyzer implements TaggedUnion<Analyzer.Kind, AnalyzerVariant>, Js
     public enum Kind implements JsonEnum {
         Custom("custom"),
 
-        Dutch("dutch"),
-
         Fingerprint("fingerprint"),
 
         IcuAnalyzer("icu_analyzer"),
@@ -91,6 +89,74 @@ public class Analyzer implements TaggedUnion<Analyzer.Kind, AnalyzerVariant>, Js
         Smartcn("smartcn"),
 
         Cjk("cjk"),
+
+        Arabic("arabic"),
+
+        Armenian("armenian"),
+
+        Basque("basque"),
+
+        Bengali("bengali"),
+
+        Brazilian("brazilian"),
+
+        Bulgarian("bulgarian"),
+
+        Catalan("catalan"),
+
+        Czech("czech"),
+
+        Danish("danish"),
+
+        Dutch("dutch"),
+
+        English("english"),
+
+        Estonian("estonian"),
+
+        Finnish("finnish"),
+
+        French("french"),
+
+        Galician("galician"),
+
+        German("german"),
+
+        Greek("greek"),
+
+        Hindi("hindi"),
+
+        Hungarian("hungarian"),
+
+        Indonesian("indonesian"),
+
+        Irish("irish"),
+
+        Italian("italian"),
+
+        Latvian("latvian"),
+
+        Lithuanian("lithuanian"),
+
+        Norwegian("norwegian"),
+
+        Persian("persian"),
+
+        Portuguese("portuguese"),
+
+        Romanian("romanian"),
+
+        Russian("russian"),
+
+        Sorani("sorani"),
+
+        Spanish("spanish"),
+
+        Swedish("swedish"),
+
+        Turkish("turkish"),
+
+        Thai("thai"),
 
         ;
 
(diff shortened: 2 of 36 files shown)


dblock:
LGTM! Fix spotless & friends?

Would love to have a sample/doc for this, https://github.com/opensearch-project/opensearch-java/tree/main/samples and/or https://github.com/opensearch-project/opensearch-java/tree/main/guides.

brentam:
> LGTM! Fix spotless & friends?
> 
> Would love to have a sample/doc for this, https://github.com/opensearch-project/opensearch-java/tree/main/samples and/or https://github.com/opensearch-project/opensearch-java/tree/main/guides.

Yes. I ran the spotlessApply command and fixed the formatting.
I will see if I can add to the samples documents. Probably not this week though

reta (changes requested):
@brentam my apologies but this is not how it is already implemented in `opensearch-java`: the language specific analyzers are the combination of `org.opensearch.client.opensearch._types.analysis.LanguageAnalyzer` and `org.opensearch.client.opensearch._types.analysis.Language` enumeration

brentam:
> @brentam my apologies but this is not how it is already implemented in `opensearch-java`: the language specific analyzers are the combination of `org.opensearch.client.opensearch._types.analysis.LanguageAnalyzer` and `org.opensearch.client.opensearch._types.analysis.Language` enumeration

@reta 
I have a couple of points:
**first:**
The test below for creating an IndexSettings from a valid index  json is using the "existing" DutchAnalyzer class, it does not touch the Language enumeration at any point.


```
   @Test
    public void testParsingAnalyzer() {

        JsonpMapper mapper = new JsonbJsonpMapper();
        String json = "{ \"index\": { \"analysis\": { \"analyzer\": { \"some_analyzer\": { \"type\": \"dutch\", \"char_filter\": [ \"html_strip\" ], \"tokenizer\": \"standard\" } } } } } ";

        JsonParser parser = mapper.jsonProvider().createParser(new StringReader(json));
        IndexSettings indexSettings = IndexSettings._DESERIALIZER.deserialize(parser, mapper);
        assertTrue(true);
    }
```
if we serialize to json we get
`{"index":{"analysis":{"analyzer":{"some_analyzer":{"type":"english"}}}}}`

It is missing the char_filter but this is a discussion to another time.
The point is , this is a valid analyzer definition in opensearch.

**second:**
When creating an IndexSettings using the LanguageAnalyzer Builder + Language. 
It generates an invalid Json
This code:
```
        Map<String, Analyzer> map = new HashMap<>();
        map.put("some_analyzer",
                LanguageAnalyzer.of(l -> l.language(Language.German).stemExclusion(new ArrayList<>()))
                ._toAnalyzer());
        IndexSettings settings = IndexSettings.of(
                it -> it.analysis(a ->
                        a.analyzer(map))
        );
```
Generates this json
`{"index":{"analysis":{"analyzer":{"some_analyzer":{"type":"language","language":"German","stem_exclusion":[]}}}}}`

This will fail in opensearch since "type": "language" is not valid:


you can test in the dashboard console using this command:

```
PUT test_idx1
{
  "mappings": {
    "dynamic": "strict",
    "properties": {
      "title": {
        "doc_values": false,
        "type": "text",
        "analyzer": "some_analyzer"
      }
    }
  },
  "settings": {
    "index": {
      "analysis": {
        "analyzer": {
          "some_analyzer": {

            "type": "language",
            "language": "Gerdfasdfmana",
            "stem_exclusion": []
          }
        }
      }
    }
  }
}
```

brentam:
@reta 
this is a correct way to specify a language analyzer
```
        "analyzer": {
          "some_analyzer": {
            "type": "german"

          }
        }
```
In opensearch the language type is checked before the index is created.

I already pointed out that the "type" : "language" generated by the LanguageAnalyzer buider is not valid.
Now , this configuration will also be accepted in Opensearch and the index will be created
```
          "some_analyzer": {
            "type": "custom",
            "language": "AlienLanguage",
            "stem_exclusion": []
          }
```
Note the "language" : "AlienLanguage" is  accepted, so this type of analyzer is not actually creating an analyzer that uses the supported language analyzers.

Therefore, the json the LanguageAnalyzer  Builder + Language is generating at the moment is not correct (was this tested?).

We should be able to generate something like if we want to.
```
          "english_analyzer": {
            "type": "english",
            "char_filter": [
              "html_strip"
            ],
            "tokenizer": "standard"
          },
```

brentam:
Now what is the fix here? 
The DutchAnalyzer class (and the newly created classes) seems to be used to deserialize the jsons with "type": "dutch".

The LanguageAnalyzer builder is not generating the correct analyzer for languages for the settings clause...

It seems to me the new code will not cause any harm and at least allow the deserization of a correct language analyzer setting.

What about the LanguageAnalyzer Builder? I am not sure how many tests we have for them, it seems that indexes created with language analyzers using that builder will always fail creation.

reta:
@brentam 

My understanding is that `LanguageAnalyzer` is used for languages that are supported by OpenSearch (hence the enumeration), whereas the `custom` analyzer is used in case you need any kind of customization.

> LanguageAnalyzer is generating a completely different analyzer than DutchAnalyzer ( Remember, DutchAnalyzer already existed, I just added other analyzers for other languages: GermanAnalyzer, ItalianAnalyzer, etc...).

I sadly don't have any insights, but I believe it appeared exactly for the same cause as yours (no way to specify language analyzer). I am not claiming that the `LanguageAnalyzer`  + `Language` actually works, but I suspect it should be serving this purpose, referring to your example:

```
 Map<String, Analyzer> map = new HashMap<>();
        map.put("some_analyzer",
                LanguageAnalyzer.of(l -> l.language(Language.German).stemExclusion(new ArrayList<>()))
                ._toAnalyzer());
        IndexSettings settings = IndexSettings.of(
                it -> it.analysis(a ->
                        a.analyzer(map))
        );
```

Should roughly be serialized as (limiting to only `analyzer` part):

```
 "analyzer": {
        "some_analyzer": {
            "type": "german",
            "stem_exclusion": []
        }
 }
```

> Therefore, I think the current changes are valid.

I am not saying that your change is not valid. I am pointing out only the fact that we already do have a mechanism to use language specific analyzers which apparently does not work - that to me is the issue that has to be fixed (vs introducing yet another way to achieve the desired goal).

brentam:
@reta 
thank you so much for your response.

So the existing DutchAnalyzer was an obsolete and should have been removed with the introduction of the LanguageAnalyzer...
In Any case, I can remove my changes and modify the  LanguageAnalyzer+Language to construct the "correct" json.
Just please confirm that you are ok with this change and I will proceed with the changes.

brentam:
The CharFilter.Builder seems to be incomplete too.
There is no way to construct an array of names, like how is expected here 
```
          "english_analyzer": {
            "type": "english",
            "char_filter": [
              "html_strip"
            ],
            "tokenizer": "standard"
          }
```

It only allows  a map
` .charFilter("some_char_filter", CharFilter.of(c -> c.name("html_strip")))`
that will result in
`{"some_char_filter":"html_strip"}`

Now, I was looking at the apis related to the Charfilter builders. There is a lot of classes and hard to understand the right ones to use, so refactoring that will be a pain...

I suggest the simple fix would be having a "reference" method in the builder...
We would set it like this
` CharFilter.of(c -> c.references(new String[]{"html_strip","some_custom_charset_defined_elsewhere"}))`

reta:
Thanks @brentam 

> As I said, it seems to me that design using the Per Language Analyzer seems the correct way since it follows the current design,
> The LanguageAnalyzer+Language will result in some code that deviates from the current design...

So the current client model reflects the builtin analyzer types as they are described in the documentation [1] (sadly I have to refer to the Elasticsearch 7.10 here as OpenSearch documentation is not yet complete). In there, all language specific analyzers falls under "Language analyzers" category. I think if LanguageAnalyzer+Language are not supposed to configure language specific analyzers, than I honestly don't know what the purpose of those really is.

[1] https://www.elastic.co/guide/en/elasticsearch/reference/7.10/analysis-analyzers.html

> If using the LanguageAnalyzer we can have two approaches.
> 1.  Keep the Language value. We will have to modify the serialization to not expect/generate the "language" value in the json
> 2. Have all the 33 new Languages values added to the enum and then mapping them to the LanguageAnalyzer later during deserialization/Serialization?

I was thinking about this approach: "language" is gone (your 1st point) and "type" becomes the value of language (which  represents all languages supported by OpenSearch since it is an enumeration, covers your 2nd point). I think it would fix two issues at once: a) allow to specify language analyzer b) fix incorrect (de)serialization. It is also quite a small change from the scope perspective.

What do you think about this?

brentam:
@reta  
If using the LanguageAnalyzer, the CjkAnalyzer will have to be deleted too. That analyzer was added in August of 2023.
If follows the pattern of Analyzer per language, the same way I implemented the other 33 languages.

brentam:
@reta 
here is the PR with the quick changes to the Analyzer/LanguageAnalyzer changes
[788](https://github.com/opensearch-project/opensearch-java/pull/788)

brentam:
> Thanks @brentam
> 
> > As I said, it seems to me that design using the Per Language Analyzer seems the correct way since it follows the current design,
> > The LanguageAnalyzer+Language will result in some code that deviates from the current design...
> 
> So the current client model reflects the builtin analyzer types as they are described in the documentation [1] (sadly I have to refer to the Elasticsearch 7.10 here as OpenSearch documentation is not yet complete). In there, all language specific analyzers falls under "Language analyzers" category. I think if LanguageAnalyzer+Language are not supposed to configure language specific analyzers, than I honestly don't know what the purpose of those really is.
> 
> [1] https://www.elastic.co/guide/en/elasticsearch/reference/7.10/analysis-analyzers.html
> 
> > If using the LanguageAnalyzer we can have two approaches.
> > 
> > 1. Keep the Language value. We will have to modify the serialization to not expect/generate the "language" value in the json
> > 2. Have all the 33 new Languages values added to the enum and then mapping them to the LanguageAnalyzer later during deserialization/Serialization?
> 
> I was thinking about this approach: "language" is gone (your 1st point) and "type" becomes the value of language (which represents all languages supported by OpenSearch since it is an enumeration, covers your 2nd point). I think it would fix two issues at once: a) allow to specify language analyzer b) fix incorrect (de)serialization. It is also quite a small change from the scope perspective.
> 
> What do you think about this?

@reta 

We cannot keep the Language("language") value,  the "language" is what is expected from the type in the json
i.e: {"type":"language"} which is wrong, so basically I think we will have to keep  one value per language; Dutch("dutch"), English("english"); etc...
During the Deserialization we transform all those values to the a LanguageAnalyzer.
Please check the new [PR](https://github.com/opensearch-project/opensearch-java/pull/788) with the draft for the changes and let me know what you think...

AshwiniKeshavarao:
Hi, We are facing same issue. This PR is in open state. When will this be merged?

reta:
> Hi, We are facing same issue. This PR is in open state. When will this be merged?

I believe the submitter is not working on the pull request anymore :(

PR #788, opened 2023-12-31
discussing changes to LanguageAnalyzer

brentam:
### Description
This a temporary PR to discuss changes to the LanguageAnalyzer neede to correct some issues.
The original PR is here [#779](https://github.com/opensearch-project/opensearch-java/pull/779)


java-client/src/main/java/org/opensearch/client/opensearch/_types/analysis/LanguageAnalyzer.java
```diff
@@ -68,7 +69,7 @@ private LanguageAnalyzer(Builder builder) {
 
         this.version = builder.version;
         this.language = ApiTypeHelper.requireNonNull(builder.language, this, "language");
-        this.stemExclusion = ApiTypeHelper.unmodifiableRequired(builder.stemExclusion, this, "stemExclusion");
+        this.stemExclusion = builder.stemExclusion; // ApiTypeHelper.unmodifiableRequired(builder.stemExclusion, this, "stemExclusion");
```
brentam, line 72: this should not be mandatory

java-client/src/main/java/org/opensearch/client/opensearch/_types/analysis/Analyzer.java
```diff
@@ -589,8 +592,11 @@ public Analyzer build() {
 
     protected static void setupAnalyzerDeserializer(ObjectDeserializer<Builder> op) {
 
+        for (Language value : Language.values()) {
+            op.add(Builder::language, LanguageAnalyzer._DESERIALIZER, value.jsonValue().toLowerCase());
+            //TODO should we lowercase in the Language Enum?
```
brentam, line 597: Language jsonValue has to match the Kind jsonValue
So we set the deserializers for all the languages here

java-client/src/main/java/org/opensearch/client/opensearch/_types/analysis/Language.java
```diff
@@ -35,76 +35,77 @@
...
 import org.opensearch.client.json.JsonpDeserializable;
 
+import java.util.Arrays;
+
 @JsonpDeserializable
 public enum Language implements JsonEnum {
-    Arabic("Arabic"),
-
-    Armenian("Armenian"),
+    Arabic("arabic"),
```
brentam, line 42: the json value has to be lowercase. to match the values Opensearch expects

java-client/src/main/java/org/opensearch/client/opensearch/_types/analysis/LanguageAnalyzer.java
```diff
@@ -83,7 +84,7 @@ public static LanguageAnalyzer of(Function<Builder, ObjectBuilder<LanguageAnalyz
      */
     @Override
     public Analyzer.Kind _analyzerKind() {
-        return Analyzer.Kind.Language;
+        return Analyzer.Kind.valueOf(this.language.name());
     }
```
brentam, line 88: We have to determine the Analyzer.Kind at runtime....

java-client/src/main/java/org/opensearch/client/opensearch/_types/analysis/CjkAnalyzer.java
reta, line 1: I believe the `Cjk` was added because it does not support `stem_exclusion` , not every language analyzer supports it actually (per documentation).

java-client/src/main/java/org/opensearch/client/opensearch/_types/analysis/LanguageAnalyzer.java
reta: It is supported by most analyzers but not all

java-client/src/test/java/org/opensearch/client/opensearch/experiments/ParsingTests.java
```diff
@@ -157,20 +157,21 @@ public void testSmartcn_StopFilter() {
...
-
-        String str = toJson(analyzer);
-        assertEquals("{\"type\":\"cjk\",\"stopwords\":[\"a\",\"b\",\"c\"],\"stopwords_path\":\"path\"}", str);
-
-        Analyzer analyzer2 = fromJson(str, Analyzer._DESERIALIZER);
-        assertTrue(analyzer2.isCjk());
-        assertEquals(analyzer.cjk().stopwords(), analyzer2.cjk().stopwords());
-        assertEquals(analyzer.cjk().stopwordsPath(), analyzer2.cjk().stopwordsPath());
-    }
+//   temporary commenting this test until we decide if we keep the CjkAnalyser
```
reta, line 160: I think we should keep `CjkAnalyser` since it does not support `stem_exclusion`

brentam: @reta  
sorry if the comment was not clear. What I meant is that if we change the LanguageAnalyzer, the CJkAnalyzer will be deleted, same as DutchAnalyzer, since all languages would be constructed using the LanguageAnalyzer+Language.

brentam: I am usure what you mean.
The whole point of the discussion was that maybe we should use the LanguageAnalyzer pattern to build the language analyzers. If we have one exception for Cjk, we will keep supporting two patterns. In that case I rather keep all the 33 analyzers are created in the the other PR, since the LanguageAnalyzer atm is not correct and the Analyzer Per language pattern is working fine..

reta: @brentam 

> The whole point of the discussion was that maybe we should use the LanguageAnalyzer pattern to build the language analyzers.

This is correct, we though need to acknowledge that not all `LanguageAnalyzer` are supporting the same customization capabilities - CJK does not support `stem_exclusion` (may be others as well but this one for sure). Making `stem_exclusion` won't solve the problem - the server side will (or at least, should) reject the CJK analyzer with `stem_exclusion`.  

Looking how we could accommodate that.

java-client/src/main/java/org/opensearch/client/opensearch/_types/analysis/LanguageAnalyzer.java
```diff
@@ -83,7 +84,7 @@ public static LanguageAnalyzer of(Function<Builder, ObjectBuilder<LanguageAnalyz
      */
     @Override
     public Analyzer.Kind _analyzerKind() {
-        return Analyzer.Kind.Language;
+        return Analyzer.Kind.valueOf(this.language.name());
```
reta, line 87: I think we don't need to change analyzer kind here, we need to change serialization (type -> language) and deserialization (`Analyzer` class):

```
        for (final Language language: Language.values()) {
            op.add(Builder::language, LanguageAnalyzer._DESERIALIZER, language.jsonValue());
        }
```

java-client/src/main/java/org/opensearch/client/opensearch/_types/analysis/LanguageAnalyzer.java
```diff
@@ -128,21 +129,19 @@ public final String stopwordsPath() {
...
-        serializeInternal(generator, mapper);
+        serializeInternal(generator);
         generator.writeEnd();
     }
 
-    protected void serializeInternal(JsonGenerator generator, JsonpMapper mapper) {
+    protected void serializeInternal(JsonGenerator generator) {
 
-        generator.write("type", "language");
+        generator.write("type", this._analyzerKind().jsonValue());
```
reta, line 138: :+1:

java-client/src/main/java/org/opensearch/client/opensearch/_types/analysis/LanguageAnalyzer.java
reta: but it is reasonable to make it optional for those that supports this property, I agree

java-client/src/main/java/org/opensearch/client/opensearch/_types/analysis/Analyzer.java
```diff
@@ -589,8 +592,11 @@ public Analyzer build() {
 
     protected static void setupAnalyzerDeserializer(ObjectDeserializer<Builder> op) {
 
+        for (Language value : Language.values()) {
+            op.add(Builder::language, LanguageAnalyzer._DESERIALIZER, value.jsonValue().toLowerCase());
```
reta, line 596: You already did that, right? 

```suggestion
            op.add(Builder::language, LanguageAnalyzer._DESERIALIZER, value.jsonValue());
```

java-client/src/test/java/org/opensearch/client/opensearch/experiments/ParsingTests.java
```diff
@@ -157,20 +157,21 @@ public void testSmartcn_StopFilter() {
...
+//        assertTrue(analyzer.isCjk());
+//
+//        String str = toJson(analyzer);
+//        assertEquals("{\"type\":\"cjk\",\"stopwords\":[\"a\",\"b\",\"c\"],\"stopwords_path\":\"path\"}", str);
+//
+//        Analyzer analyzer2 = fromJson(str, Analyzer._DESERIALIZER);
+//        assertTrue(analyzer2.isCjk());
+//        assertEquals(analyzer.cjk().stopwords(), analyzer2.cjk().stopwords());
+//        assertEquals(analyzer.cjk().stopwordsPath(), analyzer2.cjk().stopwordsPath());
+//    }
```
reta, line 175: @Test
    public void testLanguage_Analyzer() {
        for (final Language language: Language.values()) {
	        final Analyzer analyzer = new Analyzer.Builder().language(b -> b.language(language).stopwords(Arrays.asList("a", "b", "c")).stopwordsPath("path")).build();
	
	        String str = toJson(analyzer);
	        assertEquals("{\"type\":\"" + language.jsonValue() + "\",\"stopwords\":[\"a\",\"b\",\"c\"],\"stopwords_path\":\"path\"}", str);
	
	        Analyzer analyzer2 = fromJson(str, Analyzer._DESERIALIZER);
	        assertEquals(analyzer.language().stopwords(), analyzer2.language().stopwords());
	        assertEquals(analyzer.language().stopwordsPath(), analyzer2.language().stopwordsPath());
	        assertEquals(analyzer.language().language(), language);
        }
    }
}

java-client/src/test/java/org/opensearch/client/opensearch/experiments/ParsingTests.java
reta: Ok, @brentam I think I have a solution for that:
 - we keep `CjkAnalyzer` but 
 - we remove `Cjk` from `Language` enumeration

This is not ideal but the cons is that - it prevents users from referring to unsupported features at compile time (vs removing  `CjkAnalyzer` but failing at runtime)

brentam: @reta  
in practice, Cjk can have stem_exclusion, this is perfectly fine
```
PUT test_idx1
{
  "mappings": {
    "dynamic": "strict",
    "properties": {
      "title": {
        "doc_values": false,
        "type": "text"
      }
    }
  },
  "settings": {
    "index": {
      "analysis": {
        "analyzer": {
          "some_analyzer": {
            "type": "cjk",
            "stem_exclusion": ["pro", "-"]
          }
        },
        "char_filter": [
          "html_strip"
        ]
      }
    }
  }
```

brentam: @reta, after further tests, I dont think elastic is enforcing stem_exclusion at all
I saw that both `stem_exclusion` and `stem_exclusions` worked!!!
Then I tried to use other field name.
This is actually valid and the index is created!!!!
```

      "analysis": {
        "analyzer": {
          "some_analyzer": {
            "type": "cjk",
            "what_is_this": ["pro", "-"]
          }
        },
        "char_filter": [
          "html_strip"
        ]
      }
```
image

brentam: Look, it is clear our current api is not modeling the real world, it doesn't matter what we do at this point we won't be covering all cases for the analyzers..
I think the desision is to use the modified LanguageAnalyzers, or stick with the Analyzers Per language that was already used for Dutch and Cjk.

brentam: I found other big issues with the analyzers api while testing and looking at the code, I was thinking about creating other tickets because this discussion is becoming quite long but I will add them in this thread so you have an idea of the minimum I think needs to be done. Those are based on the use case we need to be supported at work.

brentam: Here it is
 There seems to be a lot of problems with the IndexSettingsAnalysis.Builder.
 _(This can be the subject of another PR)_

Check this example
```
        Map<String, Analyzer> map = new HashMap<>();
        map.put("some_analyzer",
                LanguageAnalyzer.of(l -> l.language(Language.German).
                                stemExclusion(new ArrayList<>()))
                        ._toAnalyzer());
        IndexSettingsAnalysis settings = IndexSettingsAnalysis.of(
                        anl -> anl
                                .analyzer(map)
                                .charFilter("some_char_filter", CharFilter.of(c -> c.name("html_strip")))
        );

        System.out.println(toJson(settings));
```

This will produce this json
`{"analyzer":{"some_analyzer":{"type":"language","language":"German","stem_exclusion":[]}},"char_filter":{"some_char_filter":"html_strip"}}`.

Now, we should be able to specify the char_filter under "some_analyzer" too, not only under the "analyzer" like it is here.

These two jsons are perfectly fine and we are actually using this in production:
```
          "english_analyzer": {
            "type": "english",
            "char_filter": [
              "html_strip"
            ],
            "tokenizer": "standard"
          }
```

```
          "german_analyzer" : {
            "filter" : [ "lowercase", "german_normalization", "german_stop", "german_synonyms", "german_decompounder", "german_synonyms", "possessive_stemmer", "german_stemmer" ],
            "char_filter" : [ "html_strip" ],
            "tokenizer" : "standard"
          },
```

So there is some redesigned needed, the easiest would be modify the language Deserializer to accept the "char_filter" and some other clauses.

Here is the Deserialization function for IndexSettingsAnalysis

```
    protected static void setupIndexSettingsAnalysisDeserializer(ObjectDeserializer<IndexSettingsAnalysis.Builder> op) {

        op.add(Builder::analyzer, JsonpDeserializer.stringMapDeserializer(Analyzer._DESERIALIZER), "analyzer");
        op.add(Builder::charFilter, JsonpDeserializer.stringMapDeserializer(CharFilter._DESERIALIZER), "char_filter");
        op.add(Builder::filter, JsonpDeserializer.stringMapDeserializer(TokenFilter._DESERIALIZER), "filter");
        op.add(Builder::normalizer, JsonpDeserializer.stringMapDeserializer(Normalizer._DESERIALIZER), "normalizer");
        op.add(Builder::tokenizer, JsonpDeserializer.stringMapDeserializer(Tokenizer._DESERIALIZER), "tokenizer");

    }
```
I believe what we need to do is to modify the Analyzer._DESERIALIZER and the LanguageAnalyzer.Builder to allow for
"char_filter", "filter", "normalizer" and "tokenizer".



The CharFilter.Builder seems to be incomplete too.
There is no way to construct an array of names, like how is expected here 
```
          "english_analyzer": {
            "type": "english",
            "char_filter": [
              "html_strip"
            ],
            "tokenizer": "standard"
          }
```

It only allows  a map
` .charFilter("some_char_filter", CharFilter.of(c -> c.name("html_strip")))`
that will result in
`{"some_char_filter":"html_strip"}`

Now, I was looking at the apis related to the Charfilter builders. There is a lot of classes and hard to understand the right ones to use, so refactoring that will be a pain...

I suggest the simple fix would be having a "reference" method in the builder...
We would set it like this
` CharFilter.of(c -> c.references(new String[]{"html_strip","some_custom_charset_defined_elsewhere"}))`

reta: Thanks @brentam, I till try to answer based on the documentation

> Arent we being to pendantic here?
Is really the way to construct the language analyzer using LanguageAnalyzer Builder that important??

This model is generated from the API specification so it is supposed to work or the API is completely wrong. The fact `stem_exclusion` are not enforced is probably not good, I suspect when specified, the expectation is  should be applied during analysis (I will look into it shortly). 

> That is another reason I am in favour of keeping the 33 new analyzers I have created in the other PR an not touching the LanguageAnalyzer at all

As you discovered, the `LanguageAnalyzer` does not work at all at the moment. I believe with the changes form this pull request it should work the way it is supposed to  work, right?  

> I believe what we need to do is to modify the Analyzer._DESERIALIZER and the LanguageAnalyzer.Builder to allow for
"char_filter", "filter", "normalizer" and "tokenizer".

No, I think we should not - this belongs to custom analyzers [1]

[1] https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-custom-analyzer.html#analysis-custom-analyzer

dblock:
Let's tag @msfroh here who works on search for some 👀 . I don't understand how much of this is supposed to work, and I am looking forweard to @brentam to become an expert here and help us sort this mess out!

java-client/src/test/java/org/opensearch/client/opensearch/experiments/ParsingTests.java
reta: @brentam quick update (my apologies, learning some things along with you), this is my understand of all the pieces so far:
1. Language Analyzers: OpenSearch uses some predefined language analyzers that come with Apache Lucene, those could be used by specifying `"type":"<language>"` when definiting analyzers
2. All of those predefined language analyzers support stop words but not all support `stem_exclusion`. The way it works though is very unexpected - there is no validation but the the unsupported properties (like `stem_exclusion` in case of `Cjk`) are silently ignored. More to that, to my understanding, it equally applies to `char_filter` & co - they are not taken into account since custom language analyzers do not support this customization. In practice it means is that what you think you configure is not what you get:
   ```
    "english_analyzer": {
            "type": "english",
            "char_filter": [
              "html_strip"
            ],
            "tokenizer": "standard"
          }
   ``` 

   is equivalent to 

    ```
    "english_analyzer": {
            "type": "english"
    }
    ```
   (and sadly    is equivalent to since `char_filter` and `tokenizer` are ignored completely

   ```
    "english_analyzer": {
            "type": "english",
            "char_filter": [
              "garbage"
            ],
            "tokenizer": "garbage"
          }
   ``` )

All in all, I think we are doing very good change (thanks to you) by first of all, fixing the client libraries, and secondly - by uncovering the gaps we currently have.

msfroh:
> Let's tag @msfroh here who works on search for some 👀 . I don't understand how much of this is supposed to work, and I am looking forward to @brentam to become an expert here and help us sort this mess out!

Thanks! I'll need to go through this a bit. 

I'm not clear on why the client needs to be aware of what analyzers exist server-side (especially since that can vary based on what plugins are installed on the server). I'm going to read through until I understand what the client is doing.

java-client/src/test/java/org/opensearch/client/opensearch/experiments/ParsingTests.java
brentam: omg, that is sad...lol.

brentam:
@reta @msfroh 
Sorry, I have been busy and could not visit this thread.
I can revisit this thread in the near future. Atm, out team is more concerned with another issue.
Due to some of the builders not able to deserialize our query properly due to some erros in the implementation.
What we need is a way to pass a raw json to the seach client. (without having to deserialize to the api java classes).
I will probably create an story for this, and will be glad to work on it if you guys agree is something we should have.

reta:
>  What we need is a way to pass a raw json to the seach client. (without having to deserialize to the api java classes). I will probably create an story for this, and will be glad to work on it if you guys agree is something we should have.

@brentam sure, please check if we already have an issue for that , fe https://github.com/opensearch-project/opensearch-java/issues/377 may be the one your are looking for

Issue #789, opened 2023-12-31
[BUG] PutPipelineRequest not deserializable due to missing `id` field


patschl:
### What is the bug?
Currently PutPipelineRequests cannot be deserialized due to the `id` field not being set.

### How can one reproduce the bug?
Create a PutPipelineRequest and add an `id` to the json, then use `PutPipelineRequest._DESERIALIZER` to deserialize the json.

### What is the expected behavior?
Creates a valid PutPipelineRequest.

### What is your host/environment?
Managed AWS OpenSearch/local docker container.

PR #790, opened 2023-12-31
Set 'id' field when deserializing a PutPipelineRequest

patschl:
### Description
Currently deserializing a PutPipelineRequest it not possible due to the `id` field not being set but also being required during deserialization.
This PR adds support for the `id` field. 

### Issues Resolved
Fixes https://github.com/opensearch-project/opensearch-java/issues/789

### DIFF
--- a/java-client/src/main/java/org/opensearch/client/opensearch/ingest/PutPipelineRequest.java
+++ b/java-client/src/main/java/org/opensearch/client/opensearch/ingest/PutPipelineRequest.java
@@ -481,6 +481,7 @@ public class PutPipelineRequest extends RequestBase implements JsonpSerializable
 
         op.add(Builder::meta, JsonpDeserializer.stringMapDeserializer(JsonData._DESERIALIZER), "_meta");
         op.add(Builder::description, JsonpDeserializer.stringDeserializer(), "description");
+        op.add(Builder::id, JsonpDeserializer.stringDeserializer(), "id");
         op.add(Builder::onFailure, JsonpDeserializer.arrayDeserializer(Processor._DESERIALIZER), "on_failure");
         op.add(Builder::processors, JsonpDeserializer.arrayDeserializer(Processor._DESERIALIZER), "processors");
         op.add(Builder::version, JsonpDeserializer.longDeserializer(), "version");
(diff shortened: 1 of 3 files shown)


java-client/src/main/java/org/opensearch/client/opensearch/ingest/PutPipelineRequest.java
```diff
@@ -481,6 +481,7 @@ protected static void setupPutPipelineRequestDeserializer(ObjectDeserializer<Put
 
         op.add(Builder::meta, JsonpDeserializer.stringMapDeserializer(JsonData._DESERIALIZER), "_meta");
         op.add(Builder::description, JsonpDeserializer.stringDeserializer(), "description");
+        op.add(Builder::id, JsonpDeserializer.stringDeserializer(), "id");
```
reta, line 484: @patschl the `PutPipelineRequest` does not include `id` as the payload - it is part of the URL and passed as the URL path component. Under which circumstances it is returned in the request payload? Could you please share the example? (sadly the linked issue has no payload sample).

patschl: Thanks @reta!
I'm not quite sure I understand the purpose of this deserializer correctly if it should strictly deal with the payload of a request.
`id` is a required field for the `PutPipelineRequest`, so deserializing this object will currently fail every time, since the deserializer of course won't set it. Wouldn't this make the deserializer unnecessary if it won't work anyway?

We had a similar issue with the `name` property in the `PutIndexTemplateRequest`, but managed to work around it.

We persist such index template and pipeline definitions as json in our VCS and create the requests on demand, this is the reason for this PR.

reta: Thanks @patschl 

> I'm not quite sure I understand the purpose of this deserializer correctly if it should strictly deal with the payload of a request.

The deserializer mirrors serializer: in this particular case, the `id` is not part of the serialized JSON (payload) and as such, should not be deserialized from the payload (since it should be there in the first place).

patschl: @reta Okay I see, I will close this PR then. Thank you for clearing that up.
There is no way to create this object from a json using this deserializer, but I guess this was never the use case anyway?

reta: > There is no way to create this object from a json using this deserializer, but I guess this was never the use case anyway?

Thank you @patschl, I think the question is under which circumstances it should happen, this is a request which naturally only uses one way serialization (when send to the service). But in general - yes, it seems like some models may need more context (URI components) to deserialize completely.

Issue #778, opened 2023-12-21
[FEATURE] Add an integration test that runs on JDK8


dblock:
### Is your feature request related to a problem?

Coming here via https://github.com/opensearch-project/opensearch-java/issues/156, in https://github.com/opensearch-project/opensearch-java/pull/767 we restored support for JDK8, but did not enable CI.

### What solution would you like?

A workflow that exercises the client on JDK8. It could mean building on 11, but executing an integration test on a version 8 runtime.

### What alternatives have you considered?

Let JDK8 support fail without notice.

### Do you have any additional context?

https://github.com/opensearch-project/opensearch-java/issues/156#issuecomment-1865811097

reta:
Thanks @dblock , I will take it, have some ideas on how we could approach that

PR #795, opened 2024-01-02
Add an integration test that runs on JDK-8

reta:
### Description
Add an integration test that runs on JDK-8

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-java/issues/778


java-client/build.gradle.kts
```diff
@@ -369,5 +372,17 @@ if (JavaVersion.current() >= JavaVersion.VERSION_11) {
     testClassesDirs += java11.output.classesDirs
     classpath = sourceSets["java11"].runtimeClasspath
   }
- 
+} else if (jdk8compatibility == true) {
+    java {
```
reta, line 376: @dblock @VachaShah curious what you think about this approach: we cannot run Gradle build using JDK-8 but we could use toolchains to run tests under JDK-8.

java-client/build.gradle.kts
```diff
@@ -337,7 +338,9 @@ publishing {
     }
 }
 
-if (JavaVersion.current() >= JavaVersion.VERSION_11) {
+// Use `-Pcheck-jdk8-compatibility=true` to
+val jdk8compatibility = (project.findProperty("check-jdk8-compatibility") as String?).toBoolean()
+if (JavaVersion.current() >= JavaVersion.VERSION_11 && jdk8compatibility == false) {
```
dblock, line 343: I dislike the whole `check-jdk8-compatibility=true` business. Why do we need this at all and not just

```java
if (JavaVersion.current() >= JavaVersion.VERSION_11) {
 ...
} else if (JavaVersion.current() >= JavaVersion.VERSION_8) {
 ...
}
```

reta: That the thing, the `JavaVersion.current()` is the version Gradle run with, it is always 11 or higher. I will try to experiment with toolchains here.

dblock: Ok, I see. 

Maybe we could improve on this by introducing optional `JAVA_TOOLCHAIN_VERSION` and default it to 11. Basically avoiding this whole `if` logic?

Stepping back, this does seem to work, so we should merge it and improve it later if you can't come up with something a lot nicer. Leave some TODOs with an explanation at the least?

reta: This is a good idea, thanks @dblock , I refactored it to use similar to OpenSearch core approach with `runtime.java` version.

dblock (approved):
Perfect, nice and clean!

reta:
> Perfect, nice and clean!

As always, Windows have troubles with env variables that have dots

dblock (approved):

PR #831, opened 2024-02-06
Removed required index for Hit.java

### LINKED ISSUE #825: [BUG] InnerHIt result not containing a _index is throwing Hit.Index missing
### What is the bug?
_A clear and concise description of the bug._
```
Caused by: org.opensearch.client.util.MissingRequiredPropertyException: Missing required property 'Hit.index'
```
I have a query that I am getting back from opensearch that has a Hit of a parent, and alongside that, I added the option to retrieve all the children of that parent document as Innerhits. 

I've included an example below of the query, where I think it should be getting a parent that has a child document of type `childType`, and then in the innerhits, get the children for it.
```
Query.of {
    it.hasChild {
        it.type("childType")
        it.query {
            it.matchAll { it -> it }
        }
        it.innerHits {
            it.source {
                it.fetch(true)
            }
        }
    }
}
```
Hit.java is able to convert the response to the java class for the parent perfectly fine, however when trying to convert its children, it fails because each InnerHit included with the parent Hit does not have a index specified in the response. I believe it was done under the intention that the the index of the InnerHits should always match the parent, however, the conversion is not aware of the Parent document's index for the InnerHit.

_Steps to reproduce the behavior._
Reproduction steps would be to create two documents, and associate a parent-child relation to them. And then run a query for a property that would match the parent, and then adding the innerHits field:
```
it.innerHits {
            it.source {
                it.fetch(true)
            }
        }
```
_A clear and concise description of what you expected to happen._
Hit.java should be able to convert the InnerHits into Hit objects properly.
_Operating system, version._
MacOS 14.2.1

I have to wonder why the index is set to mandatory? I did see a discussion on elasticsearch back in 2016 https://github.com/elastic/elasticsearch/issues/18091 but I am not sure if it's the same.

jvan1997:

### Description
Change removes the required field for index in the Hit.java
### Issues Resolved
Closes #825 

### DIFF
--- a/java-client/src/main/java/org/opensearch/client/opensearch/core/search/Hit.java
+++ b/java-client/src/main/java/org/opensearch/client/opensearch/core/search/Hit.java
@@ -107,7 +107,7 @@ public class Hit<TDocument> implements JsonpSerializable {
 
     private Hit(Builder<TDocument> builder) {
 
-        this.index = ApiTypeHelper.requireNonNull(builder.index, this, "index");
+        this.index = builder.index;
         this.id = builder.id;
         this.score = builder.score;
         this.explanation = builder.explanation;
@@ -134,7 +134,7 @@ public class Hit<TDocument> implements JsonpSerializable {
     }
 
     /**
-     * Required - API name: {@code _index}
+     * API name: {@code _index}
      */
     public final String index() {
         return this.index;
@@ -281,8 +281,10 @@ public class Hit<TDocument> implements JsonpSerializable {
 
     protected void serializeInternal(JsonGenerator generator, JsonpMapper mapper) {
 
-        generator.writeKey("_index");
-        generator.write(this.index);
+        if (this.index != null) {
+            generator.writeKey("_index");
+            generator.write(this.index);
+        }
 
         if (this.id != null) {
             generator.writeKey("_id");
@@ -419,6 +421,8 @@ public class Hit<TDocument> implements JsonpSerializable {
      */
 
     public static class Builder<TDocument> extends ObjectBuilderBase implements ObjectBuilder<Hit<TDocument>> {
+
+        @Nullable
         private String index;
 
         @Nullable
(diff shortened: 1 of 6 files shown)


java-client/src/main/java/org/opensearch/client/opensearch/core/search/Hit.java
```diff
@@ -107,7 +107,7 @@ public class Hit<TDocument> implements JsonpSerializable {
 
     private Hit(Builder<TDocument> builder) {
 
-        this.index = ApiTypeHelper.requireNonNull(builder.index, this, "index");
+        this.index = builder.index;
```
Sunhick, line 110: How do you ensure that this is only for a inner hit ? And outer hits should enforce ```requireNonNull```.

java-client/src/test/java/org/opensearch/client/opensearch/core/search/InnerHitsResultTest.java
```diff
@@ -0,0 +1,84 @@
...
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+
+import java.io.StringReader;
+import org.junit.Test;
+import org.opensearch.client.json.JsonData;
+import org.opensearch.client.json.JsonpMapper;
+import org.opensearch.client.json.jsonb.JsonbJsonpMapper;
+
+public class InnerHitsResultTest {
```
Sunhick, line 13: Do we have unit tests for outer hits with nullable ? Should be allowed or not allowed ?

java-client/src/main/java/org/opensearch/client/opensearch/core/search/Hit.java
jvan1997: I do think this is a good call out; The problem is that the the InnerHit is in a List of [HitsMetadata](https://github.com/opensearch-project/opensearch-java/blob/0334c78c88205eb3b1fece45796125cbe92f9134/java-client/src/main/java/org/opensearch/client/opensearch/core/search/HitsMetadata.java#L56), which is then in the [InnerHitsResult](https://github.com/opensearch-project/opensearch-java/blob/0334c78c88205eb3b1fece45796125cbe92f9134/java-client/src/main/java/org/opensearch/client/opensearch/core/search/InnerHitsResult.java#L52).

So unless there is a file made explicit to represent an InnerHit (which I think is a valid path), the other option would be to re-populate the Index and the Id to the InnerHits as well, even if they match the parent.

PR #833, opened 2024-02-06
Removed required index for Hit.java

jvan1997:

### Description

### Issues Resolved


dblock (approved):
👏

dblock:
@jvan1997 I think we also want this on a 2.x, will need a manual backport please?

jvan1997:
@dblock Raised here: https://github.com/opensearch-project/opensearch-java/pull/838 thank you for the approval!

PR #838, opened 2024-02-07
Removed required index for Hit.java (#833)

### LINKED ISSUE #825: [BUG] InnerHIt result not containing a _index is throwing Hit.Index missing

jvan1997:

Signed-off-by: jvan1997 <32347479+jvan1997@users.noreply.github.com>
(cherry picked from commit c18adb94172f2ea2c2b2df7afc737dd523b35dd2)

### Description

### Issues Resolved
Closes #825 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Fix partial success results for msearch_template ([#709](https://github.com/opensearch-project/opensearch-java/pull/709))
 - Fix deserialization of node stats response ([#745](https://github.com/opensearch-project/opensearch-java/pull/745))
 - Fix PutIndexTemplateRequest field deserialization ([#765](https://github.com/opensearch-project/opensearch-java/pull/765))
+- Fix InnerHits to no longer enforce the nullable Index field when converting to Hit. ([#825](https://github.com/opensearch-project/opensearch-java/issues/825))
 
 ### Security
 
--- a/java-client/src/main/java/org/opensearch/client/opensearch/core/search/Hit.java
+++ b/java-client/src/main/java/org/opensearch/client/opensearch/core/search/Hit.java
@@ -107,7 +107,7 @@ public class Hit<TDocument> implements JsonpSerializable {
 
     private Hit(Builder<TDocument> builder) {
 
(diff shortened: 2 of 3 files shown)


dblock:
I saw some comments on another PR from @sunhick, do those make sense? https://github.com/opensearch-project/opensearch-java/pull/831#pullrequestreview-1868648875

dblock (approved):

jvan1997:
I think that's a valid case, I think the issue may stem from using the Hit.java for both the InnerHit (which does not have the index nor id), and also for the Hit itself.

Do we potentially want to create a different class for the InnerHit entirely, or set a flag to indicate that the Hit is an InnerHit?

dblock:
> Do we potentially want to create a different class for the InnerHit entirely, or set a flag to indicate that the Hit is an InnerHit?

I think it should be two different classes. If they share functionality it should be a base class.

PR #882, opened 2024-03-01
Fix KNNQuery Precision

luyuncheng:
### Description
There is a knn query like

```
KnnQuery origin = new KnnQuery.Builder().field("field").vector(new float[] { 0.1f, 0.4f }).k(1).build();
```

When it serialize to json string, the result would be:

```
{"field":{"vector":[0.10000000149011612,0.4000000059604645],"k":1}}
```

like the test code shows:

https://github.com/luyuncheng/opensearch-java/blob/4436e89a0f48d54e47c2c33689e0b828cbb31149/java-client/src/test/java/org/opensearch/client/opensearch/_types/query_dsl/KnnQueryTest.java#L16-L19

### Proposal
I think it because in `KnnQuery#serializeInternal` call `generator.write(value);` but `jakarta.json.stream.JsonGenerator` write numeric only support double. when `KnnQuery#vector` is float type, it would make precision not correctly.

https://github.com/opensearch-project/opensearch-java/blob/045c805eb58a166f55888d27a0358d184f117761/java-client/src/main/java/org/opensearch/client/opensearch/_types/query_dsl/KnnQuery.java#L86-L97

### Issues Resolved

### DIFF
--- a/java-client/src/main/java/org/opensearch/client/opensearch/_types/query_dsl/KnnQuery.java
+++ b/java-client/src/main/java/org/opensearch/client/opensearch/_types/query_dsl/KnnQuery.java
@@ -9,6 +9,7 @@
 package org.opensearch.client.opensearch._types.query_dsl;
 
 import jakarta.json.stream.JsonGenerator;
+import java.math.BigDecimal;
 import java.util.function.Function;
 import javax.annotation.Nullable;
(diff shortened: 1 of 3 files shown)


reta:
@luyuncheng I believe the precision with float numbers is not specific to knn plugin (the float is used in many other places like boost etc, ...), why it is causing issues in your case?

Issue #886, opened 2024-03-07
[BUG] Unable to create a pattern replace character filter without flags


kmessaoudi:
### What is the bug?
We are trying to create a index using this library and trying to create a pattern replace character filter with no flags. This is generating an error.

### How can one reproduce the bug?
It works properly directly via the API.
```
PUT test-charfilter
{
  "settings": {
    "analysis": {
      
      "analyzer": {
        "analyzer_fr_FR": {
          "filter": [
            "lowercase",
            "asciifolding",
            "remove_duplicates"
          ],
          "type": "custom",
          "tokenizer": "standard",
          "char_filter": [
            "special_char_filter"
          ]
        }
      },
      "char_filter": {
        "special_char_filter": {
          "pattern": """["@\$#&*{}\[\],\.\=\+\-_\(\);'/\?"]""",
          "type": "pattern_replace",
          "replacement": ""
        }
      }
    }
  }
}
```
we will get
```
{
  "acknowledged": true,
  "shards_acknowledged": true,
  "index": "test-charfilter"
}
```

However, when using the lib : 
```
// opensearchClient is injected through a bean
String templateSettings = "{"analysis":{"analyzer":{"analyzer_fr_FR":{"filter":["lowercase","asciifolding","remove_duplicates"],"type":"custom","tokenizer":"standard","char_filter":["special_char_filter"]}},"char_filter":{"special_char_filter":{"pattern":"""["@\$#&*{}\[\],\.\=\+\-_\(\);'/\?"]""","type":"pattern_replace","replacement":""}}}}";
JsonpMapper mapper = opensearchClient._transport().jsonpMapper();
JsonParser settingsParser =
      mapper
        .jsonProvider()
        .createParser(new StringReader(templateSettings));

IndexSettings indexSettings = IndexSettings._DESERIALIZER.deserialize(settingsParser, mapper);

return CreateIndexRequest.Builder()
      .settings(indexSettings)
      .build();
```
We will get this error
```
org.opensearch.client.util.MissingRequiredPropertyException - Missing required property 'PatternReplaceCharFilter.flags'
```

When checking the class opensearch PatternReplaceCharFilter flags is required
https://github.com/opensearch-project/opensearch-java/blob/045c805eb58a166f55888d27a0358d184f117761/java-client/src/main/java/org/opensearch/client/opensearch/_types/analysis/PatternReplaceCharFilter.java#L60

[...]

PR #895, opened 2024-03-14
Fix pattern replace by making flag optional as on api

grouh:
### Description
This PR makes the flag optional on the pattern replace char filter, as on OS API
### Issues Resolved
Fixes https://github.com/opensearch-project/opensearch-java/issues/886

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 ### Fixed
 - Fix integer overflow for variables in indices stats response ([#877](https://github.com/opensearch-project/opensearch-java/pull/877))
 - Support weight function in function score query ([#880](https://github.com/opensearch-project/opensearch-java/pull/880))
+- Fix pattern replace by making flag optional as on api  ([#895](https://github.com/opensearch-project/opensearch-java/pull/895))
 
 ### Security
 
@@ -65,7 +66,6 @@ This section is for maintaining a changelog for all breaking changes for the cli
 
 ### Fixed
 - Fix missing properties on UpdateOperation ([#744](https://github.com/opensearch-project/opensearch-java/pull/744))
-- Fix pattern replace by making flag optional as on api  ([#895](https://github.com/opensearch-project/opensearch-java/pull/895))
 
 ### Security
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 
 ### Fixed
 - Fix missing properties on UpdateOperation ([#744](https://github.com/opensearch-project/opensearch-java/pull/744))
+- Fix pattern replace by making flag optional as on api  ([#895](https://github.com/opensearch-project/opensearch-java/pull/895))
 
 ### Security
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 ### Fixed
 - Fix integer overflow for variables in indices stats response ([#877](https://github.com/opensearch-project/opensearch-java/pull/877))
 - Support weight function in function score query ([#880](https://github.com/opensearch-project/opensearch-java/pull/880))
-- Fix pattern replace by making flag optional as on api  ([#895](https://github.com/opensearch-project/opensearch-java/pull/895))
+- Fix pattern replace by making flag and replacement optional as on api  ([#895](https://github.com/opensearch-project/opensearch-java/pull/895))
 
 ### Security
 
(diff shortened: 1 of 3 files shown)


reta:
Thank you, @grouh , please add CHANGELOG entry and at least one test for the change.

reta (approved):

VachaShah (dismissed):

CHANGELOG.md
```diff
@@ -65,6 +65,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 
 ### Fixed
 - Fix missing properties on UpdateOperation ([#744](https://github.com/opensearch-project/opensearch-java/pull/744))
+- Fix pattern replace by making flag optional as on api  ([#895](https://github.com/opensearch-project/opensearch-java/pull/895))
```
VachaShah, line 68: The changelog should be in Unreleased 2.x section.

harawata:
I could be wrong, but `replacement` also is optional, isn't it?

reta:
> I could be wrong, but `replacement` also is optional, isn't it?

Hm ... I don't think so, the filter is not very useful if it replaces nothing (by and large, this replace charfilter)

harawata:
@reta ,

Thank you for the reply!

I don't disagree about the usefulness and `replacement` being required does not affect me, so please carry on. :)

I commented only because 1) the server accepts `pattern_replace` without `replacement` and 2) the `replacement` is optional in the Elasticsearch doc (I couldn't find OpenSeach doc about the filter).
https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-pattern-replace-charfilter.html


```json
PUT my-index-000001
{
  "settings": {
    "analysis": {
      "analyzer": {
        "my_analyzer": {
          "tokenizer": "standard",
          "char_filter": [
            "my_char_filter"
          ],
          "filter": [
            "lowercase"
          ]
        }
      },
      "char_filter": {
        "my_char_filter": {
          "type": "pattern_replace",
          "pattern": "(?<=\\p{Lower})(?=\\p{Upper})"
        }
      }
    }
  },
  "mappings": {
    "properties": {
      "text": {
        "type": "text",
        "analyzer": "my_analyzer"
      }
    }
  }
}
```

reta:
> I don't disagree about the usefulness and `replacement` being required does not affect me, so please carry on. :)

Thank you @harawata , what would be function of the char filter when `pattern_replace` is provided without `replacement`? The fact server accepts this combination point out there could be a problem there (and yes, we do have large gaps in documentation sadly).

harawata:
@reta ,

I just did a quick test and it seems that when `replacement` is omitted, it simply removes the matched string (so, same behavior as `replacement: ""`, probably?).

For example, with this index...

```json
PUT my-index-000001
{
  "settings": {
    "analysis": {
      "analyzer": {
        "my_analyzer": {
          "tokenizer": "standard",
          "char_filter": [
            "my_char_filter"
          ],
          "filter": [
            "lowercase"
          ]
        }
      },
      "char_filter": {
        "my_char_filter": {
          "type": "pattern_replace",
          "pattern": "Ba."
        }
      }
    }
  },
  "mappings": {
    "properties": {
      "text": {
        "type": "text",
        "analyzer": "my_analyzer"
      }
    }
  }
}
```
... the following returns `["the", "foo", "method"]`.

```json
POST my-index-000001/_analyze
{
  "analyzer": "my_analyzer",
  "text": "The fooBarBaz method"
}
```

I'm still new to Elasticsearch/OpenSearch, so please forgive me if I'm missing something basic.

reta:
> I just did a quick test and it seems that when `replacement` is omitted, it simply removes the matched string (so, same behavior as `replacement: ""`, probably?).

Thanks @harawata , you are totally and absolutely correct in your observations (pasting below the server side code that uses empty string if value is not provided):

```
replacement = settings.get("replacement", ""); // when not set or set to "", use "".
```

With that being said, I would prefer the client to require `replacement`, as an explicit intent what the replacement value should be, thank you.

harawata:
Personally, I am OK with that.

I should, however, mention that the mismatch between server and client could cause an issue when retrieving the index.
Assuming that the index in my previous comment exists on the server, the following code throws the `MissingRequiredPropertyException` when deserializing the server response.

```java
osClient.indices().get(x -> x.index("my-index-000001"));
```

This was actually how I encountered #886 at first.

CHANGELOG.md
grouh: Moved to the good section!

grouh:
If the consensus si to let `replacement` as mandatory for explicit use of the replace value, can this PR considered for merging ? or is there any other blocking point ?

reta:
> If the consensus si to let `replacement` as mandatory for explicit use of the replace value, can this PR considered for merging ? or is there any other blocking point ?

I think the deserialization failure mentioned by @harawata  (mismatch between server and client could cause an issue when retrieving the index) leave as no choice - we have to make `replacement` optional. Could you please address that and also have  a test for it? Thank you and thanks to @harawata to digging it through.

reta (approved):

dblock (approved):

grouh:
Thanks @reta @dblock @harawata @VachaShah

PR #910, opened 2024-04-01
[FEATURE] Enable Generic HTTP Actions in Java Client

reta:
### Description
Enable Generic HTTP Actions in Java Client. 
USER_GUIDE.md update is on me once we settle up on the APIs.

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-java/issues/377

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 
 ### Added
 - Add missed fields to MultisearchBody: seqNoPrimaryTerm, storedFields, explain, fields, indicesBoost ([#914](https://github.com/opensearch-project/opensearch-java/pull/914))
+- [FEATURE] Enable Generic HTTP Actions in Java Client ([#910](https://github.com/opensearch-project/opensearch-java/pull/910))
 
 ### Dependencies
 - Bumps `io.github.classgraph:classgraph` from 4.8.161 to 4.8.165
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 
 ### Added
 - Add missed fields to MultisearchBody: seqNoPrimaryTerm, storedFields, explain, fields, indicesBoost ([#914](https://github.com/opensearch-project/opensearch-java/pull/914))
-- [FEATURE] Enable Generic HTTP Actions in Java Client ([#910](https://github.com/opensearch-project/opensearch-java/pull/910))
+- Add OpenSearchGenericClient with support for raw HTTP request/responses ([#910](https://github.com/opensearch-project/opensearch-java/pull/910))
 
 ### Dependencies
 - Bumps `io.github.classgraph:classgraph` from 4.8.161 to 4.8.165
--- a/guides/generic.md
+++ b/guides/generic.md
@@ -6,17 +6,17 @@
 
 # Generic Client
 
-There are rare circumstances when the typed OpenSearch client APIs are too constraining and there is a need to communicate with OpenSearch cluster (or individual nodes) over generic HTTP request / response communication. The `OpenSearchGenericClient` could be used in such cases.
+There are rare circumstances when the typed OpenSearch client APIs are too constraining and there is a need to communicate with OpenSearch cluster (or individual nodes) over generic HTTP request / response communication. Use `OpenSearchGenericClient` in such cases.
 
 ## Getting the Client
-The following sample code gets the `OpenSearchGenericClient` rom the `OpenSearchClient` instance.
+The following sample code gets the `OpenSearchGenericClient` from the `OpenSearchClient` instance.
 
 ```java
 final OpenSearchGenericClient generic = javaClient().generic();
 ```
 
 ## Sending Simple Request
-The following sample code sends a simple request that does not require any payload to be provided.
+The following sample code sends a simple request that does not require any payload to be provided (typically, `GET` requests).
 
 ```java
 // compare with what the low level client outputs
@@ -25,7 +25,24 @@ try (Response response = javaClient().generic().execute(Requests.builder().endpo
 }
 ```
 
-Please notice that the `Response` instance should be closed explicitly in order to free up any allocated resource (like response input streams or buffers). The generic client never interprets status codes and provides the direct access to the response as it was received from the server. This is responsibility of the caller to decide what should happen in case of unsuccessful invocations.
+Please notice that the `Response` instance should be closed explicitly in order to free up any allocated resource (like response input streams or buffers), the [`try-with-resource`](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) pattern is encouraged.
+
+```java
+try (Response response = javaClient().generic().execute(...)) {
+   // ...
+}
+```
+
+The generic client never interprets status codes and provides the direct access to the response as it was received from the server. This is responsibility of the caller to decide what should happen in case of unsuccessful invocations.
+
+```java
+// compare with what the low level client outputs
+try (Response response = javaClient().generic().execute(...)) {
+   if (response.getStatus() != 200) {
+      // Request was not successful
+   }
+}
+```
 
 ## Sending JSON Requests
 The following sample code a simple request with JSON body.
--- a/guides/generic.md
+++ b/guides/generic.md
@@ -2,7 +2,8 @@
   - [Getting the Client](#get-client)
   - [Sending Simple Requests](#request-bodyless)
   - [Sending JSON Requests](#request-json)
-  - [Sending JSON using POJOs](#request-pojo)
+  - [Sending JSON Requests using POJOs](#request-pojo)
+  - [Sending Requests using structured JSON](#request-structured)
 
 # Generic Client
 
@@ -66,7 +67,7 @@ The following sample code a simple request with JSON body.
 }
 ```
 
-## Sending JSON using POJOs
+## Sending JSON Requests using POJOs
 Besides providing the ability to deal with raw request and response payloads (bodies), the `OpenSearchGenericClient` could be used mixed with existing OpenSearch typed requests and responses (POJOs), like the following sample code demonstrates.
 
 
@@ -91,6 +92,36 @@ try (Response response = javaClient().generic()
   final CreateIndexResponse r = response.getBody()
       .map(b -> Bodies.json(b, CreateIndexResponse._DESERIALIZER, jsonpMapper))
       .orElse(null);
+  // ...
+}
+```
+
+## Sending Requests using structured JSON
+Dealing with strings or POJOs could be daunting sometimes, using structured JSON APIs is a middle ground of both approaches, as per following sample code that uses (`jakarta.json.Json`)[https://jakarta.ee/specifications/jsonp].
+
+```java
+try (Response response = javaClient().generic()
+      .execute(
+        Requests.builder()
+          .endpoint("/" + index)
+          .method("PUT")
+          .json(Json.createObjectBuilder()
+              .add("settings", Json.createObjectBuilder()
+                .add("index", Json.createObjectBuilder()
+                  .add("sort.field", "name"))
+                  .add("sort.order", "asc")
+              )
+              .add("mappings",Json.createObjectBuilder()
+                .add("properties", Json.createObjectBuilder()
+                  .add("name", Json.createObjectBuilder()
+                    .add("type", "keyword"))
+                    .add("doc_values", true)
+                  .add("size", Json.createObjectBuilder()
+                    .add("type", "keyword"))
+                    .add("doc_values", true))
+              )
+          )
+          .build())) {
   // ...
 }
 ```
--- a/guides/generic.md
+++ b/guides/generic.md
@@ -1,0 +1,79 @@
+- [Generic Client](#generic-client)
+  - [Getting the Client](#get-client)
+  - [Sending Simple Requests](#request-bodyless)
+  - [Sending JSON Requests](#request-json)
+  - [Sending JSON using POJOs](#request-pojo)
+
+# Generic Client
+
+There are rare circumstances when the typed OpenSearch client APIs are too constraining and there is a need to communicate with OpenSearch cluster (or individual nodes) over generic HTTP request / response communication. The `OpenSearchGenericClient` could be used in such cases.
+
+## Getting the Client
+The following sample code gets the `OpenSearchGenericClient` rom the `OpenSearchClient` instance.
+
+```java
+final OpenSearchGenericClient generic = javaClient().generic();
+```
+
+## Sending Simple Request
+The following sample code sends a simple request that does not require any payload to be provided.
+
+```java
+// compare with what the low level client outputs
+try (Response response = javaClient().generic().execute(Requests.builder().endpoint("/").method("GET").build())) {
+   // ...
+}
+```
+
+Please notice that the `Response` instance should be closed explicitly in order to free up any allocated resource (like response input streams or buffers). The generic client never interprets status codes and provides the direct access to the response as it was received from the server. This is responsibility of the caller to decide what should happen in case of unsuccessful invocations.
+
+## Sending JSON Requests
+The following sample code a simple request with JSON body.
+
+```java
+ try (Response response = javaClient().generic()
+        .execute(
+            Requests.builder()
+                .endpoint("/" + index + "/_search")
+                .method("POST")
+                .json("{"
+                    + "    \"query\": {"
+                    + "        \"match_all\": {}"
+                    + "    }"
+                    + "}"
+                )
+                .build())) {
+   // Retrieve the response body as a simple string
+   final String body = response.getBody().map(Body::getAsString).orElse("");
+   // ...
+}
+```
+
+## Sending JSON using POJOs
+Besides providing the ability to deal with raw request and response payloads (bodies), the `OpenSearchGenericClient` could be used mixed with existing OpenSearch typed requests and responses (POJOs), like the following sample code demonstrates.
+
+
+```java
+final JsonpMapper jsonpMapper = javaClient()._transport().jsonpMapper();
+
+final CreateIndexRequest request = CreateIndexRequest.of(
+  b -> b.index(index)
+    .mappings(
+        m -> m.properties("name", Property.of(p -> p.keyword(v -> v.docValues(true))))
+            .properties("size", Property.of(p -> p.keyword(v -> v.docValues(true))))
+    )
+    .settings(settings -> settings.sort(s -> s.field("name").order(SegmentSortOrder.Asc)))
+);
+
+try (Response response = javaClient().generic()
+      .execute(
+          Requests.builder()
+            .endpoint("/" + index).method("PUT")
+            .json(request, jsonpMapper).build())) {
+  // Retrieve the response body as a POJO
+  final CreateIndexResponse r = response.getBody()
+      .map(b -> Bodies.json(b, CreateIndexResponse._DESERIALIZER, jsonpMapper))
+      .orElse(null);
+  // ...
+}
+```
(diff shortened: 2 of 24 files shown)


java-client/src/main/java/org/opensearch/client/opensearch/OpenSearchClient.java
```diff
@@ -155,6 +156,9 @@ public OpenSearchClient withTransportOptions(@Nullable TransportOptions transpor
     }
 
     // ----- Child clients
+    public OpenSearchGenericClient generic() {
+        return new OpenSearchGenericClient(this.transport, this.transportOptions);
+    }
```
dblock, line 161: Why do we need a new kind of client? Why can't I use the existing client?

java-client/src/main/java/org/opensearch/client/opensearch/generic/GenericBodies.java
```diff
@@ -0,0 +1,50 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.client.opensearch.generic;
```
dblock, line 9: It's really `json`, not `generic`, should we call it that? This way if we add other serializations in the future we have the opportunity to create a base class with a different name.

dblock (commented):
Good progress!

Let's talk about how it's going to be used? I think we don't want a generic client. Is there a world where I can mix and match? For example, there's no reason why I cannot make a generic JSON request, but get back a serialized `SearchResponse`, and vice-versa.

java-client/src/main/java/org/opensearch/client/opensearch/generic/GenericBodies.java
reta: > It's really `json`, not `generic`, should we call it that?

Not, it is not JSON to be fair, example is `Bulk` which is not JSON

java-client/src/main/java/org/opensearch/client/opensearch/OpenSearchClient.java
reta: Existing clients are all about object model: requests and responses are all typed. The idea of generic one is basically allow to send raw requests and receive raw response (by raw I mean just as bytes).

java-client/src/main/java/org/opensearch/client/opensearch/generic/GenericBodies.java
dblock: Bulk is "line-delimited JSON", so that's still JSON.

reta: > Bulk is "line-delimited JSON", so that's still JSON.

To elaborate on that, the content types that OpenSearch deals with include YAML, JSON, CBOR and SMILE , this is what server and client supports (`opensearch-java` only supports JSON but that is rather only client limitation)

reta:
> Let's talk about how it's going to be used? I think we don't want a generic client. Is there a world where I can mix and match?

Yes and no (if by mix and match you mean reuse existing models), sadly it turned out to be rather difficult, for simple cases it works pretty well (snippets from the tests):
```
GenericResponse response = javaClient().generic() .execute(
        new GenericRequest(
            "PUT",
            "/" + index + "/_doc/" + id,
            Collections.emptyList(),
            GenericBodies.json(document, javaClient()) // <-- document is arbitrary Java class here
        )
    )
```

```            
final CreateIndexResponse r = GenericBodies.json(response.getBody().get(), CreateIndexResponse._DESERIALIZER, javaClient());
assertThat(r.acknowledged(), equalTo(true));
```

But others are very difficult to implement at the moment.

java-client/src/main/java/org/opensearch/client/opensearch/generic/GenericBodies.java
```diff
@@ -0,0 +1,97 @@
...
+        return json(body, deserializer, clazz, client._transport().jsonpMapper());
+    }
+
+    public static <V, C extends SearchResponse<?>> C json(
+        GenericBody body,
+        JsonpDeserializer<C> deserializer,
+        Class<V> clazz,
+        JsonpMapper jsonpMapper
+    ) {
+        try (JsonParser parser = jsonpMapper.jsonProvider().createParser(body.body())) {
```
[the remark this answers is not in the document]
dblock: I think you understand what I am after. Anything that gets us closer is good with me to merge!

java-client/src/main/java/org/opensearch/client/opensearch/generic/GenericBodies.java
dblock: What if we subclasses the generics into YamlRequest, JsonRequest, etc.? For semantics and for passing the correct content-type.

reta: But why? That would basically add a tons of abstractions but not much value vs having factory methods to deal with request / response bodies.

```
GenericBodies  {
    GenericBody json(...)
    GenericBody yaml(...)
}

```

dblock: I think I am stuck at wanting to write this:

```java
Response response = javaClient()
                .execute(
                    new Request(
                        "POST",
                        "/" + index + "/_search",
                        Collections.emptyList()
                    ).json(
                            "{"
                                + "    \"sort\": ["
                                + "        {"
                                + "            \"name\": \"asc\""
                                + "        }"
                                + "    ],"
                                + "    \"fields\": ["
                                + "        \"name\""
                                + "    ],"
                                + "    \"query\": {"
                                + "        \"bool\": {"
                                + "            \"filter\": ["
                                + "                {"
                                + "                    \"term\": {"
                                + "                        \"size\": \"huge\""
                                + "                    }"
                                + "                }"
                                + "            ]"
                                + "        }"
                                + "    },"
                                + "    \"_source\": true\n"
                                + "}"
                        )
                    )
                )
        ) 
```

The difference is that I removed the `Generic` references and moved `json` onto `Request`.

java-client/src/main/java/org/opensearch/client/opensearch/generic/GenericBody.java
```diff
@@ -0,0 +1,90 @@
...
+     * Content type of this body
+     * @return content type
+     */
+    String contentType();
+
+    /**
+     * Gets the body as {@link InputStream}
+     * @return body as {@link InputStream}
+     */
+    InputStream body();
```
reta, line 63: @dblock one decision which is doubt at the moment is using `InputStream` vs `byte[]` to deal with response / request payload. Usage of `InputStream`  has a lot of benefits for a price of requiring to be more careful to deal with (the responses should be closed upon consumption). 

The `byte[]` would make iAPI much simpler but also consume content eagerly all the time (leaving no option to user to make a choice).

java-client/src/main/java/org/opensearch/client/opensearch/generic/GenericBodies.java
reta: Oh, that could be designed as a builder, sure! I was thinking about it but was focusing on getting the basics to work.

```
Requests
  .builder()
  .uri(...)
  .method(...)
  .json(...)
   ....
```
```

Comment (dblock) [reply java-client/src/main/java/org/opensearch/client/opensearch/generic/GenericBodies.java:9]: Great. If the entire `GenericBlah` can be hidden, you're winning.

File: java-client/src/test/java11/org/opensearch/client/opensearch/integTest/AbstractGenericClientIT.java [hunk 5]
```diff
@@ -0,0 +1,213 @@
...
+
+    @Test
+    public void shouldReturnSearchResults() throws Exception {
+        final String index = "search_request";
+        createIndex(index);
+
+        try (
+            Response response = javaClient().generic()
+                .execute(
+                    Requests.builder(javaClient())
```
dblock, line 34: `javaClient().builder()` ?

java-client/src/test/java11/org/opensearch/client/opensearch/integTest/AbstractGenericClientIT.java
```diff
@@ -0,0 +1,213 @@
...
+                )
+        ) {
+            assertThat(response.getStatus(), equalTo(200));
+            assertThat(response.getBody().isPresent(), equalTo(true));
+
+            final SearchResponse<?> r = Bodies.json(
+                response.getBody().get(),
+                SearchResponse.createSearchResponseDeserializer(JsonpDeserializer.of(ShopItem.class)),
+                javaClient()
+            );
```
dblock, line 71: Pretty funny looking, especially unclear why it needs the client. I think I'd want `response.getBody().json()`

reta: I wish I have a better way, but what we need is actually `JsonpMapper` instance which we could:
 - ask directly
 - ask from `Transport`
 - ask from `OpenSearchClient`

java-client/src/test/java11/org/opensearch/client/opensearch/integTest/AbstractGenericClientIT.java
reta: Right, I was thinking about that but decided not to:
  - this builder is only for generic client, so definitely not sufficient for sneaking into public API
  - for bodyless requests(), the javaClient() is not needed at all

java-client/src/test/java11/org/opensearch/client/opensearch/integTest/AbstractGenericClientIT.java
reta: Replaced the ApiClient with JsonpMapper, thanks @dblock !

java-client/src/test/java11/org/opensearch/client/opensearch/integTest/AbstractGenericClientIT.java
```diff
@@ -0,0 +1,225 @@
...
+            assertThat(response.getStatus(), equalTo(201));
+            assertThat(response.getBody().isPresent(), equalTo(true));
+        }
+    }
+
+    private void createIndex(String index) throws IOException {
+        try (
+            Response response = javaClient().generic()
+                .execute(
+                    Requests.builder(javaClient()._transport().jsonpMapper())
```
dblock, line 119: Is there to lazy load this given that we will call `.json` later?

java-client/src/test/java11/org/opensearch/client/opensearch/integTest/AbstractGenericClientIT.java
```diff
@@ -0,0 +1,225 @@
...
+
+        createTestDocuments(index);
+        refreshIndex(index);
+    }
+
+    private void refreshIndex(String index) throws IOException {
+        try (
+            Response response = javaClient().generic()
+                .execute(
+                    Requests.builder(javaClient()._transport().jsonpMapper()).endpoint("/" + index + "/_refresh").method("POST").build()
```
dblock, line 166: Strange here because there's no body, so I shouldn't need `jsonpMapper()`?

dblock (commented):
I think this is very close! Write some basic docs.

java-client/src/test/java11/org/opensearch/client/opensearch/integTest/AbstractGenericClientIT.java
reta: Yes, sure, I could move it off to `.json()` method call

java-client/src/test/java11/org/opensearch/client/opensearch/integTest/AbstractGenericClientIT.java
reta: Absolutely, thanks!

guides/generic.md
```diff
@@ -0,0 +1,79 @@
...
+The following sample code sends a simple request that does not require any payload to be provided.
+
+```java
+// compare with what the low level client outputs
+try (Response response = javaClient().generic().execute(Requests.builder().endpoint("/").method("GET").build())) {
+   // ...
+}
+```
+
+Please notice that the `Response` instance should be closed explicitly in order to free up any allocated resource (like response input streams or buffers). The generic client never interprets status codes and provides the direct access to the response as it was received from the server. This is responsibility of the caller to decide what should happen in case of unsuccessful invocations.
```
dblock, line 28: There are two statements here that are sort of independent. I would break it up with two separate examples on their own paragraphs.

guides/generic.md
```diff
@@ -0,0 +1,79 @@
...
+                .method("POST")
+                .json("{"
+                    + "    \"query\": {"
+                    + "        \"match_all\": {}"
+                    + "    }"
+                    + "}"
+                )
+                .build())) {
+   // Retrieve the response body as a simple string
+   final String body = response.getBody().map(Body::getAsString).orElse("");
```
dblock, line 47: What's the middle ground between string and POJO? Maybe add an example of sending and getting back structured JSON (fields, values)?

java-client/src/main/java/org/opensearch/client/opensearch/generic/GenericRequest.java
```diff
@@ -0,0 +1,172 @@
...
+     */
+    @Override
+    public int hashCode() {
+        return Objects.hash(method, endpoint, parameters, headers, body);
+    }
+
+    @Override
+    public String serialize(OutputStream out) {
+        if (getBody().isPresent() == false) {
+            throw new IllegalStateException("The request has no content body provided");
```
dblock, line 156: Check whether we terminate other errors with a `.`.

guides/generic.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Getting the Client
+The following sample code gets the `OpenSearchGenericClient` rom the `OpenSearchClient` instance.
+
+```java
+final OpenSearchGenericClient generic = javaClient().generic();
+```
+
+## Sending Simple Request
+The following sample code sends a simple request that does not require any payload to be provided.
```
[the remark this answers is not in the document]
reta: > This is a weird statement. Just say "Make a GET request without a body."?

We sadly have GET requests with a body [1], so could say something like `(typically,  GET requests)`, wdyt?

[1] https://www.elastic.co/guide/en/elasticsearch/reference/current/search-your-data.html

guides/generic.md
reta: It is given in the `[Sending JSON using POJOs]` section or you have something else in mind?

reta:
> I think the main thing we need is the intermediate sample/doc/test for sending and receiving structured JSON with a library (gson?).

I would be very reluctant to do show case that, for 2 reasons:
 - we always require the instance of the `JsonpMapper` mapper when creating `OpenSearchTransport` and AFAIK we don't have `JsonpMapper`  for GSON
 - there should be a reason to introduce POJOs that are not already provided by the `opensearch-java` as part of API, I am happy to craft the example if you point me out to those (otherwise we would just encourage to do unnecessary work)

guides/generic.md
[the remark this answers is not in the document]
dblock: Sure, NBD.

guides/generic.md
dblock: Something else. The section shows how to make a `CreateIndexResponse` and friends. I am saying I want a `JsonObjectBuilder` or similar.

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 
 ### Added
 - Add missed fields to MultisearchBody: seqNoPrimaryTerm, storedFields, explain, fields, indicesBoost ([#914](https://github.com/opensearch-project/opensearch-java/pull/914))
+- [FEATURE] Enable Generic HTTP Actions in Java Client ([#910](https://github.com/opensearch-project/opensearch-java/pull/910))
```
dblock, line 38: I would be more explicit here and say "Add generic HTTP request/response support" or such. This is the Java Client so we don't need to say it is.

dblock (dismissed):
Ok this is great. Did a careful review.

I want to make sure we discuss not wanting to handle errors. Why did you decide not to throw exceptions like the regular client? IMO users want to send POJOs vs. JSON, but error handling is sort of tangent. Doesn't it make it more complicated to reason about all clients when some don't handle errors like others?

reta:
> Ok this is great. Did a careful review.

❤️ thanks a lot @dblock 

> I want to make sure we discuss not wanting to handle errors. Why did you decide not to throw exceptions like the regular client? 

Sure, great question (as always). So here is my reasoning: with generic HTTP client, we give a freedom to basically deal with raw HTTP requests and responses (the generic client is literally just a messenger, it delivers some message and brings some response back, no implied knowledge what it is). Since we don't know the intent (may be caller expects 404), it is better to leave every decision to the caller, including how to handle status codes.

dblock:
> > Ok this is great. Did a careful review.
> 
> ❤️ thanks a lot @dblock
> 
> > I want to make sure we discuss not wanting to handle errors. Why did you decide not to throw exceptions like the regular client?
> 
> Sure, great question (as always). So here is my reasoning: with generic HTTP client, we give a freedom to basically deal with raw HTTP requests and responses (the generic client is literally just a messenger, it delivers some message and brings some response back, no implied knowledge what it is). Since we don't know the intent (may be caller expects 404), it is better to leave every decision to the caller, including how to handle status codes.

From the user requests I've seen the reasons to go to a lower level client are usually that we don't support a specific API, not to get more control. Does this change your mind? Is there a way to have both of these worlds?

dblock (approved):
I am going to merge, but let's be careful about shipping this - I still want to discuss error handling more.

reta:
Thanks @dblock 

> From the user requests I've seen the reasons to go to a lower level client are usually that we don't support a specific API, not to get more control. Does this change your mind? 

There are certainly pros and cons to both approaches. From implementation perspective, it is always safer to start simple (return the response as is) and add complexity progressively.

> Is there a way to have both of these worlds?

Yes, but I would better be guided by usage patterns to be fair: we have pretty extensive typed model, with full fledged error handling. It seems like it was not enough - the generic client is an opposite of that,

dblock:
> Thanks @dblock
> 
> > From the user requests I've seen the reasons to go to a lower level client are usually that we don't support a specific API, not to get more control. Does this change your mind?
> 
> There are certainly pros and cons to both approaches. From implementation perspective, it is always safer to start simple (return the response as is) and add complexity progressively.
> 
> > Is there a way to have both of these worlds?
> 
> Yes, but I would better be guided by usage patterns to be fair: we have pretty extensive typed model, with full fledged error handling. It seems like it was not enough - the generic client is an opposite of that,

Does the underlying HTTP client throw errors on network failure? This renders error handling inconsistent, doesn't it?

dblock:
>  it is always safer to start simple (return the response as is) and add complexity progressively.

In this case it's not actually true. If the server returns a 401 right now there's no error with the generic client. If we change this in the future it will be a breaking change.

reta:
> Does the underlying HTTP client throw errors on network failure?

Not really: the request was sent and response was received. Interpretation of the HTTP status code is subjective - the caller decides what he expected or not expected. Receiving the response is not a failure, everything else is.

dblock:
> Not really: the request was sent and response was received. Interpretation of the HTTP status code is subjective - the caller decides what he expected or not expected. Receiving the response is not a failure, everything else is.

This is not true. HTTP is very clear on what's an error and what is not!

dblock:
> We should take into account that we are not building generic HTTP client, but complementary OpenSearch client favour that should only be used when typed ones do not fit.

Correct. Error codes aren't a matter of type.

reta:
> > Not really: the request was sent and response was received. Interpretation of the HTTP status code is subjective - the caller decides what he expected or not expected. Receiving the response is not a failure, everything else is.
> 
> This is not true. HTTP is very clear on what's an error and what is not! Just like TCP (which raises timeout errors for example).

This is not what I meant, sorry about that (HTTP is very clear indeed),  interpretation of the HTTP status code is subjective from the caller perspective (I have tons of examples), but I think we should be discussing that in https://github.com/opensearch-project/opensearch-java/issues/927, thanks @dblock !

PR #999, opened 2024-05-24
Modify to not use URLEncodedUtils

### LINKED ISSUE #998: The SimpleEndpoint class in version 2.10.3 references the URLEncodedUtils class in the wrong library.
### What is the problem?
https://github.com/opensearch-project/opensearch-java/blob/2.10.3/java-client/src/main/java/org/opensearch/client/transport/endpoints/SimpleEndpoint.java

If you look at the link above, in version 2.10.3, the SimpleEndpoint class imports org.apache.http.client.utils.URLEncodedUtils.

### How can one reproduce the problem?
In Spring Boot 3.x, when a request is sent to the Opensearch java client without the `org.apache.httpcomponents:httpclient:4.x.x` dependency, the `SimpleEndpoint` cannot get the method of `URLEncodedUtils` and an error occurs.
```
Caused by: java.lang.NoSuchMethodError: 'java.lang.String org.apache.http.client.utils.URLEncodedUtils.formatSegments(java.lang.String[])'
	at org.opensearch.client.transport.endpoints.SimpleEndpoint.pathEncode(SimpleEndpoint.java:137) ~[opensearch-java-2.10.0.jar:na]
	at org.opensearch.client.opensearch.core.SearchRequest.lambda$static$2(SearchRequest.java:2293) ~[opensearch-java-2.10.0.jar:na]
	at org.opensearch.client.transport.endpoints.SimpleEndpoint.requestUrl(SimpleEndpoint.java:86) ~[opensearch-java-2.10.0.jar:na]
	at org.opensearch.client.transport.endpoints.DelegatingJsonEndpoint.requestUrl(DelegatingJsonEndpoint.java:55) ~[opensearch-java-2.10.0.jar:na]
	at org.opensearch.client.transport.httpclient5.ApacheHttpClient5Transport.prepareLowLevelRequest(ApacheHttpClient5Transport.java:565) ~[opensearch-java-2.10.0.jar:na]
	at org.opensearch.client.transport.httpclient5.ApacheHttpClient5Transport.performRequestAsync(ApacheHttpClient5Transport.java:169) ~[opensearch-java-2.10.0.jar:na]
	at org.opensearch.client.opensearch.OpenSearchAsyncClient.search(OpenSearchAsyncClient.java:1400) ~[opensearch-java-2.10.0.jar:na]
	at com.kurly.orderreceipt.core.infrastructure.repository.customerfrequentpurchase.CustomerFrequentPurchaseOSRepositoryImpl.aggregateOrderCompleted(CustomerFrequentPurchaseOSRepositoryImpl.kt:104) ~[main/:na]
	at com.kurly.orderreceipt.core.infrastructure.repository.customerfrequentpurchase.CustomerFrequentPurchaseOSRepositoryImpl.aggregateByDealProductNo(CustomerFrequentPurchaseOSRepositoryImpl.kt:68) ~[main/:na]
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103) ~[na:na]
	at java.base/java.lang.reflect.Method.invoke(Method.java:580) ~[na:na]
	at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:351) ~[spring-aop-6.1.4.jar:6.1.4]
	at org.springframework.aop.framework.ReflectiveMethodInvocation.invokeJoinpoint(ReflectiveMethodInvocation.java:196) ~[spring-aop-6.1.4.jar:6.1.4]
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:163) ~[spring-aop-6.1.4.jar:6.1.4]
	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.proceed(CglibAopProxy.java:765) ~[spring-aop-6.1.4.jar:6.1.4]
...
```

### What do you want us to do?
image

Please improve the part that uses `HttpClient` so that it can be done without using `HttpClient`. If there is no immediate alternative, please modify it to use `org.apache.hc.core5.net.URLEncodedUtils` like the file in the main branch.
https://github.com/opensearch-project/opensearch-java/blob/main/java-client/src/main/java/org/opensearch/client/transport/endpoints/SimpleEndpoint.java

I want to completely remove the `org.apache.httpcomponents:httpclient:4.x.x` dependency from Spring boot 3.

- Java 21
- Spring Boot 3.2.3
- opensearch-java 2.10.3

AntCode97:
### Description
SimpleEndpoint uses org.apache.hc.core5.net.URLEncodedUtils, but we deprecated it and modified it to use our own implementation of pathEncoder.

This allows us to remove the dependency on HttpClient.

I've referenced URLEncodedUtils and made it work exactly the same, optimizing it by removing parameters that aren't needed.

### Issues Resolved
Fixes #998 


java-client/src/main/java/org/opensearch/client/util/PathEncoder.java
```diff
@@ -0,0 +1,89 @@
+package org.opensearch.client.util;
```
dblock, line 1: This needs a license header.

If the code is copied from another APLv2 project, include their license header and a clear reference, too.

dblock (changes requested):
Add some basic tests for the public method?

java-client/src/main/java/org/opensearch/client/util/PathEncoder.java
reta: @dblock I don't feel right to copy the code like that: at the moment, we always have dependency on either AHC 4 or AHC 5, we could make `PathEncoder` to detect which one is on classpath and use the respective `URLEncodedUtils` .

dblock: We certainly have the right to copy code per license, but maybe the right solution is to refactor in a way that the implementation is not in `java-client/src/main/java/org/opensearch/client/transport/endpoints/SimpleEndpoint.java`, but in a version-specific transport implementation? @AntCode97 WDYT?

reta: > We certainly have the right to copy code per license,

Sure, no issues with that but maintaining it is not something we should take lightly

AntCode97: If we're going to continue to have dependencies on AHC, I think it's better to use URLEncodedUtils per version, with PathEncoder detecting the classpath, because implementing it would be difficult to maintain, and if a security vulnerability is found, it can't be fixed by simply raising the library version.

reta: > If we're going to continue to have dependencies on AHC, I think it's better to use URLEncodedUtils per version, 

@AntCode97 could you update your pull request to use take this path? thank you

AntCode97: Okay, but I'm busy today, so I'll update it next week.

reta: Please take our own pace, no rush, thank you!

AntCode97: While working on this, it occurred to me that modifying the SimpleEndpoint class in the 2.x branch to import org.apache.hc.core5.net.URLEncodedUtils would solve the problem.
This is because the main branch uses AHC5 and the 2.x branch uses both AHC4 and AHC5.

The opensearch-rest-client uses AHC4, but I don't think it would be a problem to change everything else to AHC5.

What do you think? @reta  If you agree, I'm going to close this PR, create a new branch based on the 2.x branch, and open the PR with the modification to import org.apache.hc.core5.net.URLEncodedUtils.

reta: > While working on this, it occurred to me that modifying the SimpleEndpoint class in the 2.x branch to import org.apache.hc.core5.net.URLEncodedUtils would solve the problem.

This is the large difference between 2.x and 3.x (main): 2.x uses by default RestClient and AHC 4.x, whereas 3.x by default uses AHC 5.x all the way. We cannot use AHC 5.x in 2.x, its presence is optional.

AntCode97: I got it. I will open PR soon

AntCode97: Okay. So, since 3.x uses AHC5 by default, I assume that AHC 4 and AHC 5 can exist optionally in 2.x only.
If so, should I close this PR and create a new branch based on the 2.x branch, detect classpasth, and open a PR that uses URLEncodedUtils per version?

AntCode97: I found a simpler way. 
It is to utilize java.net.URLEncoder. 
When I encode using URLEncoder, it passes the EndpointTest, so it looks like it's okay to use it instead of URLEncodedUtils.
I've fixed it in the code.

AntCode97: If you're okay with using`URLEncoder`, I'll open a PR to make the same change to the 2.x branch.
This will also cause a modification to `EndpointTest` in the 2.x branch, since `EndpointTest` has tests written based on the `org.apache.http.client.utils.URLEncodedUtils` class.

dblock:
This works for me. Let's add a line to CHANGELOG? This fixes a bug. You can say "Fix java.lang.NoSuchMethodError: org.apache.http.client.utils.URLEncodedUtils.formatSegments w/o httpclient" or something like that.

java-client/src/main/java/org/opensearch/client/util/PathEncoder.java
dblock: For 2.x we have a backport process, once this is merged it will attempt to automatically backport.

reta: @AntCode97 @dblock this is very risky change, one of the reason why others (like AHC, ...) have to reinvent this URL encoding / decoding / parsing utils is coming from the fact that Java's URL implementation is not compliant with RFC-3986 [1]. The consequences of that is - many edge (and no so edge) cases are not covered, and weird URLs are accepted where they should not. The AHC 5.x seems to keep conformance [2].

[1] https://datatracker.ietf.org/doc/html/rfc3986
[2] https://github.com/apache/httpcomponents-core/pull/205

dblock: @AntCode97 Thanks so much for hanging in here with us for what is seemingly a simple change. It's really not! Appreciate your time and work.

AntCode97: After performing a clean build, I was unable to build.
The reason is that I can't get `org.apache.lucene:lucene-test-framework:9.11.0-snapshot-fb97840`, which is used by `org.opensearch.test:framework:3.0.0-SNAPSHOT`.
I had no problem until yesterday, but suddenly this happened. @reta, do you know anything about this?

reta: > I had no problem until yesterday, but suddenly this happened. @reta, do you know anything about this?

@AntCode97 yes, I do, but I am very much surprised you are impacted. This is by and large "normal" dependency update cycle (we bump Apache Lucene snapshots). The repo URLs are coming from OpenSearch Gradle plugin(s) which should fetch the artifacts transparently but it seems like not the case for you.

reta: I have no issues running the build on clean box (Docker):

```
$ docker run -it gradle:8.7.0-jdk11 sh                                                                                                                                                                                                                                                                        
# git clone https://github.com/opensearch-project/opensearch-java/                                                                                                                                                                                                                                                                                    
Cloning into 'opensearch-java'...                                                                                                                                                                                                                                                                                                                     
remote: Enumerating objects: 64193, done.        
remote: Counting objects: 100% (1086/1086), done.                                                                                                                                                                                                                                                                                                     
remote: Compressing objects: 100% (530/530), done.
remote: Total 64193 (delta 410), reused 926 (delta 292), pack-reused 63107                                                                                                                                                                                                                                                                            
Receiving objects: 100% (64193/64193), 26.29 MiB | 7.66 MiB/s, done.
Resolving deltas: 100% (54268/54268), done.
# cd opensearch-java        
# ./gradlew check           
Downloading https://services.gradle.org/distributions/gradle-8.7-bin.zip
............10%.............20%.............30%.............40%............50%.............60%.............70%.............80%.............90%............100%             
Starting a Gradle Daemon (subsequent builds will be faster)
                                                                                     
> Configure project :java-client                                                                                                                                           
=======================================                                                                                                                                                                                                                                                                                                               
  Runtime JDK Version   : 11                                                                                                                                               
  Gradle JDK Version    : 11                                                                                                                                                                                                                                                                                                                          
=======================================                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                      
> Task :java-client:compileJava                                                                                                                                                                                                                                                                                                                       
Note: Some input files use or override a deprecated API.                                                                                                                   
Note: Recompile with -Xlint:deprecation for details.                                                                                                                       
Note: Some input files use unchecked or unsafe operations.                                                                                                                 
Note: Recompile with -Xlint:unchecked for details.                                                                                                                         
                                                                                     
> Task :java-client:compileTestJava                                                                                                                                        
Note: Some input files use or override a deprecated API.                                                                                                                                                                                                                                                                                              
Note: Recompile with -Xlint:deprecation for details.                                                                                                                                                                                                                                                                                                  
                                                                                     
> Task :samples:compileJava                                                                                                                                                                                                                                                                                                                           
Note: /home/gradle/opensearch-java/samples/src/main/java/org/opensearch/client/samples/Search.java uses unchecked or unsafe operations.                                                                                                                                                                                                               
Note: Recompile with -Xlint:unchecked for details.                                                                                                                                                                                                                                                                                                    
                                                                                     
Deprecated Gradle features were used in this build, making it incompatible with Gradle 9.0.                                                                                
                                                                                                                                                                           
You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.                                                                                                                                                                                                             

For more on this, please refer to https://docs.gradle.org/8.7/userguide/command_line_interface.html#sec:command_line_warnings in the Gradle documentation.
                                                                                                                                                                                                                                                                                                                                                      
BUILD SUCCESSFUL in 2m 24s                                                            
14 actionable tasks: 14 executed                                                      

```

AntCode97: I didn't change anything, but now it builds fine again. 
I think there may have been a temporary issue with https://aws.oss.sonatype.org/content/repositories/snapshots.
I'll have to look into this a bit more, but in any case, I'm back to work.

java-client/src/main/java/org/opensearch/client/util/PathEncoder.java
```diff
@@ -0,0 +1,50 @@
...
+ * compatible open source license.
+ */
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.util.Collections;
+
+public class PathEncoder {
```
reta, line 17: @AntCode97 thank you, the reflection APIs are discourage from use (see please  https://openjdk.org/jeps/416), we need something like that (implementation below uses `MethodHandle`s and initializes only once at class initialization time, JVM claims to be able to heavily optimize static final `MethodHandle`s).

```suggestion
/*
 * SPDX-License-Identifier: Apache-2.0
 *
 * The OpenSearch Contributors require contributions made to
 * this file be licensed under the Apache-2.0 license or a
 * compatible open source license.
 */

package org.opensearch.client.util;

import java.lang.invoke.MethodHandle;
import java.lang.invoke.MethodHandles;
import java.lang.invoke.MethodType;
import java.nio.charset.Charset;
import java.nio.charset.StandardCharsets;
import java.util.Collections;

public class PathEncoder {
    private final static String HTTP_CLIENT4_UTILS_CLASS = "org.apache.http.client.utils.URLEncodedUtils";
    private final static String HTTP_CLIENT5_UTILS_CLASS = "org.apache.hc.core5.net.URLEncodedUtils";
    private final static MethodHandle FORMAT_SEGMENTS_MH;
    
    static {
        Class<?> clazz = null;
        try {
            // Try Apache HttpClient5 first since this is a default one
            clazz = Class.forName(HTTP_CLIENT5_UTILS_CLASS);
        } catch (final ClassNotFoundException ex) {
            try {
                // Fallback to Apache HttpClient4
                clazz = Class.forName(HTTP_CLIENT4_UTILS_CLASS);
            } catch (final ClassNotFoundException ex1) {
                clazz = null;
            }
        }
        
        if (clazz == null) {
            throw new IllegalStateException("Either '" + HTTP_CLIENT5_UTILS_CLASS + "' or '" + HTTP_CLIENT4_UTILS_CLASS + "' is required by not found on classpath");
        }
        
        try {
            FORMAT_SEGMENTS_MH = MethodHandles
                .lookup()
                .findStatic(clazz, "formatSegments", MethodType.methodType(String.class, Iterable.class, Charset.class));
        } catch (final NoSuchMethodException | IllegalAccessException ex) {
            throw new IllegalStateException("Unable to find 'formatSegments' method in " + clazz + " class");
        }
    }

    public static String encode(String uri) {
        try {
            return (String) FORMAT_SEGMENTS_MH.invoke(Collections.singletonList(uri), StandardCharsets.UTF_8);
        } catch (final Throwable ex) {
            throw new RuntimeException("Unable to encode URI: " + uri, ex);
        }
    }
}

```

AntCode97: Your code looks very good. I modified it as you wrote.

reta:
Thank you @AntCode97 , I think we only are missing entry in CHANGELOG.md under [Unreleased 2.x] section, could you please add one? Thank you.

AntCode97:
> Thank you @AntCode97 , I think we only are missing entry in CHANGELOG.md under [Unreleased 2.x] section, could you please add one? Thank you.

Okay. I've added it.

reta:
@AntCode97 my apologies, there are conflicts and and DCO check is failing

AntCode97:
> @AntCode97 my apologies, there are conflicts and and DCO check is failing

Oh, I forgot about that, all fixed.

reta (approved):

dblock (approved):
@AntCode97 Thank you!

It solves the problem so let's merge.

I dislike the fact that we're using reflection very much. I would have much rather preferred either a neutral implementation or a change in class hierarchy that ends up calling the encode method in the right library. Since I'm only the talking head I opened https://github.com/opensearch-project/opensearch-java/issues/1008. Maybe I'll get to it or maybe you or @reta feel like a challenge :)

PR #1007, opened 2024-05-31
Add text embedding processor

### LINKED ISSUE #1005: [FEATURE] Support text embedding processor
Not sure if I missed something in the `org.opensearch.client.opensearch.ingest.Processor` class, but it seems there's no constructor for the [text embedding](https://opensearch.org/docs/latest/ingest-pipelines/processors/text-embedding/) processor as specified [here](https://opensearch.org/docs/latest/ingest-pipelines/processors/text-embedding/). But, looking at the [OpenAPI spec](https://github.com/opensearch-project/opensearch-api-specification/blob/69247b090197cb49b159dd68277002417fbab84f/spec/schemas/ingest._common.yaml#L33), it seems it's missing from there as well.

Am I missing something? Does the Java client depend on the OpenAPI spec?

miguel-vila:
### Description

Adds the text embedding processor: https://opensearch.org/docs/latest/ingest-pipelines/processors/text-embedding/

### Issues Resolved

Should close #1005

### DIFF
--- a/java-client/src/main/java/org/opensearch/client/opensearch/ingest/TextEmbeddingProcessor.java
+++ b/java-client/src/main/java/org/opensearch/client/opensearch/ingest/TextEmbeddingProcessor.java
@@ -1,0 +1,214 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch B.V. licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


dblock (changes requested):
Looks pretty straightforward. Please add a test?

java-client/src/main/java/org/opensearch/client/opensearch/ingest/TextEmbeddingProcessor.java
```diff
@@ -0,0 +1,214 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+/*
```
reta, line 9: Sorry was this class copied for Elasticsearch codebase? If not, we could only keep SPDX License

miguel-vila: yeah, probably something I copied and pasted from another processor class, fixed at: [752df7b](https://github.com/opensearch-project/opensearch-java/pull/1007/commits/752df7be6d17e8fdcb674adf61cce0efea77bd99)

dblock (approved):

reta (approved):

PR #1052, opened 2024-06-26
Begin generating client implementations

Xtansia:
### Description
Starts generating client class implementations, modifies existing version of the class to be an abstract base class that the generated one extends from.

### DIFF
--- a/java-client/src/generated/java/org/opensearch/client/opensearch/OpenSearchAsyncClient.java
+++ b/java-client/src/generated/java/org/opensearch/client/opensearch/OpenSearchAsyncClient.java
@@ -34,42 +34,39 @@
 // THIS CODE IS GENERATED. MANUAL EDITS WILL BE LOST.
 //----------------------------------------------------
 
-package org.opensearch.client.opensearch.core;
+package org.opensearch.client.opensearch;
 
+import java.io.IOException;
+import java.util.concurrent.CompletableFuture;
 import javax.annotation.Generated;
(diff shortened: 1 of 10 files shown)


java-client/src/generated/java/org/opensearch/client/opensearch/OpenSearchAsyncClient.java
```diff
@@ -0,0 +1,72 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+/*
```
reta, line 9: @Xtansia my apologies for being late, why do we use ES license here? This is 100% OS generated code, right?

Xtansia: It's kind of a weird situation because the code it's generating is attempting to match the existing client classes which are licensed to ES. We had a similar case with .NET: https://github.com/opensearch-project/opensearch-net/pull/320#discussion_r1302241845

Though here the generator was written completely from scratch, I'd assume as the output is (near) identical it's probably legally basically the same as having just copy pasted parts of the file (with extra steps). 

Thoughts @dblock?

dblock: We are told to leave these in generated files if they are substantially not different.

PR #1061, opened 2024-07-01
Add workflow to ensure generated code is up to date

Xtansia:
### Description
Add workflow to ensure generated code is up to date

Example of failing run: https://github.com/opensearch-project/opensearch-java/actions/runs/9736512856/job/26867247859?pr=1061

### DIFF
--- a/.github/workflows/code-generation.yml
+++ b/.github/workflows/code-generation.yml
@@ -1,0 +1,34 @@
+name: Code Generation
+
+on: [pull_request]
+
+jobs:
+  up_to_date:
+    name: Ensure Generated Code Up To Date
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Java Client
+        uses: actions/checkout@v4
+
+      - name: Set up JDK
+        uses: actions/setup-java@v4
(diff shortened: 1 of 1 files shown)


dblock:
This means that anyone contributing to the client will have to re-run the generator. So, if I am fixing an unrelated bug, I still will get a failure here. In that case, does an automatic workflow that runs on a cron to generate a PR update like https://github.com/opensearch-project/opensearch-py/pull/767 make more sense?

Xtansia:
> This means that anyone contributing to the client will have to re-run the generator. So, if I am fixing an unrelated bug, I still will get a failure here. In that case, does an automatic workflow that runs on a cron to generate a PR update like [opensearch-project/opensearch-py#767](https://github.com/opensearch-project/opensearch-py/pull/767) make more sense?

The generator has a local copy of the spec checked into the repo (and doesn't yet have a auto-download mechanism), so this will always run against that copy, meaning it'll only detect changes if the spec or generator logic has been changed. I'm going to add a cron job later for auto-fetching & updating.

dblock (approved):

PR #1064, opened 2024-07-02
588 - [FEATURE] serialize requests to JSON

Jai2305:
### Serialize requests to JSON
* This pull request adds an interface `PlainJsonSerializable ` extending  `JsonpSerializable `  having a default method to streamline serialization process ( get a string representation of  instances of all classes implementing PlainJsonSerializable interface ) , which abstracts away the implementation of having to call serialize method with mapper and generator.
* As a result the classes implementing `JsonpSerializable ` may now implement  `PlainJsonSerializable ` if they are required to carry the added default method  otherwise  they may implement `JsonpSerializable `
* Added test cases to ensure correctness of the code.

- [x] Self Check
- [x] Added Test Cases
- [x] ChangeLog.md

### SignOff - 
I understand and agree that this project and the contribution are public and that a record of the contribution (including
all personal information I submit with it, including my sign-off) is maintained indefinitely and may be redistributed consistent with this project or the open source license(s) involved.
Please check Developer Certificate of Origin [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).


java-client/src/main/java/org/opensearch/client/json/JsonpSerializable.java
```diff
@@ -33,11 +33,33 @@
...
+     * A default method which returns string representation for the instances of classes
+     * implementing JsonpSerializable interface.<br>
+     * Usage : Eg for SearchRequest.class<pre>{@code SearchRequest implements JsonpSerializable{..}
+     * SearchRequest searchRequest = SearchRequest.of(request -> request...);
+     * String searchRequestString = searchRequest.writeValueAsString();}</pre>  <br>
+     *
+     */
+    default String writeValueAsString() {
+        StringWriter writer = new StringWriter();
+        try (JsonGenerator generator = JacksonJsonProvider.provider().createGenerator(writer)) {
```
reta, line 59: This is not going to work (since Jackson is not required), could you please port https://github.com/opensearch-project/spring-data-opensearch/blob/main/spring-data-opensearch/src/main/java/org/opensearch/data/client/osc/JsonpUtils.java - the idea there is to use the provider from the classpath (either Jakarta or Jackson) but not hardcode any/

reta (changes requested):

java-client/src/main/java/org/opensearch/client/json/JsonpSerializable.java
Jai2305: Ohh got it, ported default provider and mapper from [PR](https://github.com/opensearch-project/spring-data-opensearch/blob/main/spring-data-opensearch/src/main/java/org/opensearch/data/client/osc/JsonpUtils.java), rather than hardcoding the service providers.

java-client/src/test/java/org/opensearch/client/opensearch/json/JsonpSerializableTest.java
```diff
@@ -0,0 +1,94 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+/*
```
reta, line 9: Please keep only SPDX license, this is a new code, no attribution to Elastic

java-client/src/main/java/org/opensearch/client/json/JsonpSerializable.java
```diff
@@ -33,11 +33,30 @@
...
+    /**
+     * A default method which returns string representation for the instances of classes
+     * implementing JsonpSerializable interface.<br>
+     * Usage : Eg for SearchRequest.class<pre>{@code SearchRequest implements JsonpSerializable{..}
+     * SearchRequest searchRequest = SearchRequest.of(request -> request...);
+     * String searchRequestString = searchRequest.writeValueAsString();}</pre>  <br>
+     *
+     */
+    default String writeValueAsString() {
+        StringWriter writer = new StringWriter();
```
reta, line 54: Please close the writer:

```
    try (StringWriter writer = new StringWriter()) {
        try (JsonGenerator generator = JsonpUtils.DEFAULT_PROVIDER.createGenerator(writer)) {
            serialize(generator, JsonpUtils.DEFAULT_JSONP_MAPPER);
        }

        return writer.toString();
        }
```

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 ### Added
 - Document HTTP/2 support ([#330](https://github.com/opensearch-project/opensearch-java/pull/330))
 - Add support for phase_took & search_pipeline request params ([#1036](https://github.com/opensearch-project/opensearch-java/pull/1036))
+-  Add a serializer method for classes implementing JsonpSerializable.([#1064](https://github.com/opensearch-project/opensearch-java/pull/1064))
```
reta, line 15:
```suggestion
-  Add a serializer method for classes implementing JsonpSerializable.([#1064](https://github.com/opensearch-project/opensearch-java/pull/1064))
```

java-client/src/main/java/org/opensearch/client/json/JsonpSerializable.java
Jai2305: StringWriter does not hold any IO resources, they maintain StringBuffer , and closing them does not accomplishes anything , it will rely on Java's garbage collection for cleanUp

CHANGELOG.md
Jai2305: yes, let me fix that real quick.

java-client/src/test/java/org/opensearch/client/opensearch/json/JsonpSerializableTest.java
Jai2305: Alright , let me fix it.

java-client/src/main/java/org/opensearch/client/json/JsonpSerializable.java
reta: `StringWriter` implements `Closeable` and should be closed upon completion.  The fact that currently the `close()` method is noop does not mean the resource management has to be ignored.

Jai2305: Okay, as you said right now close() is noop, but it does throw IO Exception , should it be an unchecked exception in which case the clients will have to handle it whenever they call the method or you think we should handle it in a specific way, any thoughts ?

reta:
```
 try (StringWriter writer = new StringWriter()) {
        try (JsonGenerator generator = JsonpUtils.DEFAULT_PROVIDER.createGenerator(writer)) {
            serialize(generator, JsonpUtils.DEFAULT_JSONP_MAPPER);
        }

        return writer.toString();
    } catch (IOException ex) {
        thrown new UncheckedIOException(ex);
   }
```

java-client/src/main/java/org/opensearch/client/json/JsonpSerializable.java
```diff
@@ -33,11 +33,34 @@
...
+
+    /**
+     * A default method which returns string representation for the instances of classes
+     * implementing JsonpSerializable interface.<br>
+     * Usage : Eg for SearchRequest.class<pre>{@code SearchRequest implements JsonpSerializable{..}
+     * SearchRequest searchRequest = SearchRequest.of(request -> request...);
+     * String searchRequestString = searchRequest.writeValueAsString();}</pre>  <br>
+     *
+     */
+    default String writeValueAsString() {
```
reta, line 55: I think we have to find better name (`toJson`? `asJson`?) to this method and extract it to a different interface (fe `PlainJsonSerializable`), it does not belong to `JsonpSerializable` anyhow (it uses separate Jsonp provider):

```
public interface PlainJsonSerializable extends JsonpSerializable {
    default String toJson() {
         ...
    }
}
```

Yes, it uses opt-in model (we would need to update classes where we need the plain JSON serialization) but it clearly separates 2 different serialization mechanisms.

Jai2305: * For name I think `toJson` should do fine right ? short and defines purpose .
*  For design, my initial idea was to make a default method to handle serialization for all the classes implementing JsonpSerializable , without having to implement any other interface or calling a Utils class, clients can just call searchResponse.toJson() and done , which uses default mapper and abstracts away the procedure of having to call serialize by there own generator and mapper .

Jai2305: Okay nvm toJson results in conflict coz JsonData which implements JsonpSerializable has a method implemented with same name

dblock:
Btw, not sure this will work as part of this PR, but we also want serialization of ndjson requests (e.g. bulk).

java-client/src/main/java/org/opensearch/client/json/JsonpSerializable.java
reta: > Also serialize can still be used with mapper and generator, our method is just a default implementation of a more generalize method.

That's the thing, the same API (interface) bundles 2 different serialization implementation, and there is no way to optout from it.

Jai2305: * I see the point of no way of opting out and separation of concerns  but is it not ideal in this case, do you see any disadvantages in future for this approach.
* Also  for Opt In using separate interface, do we need to re assess the classes we want to implement the new interface or all current classes which are implementing JsonpSerializable.
* Related to the previous point , if new classes are introduced implementing JsonpSerializable we will have to asses, if we want to OptIn right ?

reta: > I see the point of no way of opting out and separation of concerns but is it not ideal in this case, do you see any disadvantages in future for this approach.

I don't see any disadvantages to be fair

> Also for Opt In using separate interface, do we need to re assess the classes we want to implement the new interface or all current classes which are implementing JsonpSerializable.

Correct, this is one time job but I truly believe it is worth the effort to keep APIs clean and concise. 

> Related to the previous point , if new classes are introduced implementing JsonpSerializable we will have to asses, if we want to OptIn(Implement new Interface) right ?

The code is supposed to be generated: as far as generator is aware of this change, it won't be  noticeable.

reta:
> Btw, not sure this will work as part of this PR, but we also want serialization of ndjson requests (e.g. bulk).

There are 2 ways to tackle that: 
- since this is not a correct Json (bu ndjson), we could not implement PlainJsonSerializable (but fe have PlainNdJsonSerializable)
- do override the `toJsonString` for `BulkRequest` & co

Jai2305:
> There are 2 ways to tackle that (off the top of my head, there are more for sure):
> 
> * since this is not a correct Json (but ndjson), we could not implement PlainJsonSerializable (but fe have PlainNdJsonSerializable)
> * do override the `toJsonString` for `BulkRequest` & co

* For both of the above discussions, why not we make a general interface, where we will implement only serialize method , it will be one level above `JsonpSerializable `, and `NdJsonSerializable `.
* This way we have an optout option from `JsonpSerializable  ` default method and in future we could implement  different Serializable interfaces if the need arise for eg: `NdJsonSerializable `
``` Java
public interface JsonSerializable {
    void serialize(JsonGenerator generator, JsonpMapper mapper);
}
``` 
``` Java
public interface JsonpSerializable implements JsonSerializable  {
    default String toJsonString() {
        try (StringWriter writer = new StringWriter()) {
            try (JsonGenerator generator = JsonpUtils.DEFAULT_PROVIDER.createGenerator(writer)) {
                serialize(generator, JsonpUtils.DEFAULT_JSONP_MAPPER);
            }
            return writer.toString();
        } catch (IOException ex) {
            throw new UncheckedIOException(ex);
        }
    }
}
```
``` Java
public interface NDJsonSerializable implements JsonSerializable  {
   .... 
}
```

reta:
> Yes , I am suggesting a new parent interface and keep `JsonpSerializable` extending that new interface

By why? Building on top of `JsonpSerializable` looks very straightforward, no?

Jai2305:
> By why? Building on top of `JsonpSerializable` looks very straightforward, no?

By defining a top layer , we won't be touching the  current implementations and uses (all classes ) of JsonpSerializable , at the same time provide an option to opt out from the default method , which was our concern.
Won't this accomplish the same thing?
please correct me if I am misunderstanding anything.

reta:
> By defining a top layer , we won't be touching the current implementations and uses (all classes ) of JsonpSerializable , at the same time provide an option to opt out from the default method , which was our concern.

I think it circles back to the reasons why not to do that (https://github.com/opensearch-project/opensearch-java/pull/1064#discussion_r1664259465). The goal is to have 2 separate hierarchies, `PlainJsonSerializable` & `PlainNdJsonSerializable`, so `toJsonString` would be **only implemented** once by the right class in question.

Jai2305:
> The goal is to have 2 separate hierarchies, `PlainJsonSerializable` & `PlainNdJsonSerializable`, so `toJsonString` would be **only implemented** once by the right class in question (if we change `JsonpSerializable`, every single class will inherit its `toJsonString` right away, including the ones that use NdJson type of serialization)

* One last clarification, we already have NdJsonpSerializable interface  right ?
```  Java
public interface NdJsonpSerializable {
    Iterator<?> _serializables();
}
```
how does this fit in the suggested model.

reta:
>   One last clarification, we already have NdJsonpSerializable interface right ? how does this fit in the suggested model.

Correct, it fits super well I believe, for example:

```
public interface PlainNdJsonSerializable extends NdJsonpSerializable, JsonpSerializable  {
    default Iterator<String> toJsonStrings() {
     ....
    }
}
```

Jai2305:
@reta  added a separate commit  for a new interface `PlainJsonSerializable ` , which extends `JsonpSerializable` and has the default new method 
* As of now I have made every class implementing `JsonpSerializable`, implement `PlainJsonSerializable ` , as a result many files were changed due to the import statement 
* Only concern is if we want to Opt out any class from `PlainJsonSerializable  `and  implement `JsonpSerializable` , so that we do not inherit the newly added default method, as there are many classes , how will we determine which class to opt in or out.

reta:
Thanks @Jai2305 

> * As of now I have made every class implementing `JsonpSerializable`, implement `PlainJsonSerializable ` , as a result many files were changed due to the import statement

Expected, thank you!

> Only concern is if we want to Opt out any class from PlainJsonSerializable now and implement JsonpSerializable , so that we do not inherit the newly added default method, as there are many classes , how will we determine which class to opt in or out ?

That could be done on case by case I believe , the one of such is `BulkRequest` fe (it is NdJson)

java-client/src/main/java/org/opensearch/client/json/JsonpSerializable.java
```diff
@@ -40,4 +40,5 @@
 public interface JsonpSerializable {
 
     void serialize(JsonGenerator generator, JsonpMapper mapper);
+
```
reta, line 43: Super nit, could you please revert this change? (so the commit would be clean)

java-client/src/main/java/org/opensearch/client/json/PlainJsonSerializable.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package org.opensearch.client.json;
+
+import jakarta.json.stream.JsonGenerator;
+import java.io.IOException;
+import java.io.StringWriter;
+import java.io.UncheckedIOException;
+
+public interface PlainJsonSerializable extends JsonpSerializable {
```
reta, line 16: :+1:

java-client/src/test/java/org/opensearch/client/opensearch/json/JsonpSerializableTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.util.Collections;
+import org.junit.Assert;
+import org.junit.Test;
+import org.opensearch.client.opensearch._types.FieldValue;
+import org.opensearch.client.opensearch._types.Result;
+import org.opensearch.client.opensearch.core.IndexResponse;
+import org.opensearch.client.opensearch.core.SearchRequest;
+
+public class JsonpSerializableTest extends Assert {
```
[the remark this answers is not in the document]
Jai2305: Fixed according to suggestions , also changed CHANGELOG.md to reflect the addition of new interface, I will have a check on the code, but please let me know , if you find anything out of order.

reta (dismissed):

reta:
@dblock could you take a look as well please? thank you!

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 ### Added
 - Document HTTP/2 support ([#330](https://github.com/opensearch-project/opensearch-java/pull/330))
 - Add support for phase_took & search_pipeline request params ([#1036](https://github.com/opensearch-project/opensearch-java/pull/1036))
+- Add an interface PlainJsonSerializable extending JsonpSerializable having default serialize method.([#1064](https://github.com/opensearch-project/opensearch-java/pull/1064))
```
dblock, line 15: Remove the period and add a space to match other CHANGELOG lines.

I think we should explain the purpose of this `Add support for serializing requests and responses to plain JSON and nd-JSON`?

java-client/src/main/java/org/opensearch/client/json/JsonpUtils.java
```diff
@@ -59,6 +61,39 @@ public class JsonpUtils {
...
+
+        @Override
+        public <T> void serialize(T value, JsonGenerator generator) {
+            if (value instanceof JsonpSerializable) {
+                ((JsonpSerializable) value).serialize(generator, this);
+                return;
+            }
+
+            throw new JsonException(
+                "Cannot find a serializer for type " + value.getClass().getName() + ". Consider using a full-featured JsonpMapper"
```
dblock, line 85: Does this need a period at the end (check other errors)?

java-client/src/main/java/org/opensearch/client/opensearch/_types/query_dsl/XyShapeFieldQuery.java
```diff
@@ -35,13 +35,18 @@
 import jakarta.json.stream.JsonGenerator;
 import java.util.function.Function;
 import javax.annotation.Nullable;
-import org.opensearch.client.json.*;
+import org.opensearch.client.json.JsonData;
+import org.opensearch.client.json.JsonpDeserializer;
+import org.opensearch.client.json.JsonpMapper;
+import org.opensearch.client.json.ObjectBuilderDeserializer;
+import org.opensearch.client.json.ObjectDeserializer;
+import org.opensearch.client.json.PlainJsonSerializable;
```
dblock, line 43: For later, we should have a lint violation for non-expanded imports.

dblock (changes requested):
I have some nits on CHANGELOG and errors.

More importantly, can we update https://github.com/opensearch-project/opensearch-java/blob/main/USER_GUIDE.md or one of the [guides](https://github.com/opensearch-project/opensearch-java/tree/main/guides) with this (possibly a new one called json.md)? Even something basic would be great.

java-client/src/main/java/org/opensearch/client/json/JsonpUtils.java
Jai2305: I found many error messages not ending with a "." , so I went with that, but I am adding it because I think it's more neat that way.

dblock: Would be great to standardize in a separate PR, wink wink :)

CHANGELOG.md
Jai2305: Sorry for the delay, I have made a change on this, tried to keep it short and clear in meaning, please let me know if any work is remaining , and when can we merge this PR and [withJson ](https://github.com/opensearch-project/opensearch-java/pull/657/files), very much needed .
thanks a lot  :)

dblock (changes requested):
I am good with merging this. 

There's a build failure with the generator, which I think is legit, and a template needs to be modified, please take a look?

Thank you!

Jai2305:
> There's a build failure with the generator, which I think is legit, and a template needs to be modified, please take a look?
- Yes , I think I accidently changed some generated files, Fixed it.  
But the white source Security check is failing and the only message that I see is 
` Oops! An error occurred while running the Security Check.The Scanner was unable to connect to the repository or branch and clone it.` ,  is it trying to connect with my repo, sorry not familiar with this myself .

dblock:
> > There's a build failure with the generator, which I think is legit, and a template needs to be modified, please take a look?
> 
> * Yes , I think I accidently changed some generated files, Fixed it.
>   But the white source Security check is failing and the only message that I see is
>   ` Oops! An error occurred while running the Security Check.The Scanner was unable to connect to the repository or branch and clone it.` ,  is it trying to connect with my forked repo? sorry not familiar with this myself .

Ignore it. I think someone was looking into it.

dblock (approved):

dblock:
Merged, thanks.

@Jai2305, will you document this in USERS_GUIDE, maybe a new `json.md` guide in `guides` and a working sample?

PR #1065, opened 2024-07-03
Add workflow to auto-update codegen

Xtansia:
### Description

### Issues Resolved


java-codegen/opensearch-openapi.yaml
reta, line 1: I don't think we should store the spec file in this repo

java-codegen/src/main/java/org/opensearch/client/codegen/openapi/MimeType.java
```diff
@@ -14,8 +14,11 @@
 import org.opensearch.client.codegen.utils.Strings;
 
 public enum MimeType {
+    Cbor("application/cbor"),
```
reta, line 17: Smile seems to be missing (we do support it)

java-codegen/opensearch-openapi.yaml
dblock: I think the idea is that another workflow updates it. The spec changes often and if you were to add a feature or fix a bug you'd always have the generator workflow fail in your PR.

java-codegen/src/main/java/org/opensearch/client/codegen/openapi/MimeType.java
dblock: Note that the spec doesn't have it yet, https://github.com/opensearch-project/opensearch-api-specification/issues/372.

java-codegen/opensearch-openapi.yaml
reta: Oh I see, should we use the commit hash than?

dblock: We auto-draft a release which I don't think has a commit hash :( 

https://github.com/opensearch-project/opensearch-api-specification/releases/download/main-latest/opensearch-openapi.yaml

https://github.com/opensearch-project/opensearch-api-specification/releases/tag/main-latest

reta: Screenshot from 2024-07-03 15-25-25

reta: (always :-))

dblock: That's the commit hash of the last revision in this release. What's a permalink for https://github.com/opensearch-project/opensearch-api-specification/releases/download/main-latest/opensearch-openapi.yaml?

reta: Got it now, I thought the OpenAPI spec is stored in `opensearch-api-specification` repo but it is not (just published upon release), could we publish the versioned spec? (0.0.1,....)

java-codegen/build.gradle.kts
```diff
@@ -66,9 +68,16 @@ application {
     )
 }
 
+val localSpecification = "$projectDir/opensearch-openapi.yaml"
+
+tasks.create<Download>("downloadLatestSpec") {
+    src("https://github.com/opensearch-project/opensearch-api-specification/releases/download/main-latest/opensearch-openapi.yaml")
```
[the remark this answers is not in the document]
dblock: In some hypothetical future when the spec is complete we will make releases every time a new API releases with OpenSearch. How strongly do you feel about this being checked in in the meantime?

java-codegen/opensearch-openapi.yaml
dblock: If you feel strongly about it, yes. It's iterating fast, but if you want to add a release workflow to https://github.com/opensearch-project/opensearch-api-specification @Xtansia go for it.

java-codegen/build.gradle.kts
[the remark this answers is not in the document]
reta: We are in early stages indeed, so if that is the way to move forward (for now), no hard feelings

java-codegen/opensearch-openapi.yaml
reta: I think at least attaching the commit it to the release/draft would be a great, let me take it

dblock (approved):

java-codegen/opensearch-openapi.yaml
dblock: https://github.com/opensearch-project/opensearch-api-specification/issues/380

PR #1072, opened 2024-07-10
Add failing test for GetRequest w/ AwsSdk2Transport (#1068)

workmanw:
This adds a failing test for #1068

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 
 ### Fixed
 - Deserialize aggregation containing a parent aggregation ([#706](https://github.com/opensearch-project/opensearch-java/pull/706))
+- Deserialize error response when using AwsSdk2Transport ([#1068](https://github.com/opensearch-project/opensearch-java/pull/1068))
 
 ### Security
 
--- a/java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
+++ b/java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
@@ -48,6 +48,7 @@ import org.opensearch.client.transport.TransportException;
 import org.opensearch.client.transport.TransportOptions;
 import org.opensearch.client.transport.endpoints.BooleanEndpoint;
 import org.opensearch.client.transport.endpoints.BooleanResponse;
+import org.opensearch.client.util.MissingRequiredPropertyException;
 import org.opensearch.client.util.OpenSearchRequestBodyBuffer;
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
 import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider;
@@ -64,6 +65,7 @@ import software.amazon.awssdk.http.SdkHttpResponse;
 import software.amazon.awssdk.http.async.AsyncExecuteRequest;
 import software.amazon.awssdk.http.async.SdkAsyncHttpClient;
 import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.utils.IoUtils;
 import software.amazon.awssdk.utils.SdkAutoCloseable;
 
 /**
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 
 ### Fixed
 - Deserialize aggregation containing a parent aggregation ([#706](https://github.com/opensearch-project/opensearch-java/pull/706))
+- Error deserialization w/ AwsSdk2Transport ([#1068](https://github.com/opensearch-project/opensearch-java/pull/1068))
```
dblock, line 48: Let's try to be more specific? `Deserialize error response when using AwsSdk2Transport`?

dblock (commented):
@Xtansia @reta mind taking a quick look at this one?

CHANGELOG.md
workmanw: 👍

java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
```diff
@@ -617,6 +633,22 @@ private static <T> Optional<T> or(Optional<T> opt, Supplier<? extends Optional<?
         }
     }
 
+    private static ByteArrayInputStream toByteArrayInputStream(InputStream is) throws IOException {
```
workmanw, line 636: I'm not crazy about this solution, but it's effectively what `ApacheHttpClient5Transport` is doing by using `BufferedHttpEntity`.

java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
```diff
@@ -617,6 +633,22 @@ private static <T> Optional<T> or(Optional<T> opt, Supplier<? extends Optional<?
...
+        }
+
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        byte[] buffer = new byte[1024];
+        int len;
+        while ((len = is.read(buffer)) > -1) {
+            baos.write(buffer, 0, len);
+        }
+        baos.flush();
+        return new ByteArrayInputStream(baos.toByteArray());
```
Xtansia, line 649: We could re-use aws-sdk-java's `IoUtils` to shorten this: https://github.com/aws/aws-sdk-java-v2/blob/master/utils/src/main/java/software/amazon/awssdk/utils/IoUtils.java#L39-L52
```suggestion
        return new ByteArrayInputStream(IoUtils.toByteArray(is));
```

Xtansia (dismissed):
The changes all look reasonable to me and in-line with the HttpClient5 transport's behaviour

java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
workmanw: Oh awesome. This definitely removes verbosity 💯 .

java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
```diff
@@ -10,10 +10,7 @@
 
 import jakarta.json.JsonObject;
 import jakarta.json.stream.JsonParser;
-import java.io.ByteArrayInputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.UnsupportedEncodingException;
+import java.io.*;
```
reta, line 13: Could you please revert to non-`*` import? (`spotless` should have spotted that)

reta (approved):

java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
workmanw: Done. I think `spotless` actually made this change. My IDE is set to never do wildcards. Either way, it's fixed.

Xtansia (approved):

PR #1104, opened 2024-07-23
add checks to existing unit tests

### LINKED ISSUE #1103: [BUG] main branch and 2.x branch behave differently
### What is the bug?
See unit test updates. They work on main but not 2.x.

Run EndpointTest on main (passes) and then 2.x (fails)

Should behave the same if all code changes were moved to 2.x

Eclipse on Linux

No but can provide if critical

Updated unit tests on the way. Related to #1068 and #999 and #1012 and possibly others.

al-niessner:
### Description
Shows that 'a:b:c::2.0' works on main but not 2.x branches.

### Issues Resolved
#1103 

### DIFF
--- a/java-client/src/test/java/org/opensearch/client/opensearch/model/EndpointTest.java
+++ b/java-client/src/test/java/org/opensearch/client/opensearch/model/EndpointTest.java
@@ -35,9 +35,15 @@ package org.opensearch.client.opensearch.model;
 import org.junit.Assert;
 import org.junit.Test;
 import org.opensearch.client.opensearch._types.ExpandWildcard;
+import org.opensearch.client.opensearch.core.GetRequest;
 import org.opensearch.client.opensearch.indices.RefreshRequest;
 
 public class EndpointTest extends Assert {
+    @Test
+    public void testIdEncoding() {
+        GetRequest req = new GetRequest.Builder().index("db").id("a:b:c::2.0").build();
+        assertEquals("/db/_doc/a%3Ab%3Ac%3A%3A2.0", GetRequest._ENDPOINT.requestUrl(req));
+    }
 
     @Test
     public void testArrayPathParameter() {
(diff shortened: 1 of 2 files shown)


al-niessner:
What would you like added to change log since this is just unit test changes and not overly complicated changes.

dblock:
We can skip the changelog, but not sure what's up with the integ test failure.

```
Execution failed for task ':java-client:compileTestJava'.
> Could not resolve all files for configuration ':java-client:testCompileClasspath'.
   > Could not find any matches for software.amazon.awssdk:aws-crt-client:[2.15,3.0) as no versions of software.amazon.awssdk:aws-crt-client are available.
```

al-niessner:
> We can skip the changelog, but not sure what's up with the integ test failure.
> 
> ```
> Execution failed for task ':java-client:compileTestJava'.
> > Could not resolve all files for configuration ':java-client:testCompileClasspath'.
>    > Could not find any matches for software.amazon.awssdk:aws-crt-client:[2.15,3.0) as no versions of software.amazon.awssdk:aws-crt-client are available.
> ```
> 
> Is it caused by this import which attempts to bring the kitchen sink from OpenSearch?
> 
> ```
> import org.opensearch.client.opensearch.core.GetRequest;
> ```

Is there a different GetRequest I should be importing? I have all the amazon cruft in my environment so does not matter to me. I think the point is that the GetRequest is not the same on main as 2.x. It is the encoding of the ID that we ultimately care about (my project) and the RefreshRequest is insufficient to test the ID encoding. Probably similar failure with DeleteByQueryRequest too but maybe fixing GetRequest (or its root cause) will fix the DBQR as well.

al-niessner:
I take it back. My gradle build in eclipse is showing 3 errors in the integrations tests but it looks like eclipse is java 8 while these tests are in src/test/java11 so I am going to say that it local to me. Still the changes I made should not make any difference to your build beyond importing from core somehow being evil.

Xtansia:
The failures were completely unrelated to the changes in this PR, just got unlucky and they ran at the exact time a release of AWS SDK was happening, so got some of the newer version libraries while their dependencies hadn't been pushed to maven yet.

Xtansia:
Only failures now are DCO and spotless check.

al-niessner:
Sorry I see two MAC build fails and DCO. What is spotless?

dblock:
> Sorry I see two MAC build fails and DCO. What is spotless?

Check out https://github.com/opensearch-project/OpenSearch/blob/main/DEVELOPER_GUIDE.md#java-language-formatting-guidelines, we use the same linter here.

al-niessner:
Oh, I see. I thought MacOS-latest was some attempt to build the latest main branch on a MacOS. Anyway, followed directions in the output and used gradlew to fix it.

dblock (approved):

PR #1107, opened 2024-07-24
possible fix for #1106

al-niessner:
### Description
Change order of libraries found for URL encoding

### Issues Resolved
Might solve #1106 but it leaves a bad bug in the code for those that do not use httpClient5.


al-niessner:
Nevermind

PR #1117, opened 2024-08-01
Fixed deserialization of SearchRequest when `_source` is an array

### LINKED ISSUE #1087: [BUG] Unable to deserialize SearchRequest with _source field as an array
### What is the bug?
The error `org.opensearch.client.json.UnexpectedJsonEventException: Unexpected JSON event 'START_ARRAY' instead of '[START_OBJECT, KEY_NAME, VALUE_STRING, VALUE_TRUE, VALUE_FALSE]` is raised when trying to deserialize following json payload to `SearchRequest`.
```
{
  "query": {
    "match_all": {}
  },
  "_source": ["_id"],
  "size": 1
}
```
Context: We are implementing a kind of proxy service which accepts any OpenSearch requests and overwrite some query logic before passing them to the OpenSearch cluster. Previously, we use `opensearch-rest-high-level-client` with `XContentParser` to parse incoming requests but now we would like to migrate to `opensearch-java` library.

Note that if `_source` is removed or change to string field, it can parse successfully but we cannot do this as a workaround as we expect to accept any form of valid requests from clients.

Execute following code and then will get `org.opensearch.client.json.UnexpectedJsonEventException`
```java
import com.fasterxml.jackson.core.JsonFactory;
import org.opensearch.client.json.jackson.JacksonJsonpMapper;
import org.opensearch.client.json.jackson.JacksonJsonpParser;
import org.opensearch.client.opensearch.core.SearchRequest;

import java.io.IOException;

public class Test {
    public static void main(String[] args) throws IOException {
        String jsonPayload = "{\n" +
                "  \"query\": {\n" +
                "    \"match_all\": {}\n" +
                "  },\n" +
                "  \"_source\": [\"_id\"],\n" +
                "  \"size\": 1\n" +
                "}\n";

        SearchRequest searchRequest = SearchRequest._DESERIALIZER.deserialize(new JacksonJsonpParser(new JsonFactory().createParser(jsonPayload)), new JacksonJsonpMapper());
    }
}
```

We expect `SearchRequest._DESERIALIZER` can parse the json payload successfully.

- library = opensearch-java 2.11.1
- SDK = Amazon Correto 11
- OS = MacOS Somona 14.5

Xtansia:
### Description
Fixed deserialization of SearchRequest when `_source` is an array

### Issues Resolved
Fixes #1087 

### DIFF
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -41,8 +41,8 @@ jobs:
       - name: Run Docker
         run: |
           echo "PASSWORD=admin" >> $GITHUB_ENV
-          docker-compose --project-directory .ci/opensearch build --build-arg OPENSEARCH_VERSION=${{ matrix.entry.opensearch_version }}
-          docker-compose --project-directory .ci/opensearch up -d
+          docker compose --project-directory .ci/opensearch build --build-arg OPENSEARCH_VERSION=${{ matrix.entry.opensearch_version }}
+          docker compose --project-directory .ci/opensearch up -d
           sleep 60
 
       - name: Sets password (new versions)
@@ -64,4 +64,4 @@ jobs:
 
       - name: Stop Docker
         run: |
-          docker-compose --project-directory .ci/opensearch down
+          docker compose --project-directory .ci/opensearch down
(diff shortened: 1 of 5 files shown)


dblock (approved):

dblock:
Looks like backport didn't run 🤷

PR #1124, opened 2024-08-02
Regenerate dangling_indices namespace

Xtansia:
### Description
Regenerate dangling_indices namespace

### DIFF
--- a/java-client/src/generated/java/org/opensearch/client/opensearch/OpenSearchAsyncClient.java
+++ b/java-client/src/generated/java/org/opensearch/client/opensearch/OpenSearchAsyncClient.java
@@ -43,6 +43,7 @@ import javax.annotation.Nullable;
 import org.opensearch.client.opensearch._types.OpenSearchException;
 import org.opensearch.client.opensearch.core.InfoRequest;
 import org.opensearch.client.opensearch.core.InfoResponse;
+import org.opensearch.client.opensearch.dangling_indices.OpenSearchDanglingIndicesAsyncClient;
 import org.opensearch.client.transport.OpenSearchTransport;
 import org.opensearch.client.transport.TransportOptions;
 
@@ -62,6 +63,12 @@ public class OpenSearchAsyncClient extends OpenSearchAsyncClientBase<OpenSearchA
     @Override
     public OpenSearchAsyncClient withTransportOptions(@Nullable TransportOptions transportOptions) {
         return new OpenSearchAsyncClient(this.transport, transportOptions);
+    }
+
+    // ----- Child clients
+
+    public OpenSearchDanglingIndicesAsyncClient danglingIndices() {
+        return new OpenSearchDanglingIndicesAsyncClient(this.transport, this.transportOptions);
     }
 
     // ----- Endpoint: info
--- a/java-client/src/generated/java/org/opensearch/client/opensearch/OpenSearchAsyncClient.java
+++ b/java-client/src/generated/java/org/opensearch/client/opensearch/OpenSearchAsyncClient.java
@@ -46,6 +46,9 @@ import org.opensearch.client.opensearch.core.InfoResponse;
 import org.opensearch.client.transport.OpenSearchTransport;
 import org.opensearch.client.transport.TransportOptions;
 
+/**
+ * Client for the namespace.
+ */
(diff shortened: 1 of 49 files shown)


java-client/src/generated/java/org/opensearch/client/opensearch/OpenSearchAsyncClient.java
```diff
@@ -61,6 +65,12 @@ public OpenSearchAsyncClient withTransportOptions(@Nullable TransportOptions tra
         return new OpenSearchAsyncClient(this.transport, transportOptions);
     }
 
+    // ----- Child clients
+
+    public OpenSearchDanglingIndicesAsyncClient danglingIndices() {
```
reta, line 70: Why it was moved from `OpenSearchXxxClientBase` class?

java-client/src/generated/java/org/opensearch/client/opensearch/dangling_indices/list_dangling_indices/DanglingIndex.java
```diff
@@ -138,15 +159,29 @@ protected void serializeInternal(JsonGenerator generator, JsonpMapper mapper) {
...
         private String indexName;
-
         private String indexUuid;
+        private List<String> nodeIds;
 
-        private String creationDateMillis;
+        /**
+         * API name: {@code creation_date}
+         */
+        public final Builder creationDate(@Nullable String value) {
```
dblock, line 173: This is not in the spec, is it? Should it take a `LocalDateTime` (or have an override like that)?

java-client/src/generated/java/org/opensearch/client/opensearch/_types/ErrorCause.java
```diff
@@ -51,139 +56,133 @@
...
 @JsonpDeserializable
+@Generated("org.opensearch.client.codegen.CodeGenerator")
 public class ErrorCause implements PlainJsonSerializable {
-    private final Map<String, JsonData> metadata;
 
-    private final String type;
+    @Nullable
+    private final ErrorCause causedBy;
 
+    @Nullable
```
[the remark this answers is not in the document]
Xtansia: [`reason` comes from `Throwable::getMessage`](https://github.com/opensearch-project/OpenSearch/blob/main/libs/core/src/main/java/org/opensearch/OpenSearchException.java#L384-L400) which [may be null](https://docs.oracle.com/javase/8/docs/api/java/lang/Throwable.html#getMessage--), and there's no other guarantees of non-nullability in `OpenSearchException`.

java-client/src/generated/java/org/opensearch/client/opensearch/dangling_indices/list_dangling_indices/DanglingIndex.java
Xtansia: I'm adding it to the spec in https://github.com/opensearch-project/opensearch-api-specification/pull/462

Though now I think it probably shouldn't actually get added to the client as it's one of the "human-readable" fields that's conditionally returned based on `?human`, but we don't expose the `human` query parameter in the client

java-client/src/generated/java/org/opensearch/client/opensearch/OpenSearchAsyncClient.java
Xtansia: Because it's now generated it's moved to the generated version of the client class

reta: Hm ... the `OpenSearchXxxClientBase` classes are public so we cannot just remove methods from them I believe for 2.x at least. We could keep the  `OpenSearchXxxClientBase` unchanged, right? (since it will be just an override in this case).

Xtansia: The `*Base` classes only exist in main, and haven't been backported to `2.x` so we haven't actually locked into its public contract yet

reta: Ah got it, thanks @Xtansia !

reta (approved):

PR #1131, opened 2024-08-07
Removed null constraints on NodeInfo fields that are returned null from Amazon OpenSearch Service

### LINKED ISSUE #1129: [BUG] NodeInfoRequest fails when connecting to Amazon OpenSearch Service
### What is the bug?
When sending a NodeInfoRequest to Amazon OpenSearch Service the request fails with the following error:

`Missing required property 'NodeInfo.host'`

This issue appears to be caused by  Amazon OpenSearch Service obfuscating/removing some fields from the REST API responses that the client code expects to be not null.

Beside NodeInfo.host the following fields are also missing from the response and are causing a similar error:

- NodeInfo.IP
- NodeInfo.TransportAddress

1. Connect the client to Amazon OpenSearch Service with basic auth and only fine-grained access control enabled (IAM is disabled).
2.  Send a NodeInfoRequest:
```
   NodesInfoRequest nodesInfoRequest = new NodesInfoRequest.Builder().metric(List.of("plugins")).build();
   NodesInfoResponse nodesInfoResponse = openSearchNodesClient.info(nodesInfoRequest);
```

I can raise a PR with a fix which consists in removing _requireNonNull_ from these specific fields.

dancristiancecoi:
### Description
This change removes null constraints on NodeInfo.Host, NodeInfo.Ip, NodeInfo.TransportAddress that are returned null from Amazon OpenSearch Service when performing a NodeInfoRequest

### Issues Resolved
#1129 


dancristiancecoi:
Have some questions about the Changelog. I want this backported to 2.x if approved, do I add an entry to the 2.x section as well or just to the 3.x?

dblock:
> Have some questions about the Changelog. I want this backported to 2.x if approved, do I add an entry to the 2.x section as well or just to the 3.x one?

2.x I believe

Xtansia:
> Have some questions about the Changelog. I want this backported to 2.x if approved, do I add an entry to the 2.x section as well or just to the 3.x one?

 Entries should only go under one section or the other not both, so for your case It would go under only the `Unreleased 2.x` section

java-client/src/main/java/org/opensearch/client/opensearch/nodes/info/NodeInfo.java
```diff
@@ -60,11 +60,13 @@ public class NodeInfo implements PlainJsonSerializable {
 
     private final String buildType;
 
+    @Nullable
```
Xtansia, line 63: The relevant getters below should also get `@Nullable` annotations

java-client/src/main/java/org/opensearch/client/opensearch/nodes/info/NodeInfo.java
```diff
@@ -828,7 +834,7 @@ public final Builder transport(Function<NodeInfoTransport.Builder, ObjectBuilder
          * <p>
          * API name: {@code transport_address}
          */
-        public final Builder transportAddress(String value) {
+        public final Builder transportAddress(@Nullable  String value) {
```
Xtansia, line 837: Please fix the double space here, or just run `./gradlew spotlessApply`
```suggestion
        public final Builder transportAddress(@Nullable String value) {
```

Xtansia (changes requested):

dancristiancecoi:
Thanks guys!

dancristiancecoi:
Ugh, I messed up the rebase trying to solve the conflicts. I will close this PR and open a new one as it will be quicker this way. Sorry :D

PR #1147, opened 2024-08-19
Fixes #1116 - Add cancel_after_time_interval

### LINKED ISSUE #1116: [FEATURE] Add cancel_after_time_interval to SearchRequest and MsearchRequest
The `cancel_after_time_interval` parameter does not exist for search and msearch requests in the Java client.

Add `cancel_after_time_interval` to `SearchRequest` and `MsearchRequest` in the Java client.

I have not found a workaround.

Being able to use this parameter is especially useful when using an AWS hosted OpenSearch cluster. Because AWS does not provide a way to set the default `cancel_after_time_interval` setting for a cluster, there doesn't seem to be any other way to ensure that a request which can't be handled in a reasonable time will be cancelled. We have occasionally encountered outages caused by extremely complicated, long-running search requests bogging down the cluster.

This parameter was missing from the OpenSearch Open API spec for years, and was finally added in June 2024: 
https://github.com/opensearch-project/opensearch-api-specification/issues/273

I see there is work underway to generate the Java client code from the Open API specification, which hopefully will eventually provide a solution. But it would be very useful to be able to set this parameter in the interim. 
https://github.com/opensearch-project/opensearch-java/pull/366 is the recently merged PR to which I am referring.

urinud:
### Description
Add `cancel_after_time_interval` field to `SearchRequest` and `MsearchRequest` requests in the Java client.

### Issues Resolved
- Fixes #1116 

### DIFF
--- a/java-client/src/main/java/org/opensearch/client/opensearch/core/MsearchRequest.java
+++ b/java-client/src/main/java/org/opensearch/client/opensearch/core/MsearchRequest.java
@@ -353,6 +353,16 @@ public class MsearchRequest extends RequestBase implements NdJsonpSerializable, 
         }
 
         /**
+         * The time after which the search request will be canceled.
+         * Request-level parameter takes precedence over cancel_after_time_interval cluster setting.
+         * <p>
+         * API name: {@code cancel_after_time_interval}
+         */
+        public final Builder cancelAfterTimeInterval(Function<Time.Builder, ObjectBuilder<Time>> fn) {
+            return this.cancelAfterTimeInterval(fn.apply(new Time.Builder()).build());
+        }
+
+        /**
          * If true, network roundtrips between the coordinating node and remote clusters
          * are minimized for cross-cluster search requests.
          * <p>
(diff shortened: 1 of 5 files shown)


urinud:
Hello @dblock what do you think of this PR ?

dblock (approved):
Thanks!

1. @Xtansia is working to generate this code from spec in https://github.com/opensearch-project/opensearch-api-specification, @urinud can you help making sure this field exists and is used in tests in that project?
2. Looks good! 

Leaving this open for @Xtansia to make sure this is in line with what the spec will do.

urinud:
@dblock & @Xtansia , I see the field defined in:
https://github.com/urinud/opensearch-api-specification/blob/7dee041660986c5eb4bc85c2d19af006730746e9/spec/namespaces/_core.yaml#L5376
I also see a test created for this field too. LGTM.

java-client/src/main/java/org/opensearch/client/opensearch/core/MsearchRequest.java
```diff
@@ -321,6 +341,17 @@ public final Builder allowNoIndices(@Nullable Boolean value) {
             return this;
         }
 
+        /**
+         * The time after which the search request will be canceled.
+         * Request-level parameter takes precedence over cancel_after_time_interval cluster setting.
+         * <p>
+         * API name: {@code cancel_after_time_interval}
+         */
+        public final Builder cancelAfterTimeInterval(@Nullable Time value) {
```
Xtansia, line 350: This should also have the builder `Function` overload like you've added to `SearchRequest`

Xtansia (changes requested):
Changes look good, just one small nit @urinud

java-client/src/main/java/org/opensearch/client/opensearch/core/MsearchRequest.java
urinud: @Xtansia , added the function overload method.
I did not added initially becase no other method has function overloads.

Xtansia (approved):

PR #1148, opened 2024-08-20
support for withJson-618

### LINKED ISSUE #618: [FEATURE] Provide an easy way to deserialize classes annotated as @JsonpDeserializable
We are migrating from Elasticsearch to Opensearch and we have some JSON files that are used to create index templates. That is how we used to do it:
````
    public void putIndexTemplate(final String pTemplateName, final InputStream pTemplateJson) throws IOException {
        PutIndexTemplateRequest putIndexTemplateRequest = new PutIndexTemplateRequest.Builder()
            .name(pTemplateName)
            .withJson(pTemplateJson)
            .build();
       PutIndexTemplateResponse putIndexTemplateResponse = client.indices().putIndexTemplate(putIndexTemplateRequest);
        log.info(putIndexTemplateResponse.toString());
    }
````
But that `.withJson` method is no longer available.

We could have the `.withJson` method added to the Builder, since the request class already has a `JsonpDeserializer<PutIndexTemplateRequest> _DESERIALIZER`. Or another alternative that would allow us to keep using the builder.

First I have tried to use the built-in deserializer, like this:
````
public void putIndexTemplate(final String pTemplateName, final InputStream pTemplateJson) throws IOException {
        JsonpMapper mapper = client._transport().jsonpMapper();
        JsonParser parser = mapper.jsonProvider().createParser(pTemplateJson);

        PutIndexTemplateRequest request = PutIndexTemplateRequest._DESERIALIZER.deserialize(parser, mapper);
        // request is immutable, I can't set request.setName(pTemplateName)
        PutIndexTemplateResponse response = client.indices().putIndexTemplate(request);

        log.info(response.toString());
}
````

Then I had to build my own Deserializer to return a Builder instead:
````
public final class PutIndexTemplateRequestBuilderDeserializer {
    private static final ObjectDeserializer<PutIndexTemplateRequest.Builder> od = new ObjectDeserializer<>(PutIndexTemplateRequest.Builder::new);

    static {
        od.add(PutIndexTemplateRequest.Builder::meta, stringMapDeserializer(JsonData._DESERIALIZER), "_meta");
        od.add(PutIndexTemplateRequest.Builder::composedOf, arrayDeserializer(stringDeserializer()), "composed_of");
        od.add(PutIndexTemplateRequest.Builder::dataStream, DataStream._DESERIALIZER, "data_stream");
        od.add(PutIndexTemplateRequest.Builder::indexPatterns, arrayDeserializer(stringDeserializer()), "index_patterns");
        od.add(PutIndexTemplateRequest.Builder::priority, JsonpDeserializer.integerDeserializer(), "priority");
        od.add(PutIndexTemplateRequest.Builder::template, IndexTemplateMapping._DESERIALIZER, "template");
        od.add(PutIndexTemplateRequest.Builder::version, JsonpDeserializer.longDeserializer(), "version");
    }

    public static PutIndexTemplateRequest.Builder deserialize(JsonpMapper mapper, JsonParser parser) {
        return od.deserialize(parser, mapper);
    }
}
````

And finally use it in my class:
````
    public void putIndexTemplate(final String pTemplateName, final InputStream pTemplateJson) throws IOException {
        JsonpMapper mapper = client._transport().jsonpMapper();
        JsonParser parser = mapper.jsonProvider().createParser(pTemplateJson);

        PutIndexTemplateRequest request = PutIndexTemplateRequestBuilderDeserializer
            .deserialize(mapper, parser)
            .name(pTemplateName)
            .build();
        PutIndexTemplateResponse response = client.indices().putIndexTemplate(request);

        log.info(response.toString());
    }
````

One last thing, there is no documentation whatsoever describing how to do this, and I think it is not a very unusual scenario.

Thanks for considering it.

### PR DESCRIPTION
**Pull request by Jai2305**
Description
Added PlainDeserializable interface with methods for streamlining deserialization process .

Issues Resolved
Closes #618 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,7 +53,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 - Added `queryImage` (query_image) field to `NeuralQuery`, following definition in ([Neural Query](https://opensearch.org/docs/latest/query-dsl/specialized/neural/)) ([#1137](https://github.com/opensearch-project/opensearch-java/pull/1138))
 - Added `cancelAfterTimeInterval` to `SearchRequest` and `MsearchRequest` ([#1147](https://github.com/opensearch-project/opensearch-java/pull/1147))
 - Added the `ml` namespace operations ([#1158](https://github.com/opensearch-project/opensearch-java/pull/1158))
-- Added `PlainDeserializable` with `withJson` method for streamlining deserialization ([#1148](https://github.com/opensearch-project/opensearch-java/pull/1148))
+- Added `IndexTemplateMapping.Builder#withJson`, `SourceField.Builder#withJson` and `IndexSettings.Builder#withJson` for streamlining deserialization ([#1148](https://github.com/opensearch-project/opensearch-java/pull/1148))
 
 ### Dependencies
 - Bumps `commons-logging:commons-logging` from 1.3.3 to 1.3.4
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 - Added `queryImage` (query_image) field to `NeuralQuery`, following definition in ([Neural Query](https://opensearch.org/docs/latest/query-dsl/specialized/neural/)) ([#1137](https://github.com/opensearch-project/opensearch-java/pull/1138))
 - Added `cancelAfterTimeInterval` to `SearchRequest` and `MsearchRequest` ([#1147](https://github.com/opensearch-project/opensearch-java/pull/1147))
 - Added the `ml` namespace operations ([#1158](https://github.com/opensearch-project/opensearch-java/pull/1158))
+- Added `PlainDeserializable` with `withJson` method for streamlining deserialization ([#1148](https://github.com/opensearch-project/opensearch-java/pull/1148))
 
 ### Dependencies
 - Bumps `commons-logging:commons-logging` from 1.3.3 to 1.3.4
--- a/java-client/src/main/java/org/opensearch/client/json/PlainDeserializable.java
+++ b/java-client/src/main/java/org/opensearch/client/json/PlainDeserializable.java
@@ -38,8 +38,9 @@ public interface PlainDeserializable<B> {
       * **/
     default B withJson(InputStream inputStream) {
         JsonpMapper defaultMapper = JsonpUtils.DEFAULT_JSONP_MAPPER;
-        JsonParser parser = defaultMapper.jsonProvider().createParser(inputStream);
-        return withJson(parser, defaultMapper);
+        try (JsonParser parser = defaultMapper.jsonProvider().createParser(inputStream)) {
+            return withJson(parser, defaultMapper);
+        }
     }
 
     /** Updates object with newly provided JSON properties
@@ -48,8 +49,9 @@ public interface PlainDeserializable<B> {
       * **/
     default B withJson(Reader reader) {
         JsonpMapper defaultMapper = JsonpUtils.DEFAULT_JSONP_MAPPER;
-        JsonParser parser = defaultMapper.jsonProvider().createParser(reader);
-        return withJson(parser, defaultMapper);
+        try (JsonParser parser = defaultMapper.jsonProvider().createParser(reader)) {
+            return withJson(parser, defaultMapper);
+        }
     }
 
 }
--- a/java-client/src/main/java/org/opensearch/client/json/PlainDeserializable.java
+++ b/java-client/src/main/java/org/opensearch/client/json/PlainDeserializable.java
@@ -1,0 +1,55 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.client.json;
+
+import jakarta.json.stream.JsonParser;
+import java.io.InputStream;
+import java.io.Reader;
+
+/** Base interface to set JSON properties **/
+
+public interface PlainDeserializable<B> {
+
+    B self();
+
+    /** Updates object with newly provided JSON properties
+     @param parser the JsonParser parser
+     @param mapper the JsonpMapper mapper used to deserialize values
+     @return this object
+     **/
+
+    default B withJson(JsonParser parser, JsonpMapper mapper) {
+        JsonpDeserializer<?> deserializer = JsonpMapperBase.findDeserializer(this.getClass().getEnclosingClass());
+        @SuppressWarnings("unchecked")
+        ObjectDeserializer<B> objectDeserializer = (ObjectDeserializer<B>) DelegatingDeserializer.unwrap(deserializer);
+        assert objectDeserializer != null;
+        return objectDeserializer.deserialize(self(), parser, mapper, parser.next());
+    }
+
+    /** Updates object with newly provided JSON properties
+     @param inputStream the stream to read from
+     @return this object
+      * **/
+    default B withJson(InputStream inputStream) {
+        JsonpMapper defaultMapper = JsonpUtils.DEFAULT_JSONP_MAPPER;
+        JsonParser parser = defaultMapper.jsonProvider().createParser(inputStream);
+        return withJson(parser, defaultMapper);
+    }
+
+    /** Updates object with newly provided JSON properties
+     @param reader the stream to read from
+     @return this object
+      * **/
+    default B withJson(Reader reader) {
+        JsonpMapper defaultMapper = JsonpUtils.DEFAULT_JSONP_MAPPER;
+        JsonParser parser = defaultMapper.jsonProvider().createParser(reader);
+        return withJson(parser, defaultMapper);
+    }
+
+}
--- a/java-client/src/test/java/org/opensearch/client/opensearch/json/PlainDeserializableTest.java
+++ b/java-client/src/test/java/org/opensearch/client/opensearch/json/PlainDeserializableTest.java
@@ -1,0 +1,42 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.client.opensearch.json;
+
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.List;
+import org.junit.Test;
+import org.opensearch.client.opensearch.indices.PutIndexTemplateRequest;
+import org.opensearch.client.opensearch.indices.put_index_template.IndexTemplateMapping;
+
+public class PlainDeserializableTest {
+    @Test
+    public void testWithJsonPutIndexTemplateRequest() {
+
+        String stringTemplate =
+            "{\"mappings\":{\"properties\":{\"age\":{\"type\":\"integer\"}}},\"settings\":{\"number_of_shards\":\"2\",\"number_of_replicas\":\"1\"}}";
+        InputStream inputStream = new ByteArrayInputStream(stringTemplate.getBytes(StandardCharsets.UTF_8));
+
+        PutIndexTemplateRequest indexTemplateRequest = new PutIndexTemplateRequest.Builder().name("My index")
+            .indexPatterns("index_pattern1")
+            .template(new IndexTemplateMapping.Builder().withJson(inputStream).build())
+            .build();
+
+        String expectedName = "My index";
+        List<String> expectedIndexPatterns = Arrays.asList("index_pattern1");
+        String expectedNumberOfShards = "2";
+
+        assert expectedName.equals(indexTemplateRequest.name());
+        assert expectedIndexPatterns.equals(indexTemplateRequest.indexPatterns());
+        assert expectedNumberOfShards.equals(indexTemplateRequest.template().settings().numberOfShards());
+
+    }
+}
--- a/java-client/src/test/java/org/opensearch/client/opensearch/json/PlainDeserializableTest.java
+++ b/java-client/src/test/java/org/opensearch/client/opensearch/json/PlainDeserializableTest.java
@@ -8,6 +8,8 @@
 
 package org.opensearch.client.opensearch.json;
 
+import static org.junit.Assert.assertEquals;
+
 import java.io.ByteArrayInputStream;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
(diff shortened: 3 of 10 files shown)

### CODE REVIEW

**Comment by Jai2305** [comment]
Hi, I am converting this PR into a draft , I wanted to share the general approach , i.e for all the classes annotated with `@JsonpDeserializable ` , I want to add an interface which shall be implemented by respective Builder classes and thus can be chained with the Builder methods to create the instance 

Once you are aligned on this approach without any concerns , then changes for all the classes will be implemented in this PR

**Comment by dblock** [comment]
I like it! Iterate to green/add some documentation and maybe a sample?

**Comment by Jai2305** [comment]
Can you also please clarify  - 
* As we are coding to an interface , we have a choice whether a class's Builder implements it or not , should we first do it only for some classes and opt out from others , or are we going for every class which is annotated with `@JsonpDeserializable`.

**Comment by Jai2305** [comment]
@dblock any thoughts on above ?

**Comment by dblock** [comment]
> @dblock any thoughts on above ?

I think users want it everywhere - are there any arguments to limit it?

File: java-client/src/main/java/org/opensearch/client/opensearch/cluster/AllocationExplainRequest.java [hunk 1]
```diff
@@ -37,12 +37,7 @@
 import java.util.Map;
 import java.util.function.Function;
 import javax.annotation.Nullable;
-import org.opensearch.client.json.JsonpDeserializable;
-import org.opensearch.client.json.JsonpDeserializer;
-import org.opensearch.client.json.JsonpMapper;
-import org.opensearch.client.json.ObjectBuilderDeserializer;
-import org.opensearch.client.json.ObjectDeserializer;
-import org.opensearch.client.json.PlainJsonSerializable;
+import org.opensearch.client.json.*;
```
dblock, line 40: I am surprised this doesn't get trapped by spotless, let's expand these like in all other places (and check spotless configuration please).

dblock (commented):
This looks alright to me, but as you point out I think we want it at a lower level. Basically any builder that accepts a body should also accept `withJson`, shouldn't it?

java-client/src/main/java/org/opensearch/client/opensearch/cluster/AllocationExplainRequest.java
Jai2305: Yeah that would be my ide auto import :) , I will fix these.

Jai2305:
> > @dblock any thoughts on above ?
> 
> I think users want it everywhere - are there any arguments to limit it?

I was unsure about the universal application of the withJson feature. For eg. some response classes which are annotated with `@JsonpDeserializable` , in those cases the data is expected to be returned by search engine so I was not sure about the usecase of withJson in those type of classes , at first I thought of implementing the feature to only the request classes which are annotated with `@JsonpDeserializable` .
With that being said I do not have any argument against it, let's implement it to @JsonpDeserializable classes with builders.

dblock:
I think you have valid concerns, @Xtansia @reta WDYT?

Will we be able to have `withNdJson()` for `bulk`?

Jai2305:
* If we decide it to be appropriate for every Builder to have withJson then one solution that I have is to make new abstract object Builder class which extends the classes like `ObjectBuilders` and `AbstractBuilders` and implements the new `PlainDeserializable` interface  - these are basically the classes extended by every builder . By making an abstract class or different abstract classes we can control the access of  self() method -  this is the method which returns the Builder of a class.  
                     - cons - All files will have to implement the extend the new classes, will have to add new abstract classes and I think similar code will have to be repeated .. 
                      - pro - better control, extensibility 

* Or we can modify `ObjectBuilder` class itself 
                     - cons - a bit dangerous, no opt out option, 
                      - pro -changes will be applied to all the builders in the library at once.

dblock:
@Jai2305 some builders will need to serialize JSON but others ND-JSON (bulk), and some both that may be implemented differently (you can make nd-json from json). So I think (2) doesn't work, no?

Jai2305:
> @Jai2305 some builders will need to serialize JSON but others ND-JSON (bulk), and some both that may be implemented differently (you can make nd-json from json). So I think (2) doesn't work, no?

You are right. There can be a work around to add new Abstract classes for NdJson cases, but,  this is in general a pretty clunky solution, lets rule it out then.

reta:
> @Jai2305 some builders will need to serialize JSON but others ND-JSON (bulk), and some both that may be implemented differently (you can make nd-json from json). So I think (2) doesn't work, no?

@dblock @Jai2305 I have concerns regarding this approach:
 1. I believe we **should not** deserialize builder from JSON, it lets to very weird code constructs like:
   ```
        PutIndexTemplateRequest indexTemplateRequest = new PutIndexTemplateRequest.Builder().name("My index 2")
            .indexPatterns("index_pattern3")
            .withJson(inputStream)
            .name("name")
            .build();
   ``` 
   
   (what is the values of `name` and `indexPatterns` after that? not to mention calling `withJson` a few times)
 2. However, some type of requests that accepts pretty much freestyle payloads (like `source`, `template` fe) would benefit from this particular feature
 3. If we really want to support builder desearlization, than we would need to look for something like that (very rough idea, but essentially have a builder instance deserialized first and tailored later): `new PutIndexTemplateRequest.Builder(json).name("My index 2")`

reta:
> * While `withJson` may appear unconventional, but it merely acts as a mutator that populates fields within the builder, potentially overwriting previously assigned values and then returning the Builder again for potential tailoring (client's choice).

Thanks @Jai2305, so what is the behaviour for properties that have not been specified? Will be replaced by `null`? Will be they kept untouched?  To me, it looks not only unconventional but obscure - the users of the API have to guess what is going to happen. I think we should come up with conventional API that leaves no guesses.

Jai2305:
@reta  
* they will be untouched, `withJson` is behaving exactly as if you are making your object using individual builders, because internally thats what it is doing , it is taking whatever the value of current builder is and setting the values in that builder using whatever values you have provided in your Json request . 
* This way you are getting all the advantages you get when you use builder pattern to populate the fields. 
 Eg in  `PutIndexTemplateRequest` , you will get an error if you call build without setting the name since it is a required property so either you can use something like this -
``` Java 
//  setting name using name field on builder 
String stringTemplate =
            "{\"composed_of\":[\"component1\",\"component2\"],\"index_patterns\":[\"index_pattern1\"]}";
        InputStream inputStream = new ByteArrayInputStream(stringTemplate.getBytes(StandardCharsets.UTF_8));
        
        PutIndexTemplateRequest indexTemplateRequest = new PutIndexTemplateRequest.Builder().name("My index 2")...
            .withJson(inputStream)
            .build();
```
Or
``` Java 
//  setting name  using withJson
  String stringTemplate =
            "{\"name\":\"My index\",\"composed_of\":[\"component1\",\"component2\"],\"index_patterns\":[\"index_pattern1\"]}";
        InputStream inputStream = new ByteArrayInputStream(stringTemplate.getBytes(StandardCharsets.UTF_8));

        PutIndexTemplateRequest indexTemplateRequest = new PutIndexTemplateRequest.Builder()
            .withJson(inputStream)
            .build();
```

I think it is perfect that way no?

reta:
> * they will be untouched, `withJson` is behaving exactly as if you are making your object using individual builders, because internally thats what it is doing , it is taking whatever the value of current builder is and setting the values in that builder using whatever values you have provided in your Json request .

Thanks @Jai2305, but than how to set specific value to `null` using JSON? The `Builder` pattern with per field access is very easily to reason about, the suggested `withJson` is not (in my opinion). Semantically, `withJson` is equivalent to creating a new instance of the `Builder`. And I still don't see any compelling use case for that to be fair (yes, I understand that we replicate what Elasticsearch Java does but it does not mean we should). 

@dblock do you have any thoughts on this feature in general?

dblock:
> @dblock do you have any thoughts on this feature in general?

I was 100% in agreement with @Jai2305's comment in https://github.com/opensearch-project/opensearch-java/pull/1148#issuecomment-2312748170. A builder's `.build()` ultimately is a JSON representation of the query, so `withJson` merges whatever you pass in with whatever is already there. I don't think it's confusing.

The other option is to only have `.withJson` for free-formed parameters (e.g. an index mapping). It's certainly safer. Is that what you're arguing for @reta?

> but than how to set specific value to null using JSON? 
By specifying `null` in the JSON `"{ key: null }"` which is different from absence of "key".

> Semantically, withJson is equivalent to creating a new instance of the Builder. 
Right, and the result is the modified builder with whatever you set via `withJson`.

reta:
> The other option is to only have `.withJson` for free-formed parameters (e.g. an index mapping). I hope we don't have APIs that take 2 free formed JSONs :) It's certainly safer. Is that what you're advocating we do @reta? I like it too.

That's right

dblock:
> Thanks @dblock
> 
> > The other option is to only have `.withJson` for free-formed parameters (e.g. an index mapping). I hope we don't have APIs that take 2 free formed JSONs :) It's certainly safer. Is that what you're advocating we do @reta? I like it too.
> 
> That's right

@Jai2305 WDYT? Thanks so much for hanging in here with us!

Jai2305:
So basically we only implement `withJson`  for these type of properties which are free formed. I do agree that it's certainly more safer that way. Do we have a strict list of such parameters ( or any pattern to determine if  they are appropriate or can be grouped in as free formed) , do you think that there is possibility for us not anticipating a property or deem it necessary right now to implement withJson feature and in future we will have to open for changes ?

dblock:
> So basically we only implement `withJson` for these type of properties which are free formed. I do agree that it's certainly more safer that way. Do we have a strict list of such parameters ( or any pattern to determine if they are appropriate or can be grouped in as free formed) , do you think that there is possibility for us not anticipating a property or deem it necessary right now to implement withJson feature and in future we will have to be open for changes ?

I don't think we have a list but adding is never a problem, it's removing that breaks backwards compat. So I'd add an interface, implement it for the couple of cases we know and then find the missing ones incrementally as people need them. Of course you can comb through https://github.com/opensearch-project/opensearch-api-specification but ... it's a lot of APIs :)

Jai2305:
I have worked on implementing PlainDeserializable for certain free-form fields within our application. To ensure expected behaviors, I have also added test cases.

I've identified these fields as potential candidates based on initial analysis, but I'm also reviewing various request-response cycles to identify additional fields that might benefit from this implementation.

I'd appreciate any insights or suggestions you may have regarding other fields that could be suitable for the withJson feature.
Or we can go with these fields initially and open an issue so that people can add their opinions and requests for other fields

CHANGELOG.md
```diff
@@ -53,6 +53,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 - Added `queryImage` (query_image) field to `NeuralQuery`, following definition in ([Neural Query](https://opensearch.org/docs/latest/query-dsl/specialized/neural/)) ([#1137](https://github.com/opensearch-project/opensearch-java/pull/1138))
 - Added `cancelAfterTimeInterval` to `SearchRequest` and `MsearchRequest` ([#1147](https://github.com/opensearch-project/opensearch-java/pull/1147))
 - Added the `ml` namespace operations ([#1158](https://github.com/opensearch-project/opensearch-java/pull/1158))
+- Added `PlainDeserializable` with `withJson` method for streamlining deserialization ([#1148](https://github.com/opensearch-project/opensearch-java/pull/1148))
```
dblock, line 56: I think this should spell what the end-user result is:

```suggestion
- Added `IndexTemplateMapping.Builder#withJson`, `SourceField.Builder#withJson` and `IndexSettings.Builder#withJson` for streamlining deserialization ([#1148](https://github.com/opensearch-project/opensearch-java/pull/1148))
```

dblock (dismissed):
I'm good with this, @reta?

Let's add some working examples in https://github.com/opensearch-project/opensearch-java/tree/main/samples please?

java-client/src/test/java/org/opensearch/client/opensearch/json/PlainDeserializableTest.java
```diff
@@ -0,0 +1,42 @@
...
+        PutIndexTemplateRequest indexTemplateRequest = new PutIndexTemplateRequest.Builder().name("My index")
+            .indexPatterns("index_pattern1")
+            .template(new IndexTemplateMapping.Builder().withJson(inputStream).build())
+            .build();
+
+        String expectedName = "My index";
+        List<String> expectedIndexPatterns = Arrays.asList("index_pattern1");
+        String expectedNumberOfShards = "2";
+
+        assert expectedName.equals(indexTemplateRequest.name());
```
reta, line 37: Please use JUnit assertions:

```suggestion
        assertThat(indexTemplateRequest.name(), equalsTo(expectedName));
```

java-client/src/main/java/org/opensearch/client/json/PlainDeserializable.java
```diff
@@ -0,0 +1,55 @@
...
+        return objectDeserializer.deserialize(self(), parser, mapper, parser.next());
+    }
+
+    /** Updates object with newly provided JSON properties
+     @param inputStream the stream to read from
+     @return this object
+      * **/
+    default B withJson(InputStream inputStream) {
+        JsonpMapper defaultMapper = JsonpUtils.DEFAULT_JSONP_MAPPER;
+        JsonParser parser = defaultMapper.jsonProvider().createParser(inputStream);
```
reta, line 41:
```suggestion
        try (JsonParser parser = defaultMapper.jsonProvider().createParser(inputStream)) {
        	return withJson(parser, defaultMapper);
        }
```

Jai2305: Addressed thanks .

java-client/src/main/java/org/opensearch/client/json/PlainDeserializable.java
```diff
@@ -0,0 +1,55 @@
...
+        return withJson(parser, defaultMapper);
+    }
+
+    /** Updates object with newly provided JSON properties
+     @param reader the stream to read from
+     @return this object
+      * **/
+    default B withJson(Reader reader) {
+        JsonpMapper defaultMapper = JsonpUtils.DEFAULT_JSONP_MAPPER;
+        JsonParser parser = defaultMapper.jsonProvider().createParser(reader);
```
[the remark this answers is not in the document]
Jai2305: Addressed thanks.

Jai2305:
> Let's add some working examples in https://github.com/opensearch-project/opensearch-java/tree/main/samples please?

I have also added a working sample and an explanation in docs directory for user's reference.

java-client/src/main/java/org/opensearch/client/json/JsonpUtils.java
```diff
@@ -68,7 +68,7 @@ static JsonProvider provider() {
         return JsonProvider.provider();
     }
 
-    static final JsonpMapper DEFAULT_JSONP_MAPPER = new JsonpMapperBase() {
+    public static final JsonpMapper DEFAULT_JSONP_MAPPER = new JsonpMapperBase() {
```
reta, line 71: Sorry, I missed that, please keep it package private:

```suggestion
    static final JsonpMapper DEFAULT_JSONP_MAPPER = new JsonpMapperBase() {
    
```

java-client/src/test/java/org/opensearch/client/opensearch/json/PlainJsonSerializableTest.java
```diff
@@ -52,7 +52,7 @@ public void testIndexResponse() {
 
     // Test SearchRequest which implements PlainJsonSerializable
     @Test
-    public void testSearchResponse() {
+    public void testSearchRequest() {
```
reta, line 55: :+1:

reta (approved):

reta:
@dblock LGTM, please go ahead if you concerns are resolved, thank a  lot @Jai2305 !

dblock (approved):
Great job, thanks!

dblock:
@Jai2305 will you please help backport this to 2.x manually? looks like the automatic workflow failed

PR #1152, opened 2024-08-22
Tweak codegen logic for formatting, documentation and handling of anyOf

Xtansia:
### Description
Mmainly impacts work/generation to come, but PRing separately to ease reviewing.

### DIFF
--- a/java-codegen/opensearch-openapi.yaml
+++ b/java-codegen/opensearch-openapi.yaml
@@ -2,13 +2,15 @@ openapi: 3.1.0
 info:
   title: OpenSearch API Specification
   version: 1.0.0
-  x-api-version: 2.15.0
+  x-api-version: 2.16.0
 paths:
   /:
     get:
       operationId: info.0
       x-operation-group: info
       x-version-added: '1.0'
+      x-distributions-excluded:
+        - amazon-serverless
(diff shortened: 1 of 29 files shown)


java-codegen/opensearch-openapi.yaml
```diff
@@ -334,6 +342,9 @@ paths:
       operationId: indices.clear_cache.0
       x-operation-group: indices.clear_cache
       x-version-added: '1.0'
+      x-distributions-excluded:
```
reta, line 345: @Xtansia I understand the bias towards supporting AWS, but I honestly believe the spec has to be vendor neutral

Xtansia: @reta I understand your concern, in my opinion the spec is vendor neutral, it's just not vendor agnostic. This information is not just useful for future documentation purposes but also for automated testing against any cluster distribution (opensource, AWS managed or otherwise). I for one would certainly welcome other vendors to annotate any APIs they do not support or aid in setting up automated testing. And of course open an issue over on [opensearch-api-specification](https://github.com/opensearch-project/opensearch-api-specification/issues) so we can discuss this more in depth if you like 😄

dblock (approved):
LGTM

PR #1158, opened 2024-08-26
Generate the ml namespace

Xtansia:
### Description
Generates the ml namespace.

Will do a manual backport to 2.x to avoid the breaking changes in ShardStatistics.

### DIFF
--- a/java-codegen/src/main/java/org/opensearch/client/codegen/utils/Versions.java
+++ b/java-codegen/src/main/java/org/opensearch/client/codegen/utils/Versions.java
@@ -1,3 +1,11 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
 package org.opensearch.client.codegen.utils;
 
 import javax.annotation.Nonnull;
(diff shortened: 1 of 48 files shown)


dblock (approved):
Nice introduction of semver.

java-codegen/src/main/java/org/opensearch/client/codegen/utils/Versions.java
```diff
@@ -0,0 +1,25 @@
...
+
+package org.opensearch.client.codegen.utils;
+
+import javax.annotation.Nonnull;
+import org.semver4j.Semver;
+
+public final class Versions {
+    private Versions() {}
+
+    public static final Semver V2_0_0 = Semver.of(2, 0, 0);
```
reta, line 17: @Xtansia so we effectively 3 major versions now: 1.x , 2.x  - released, 3.x - unreleased. Why we only consider 2.0.0 here?

Xtansia: @reta This was primarily introduced for handling the fact `WriteResponseBase` used to include the doc `type` field  which has been already been removed in the client. Mainly trying to keep things as back-portable as possible for `2.x` for now. Can always expand upon this as the need arises

reta (approved):

PR #1170, opened 2024-08-29
Invert abstractness of generated vs "handwritten" client classes

Xtansia:
### Description
Invert abstractness of generated vs "handwritten" client classes. For the client classes that are "split" between generated and not, makes the generated part the abstract/"Base" class. Moving definitions from the sub-class to the base class will be API compatible. This will make it feasible to backport the codegen to 2.x.

### DIFF
--- a/java-codegen/src/main/java/org/opensearch/client/codegen/model/Type.java
+++ b/java-codegen/src/main/java/org/opensearch/client/codegen/model/Type.java
@@ -15,7 +15,6 @@ import com.samskivert.mustache.Mustache;
 import java.util.Set;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import org.checkerframework.checker.nullness.qual.NonNull;
 import org.opensearch.client.codegen.renderer.lambdas.TypeIsDefinedLambda;
 import org.opensearch.client.codegen.renderer.lambdas.TypeQueryParamifyLambda;
 import org.opensearch.client.codegen.renderer.lambdas.TypeSerializerLambda;
@@ -48,7 +47,7 @@ public class Type {
     }
 
     @Nullable
-    private final String package_;
+    private final String packageName;
     @Nonnull
     private final String name;
     @Nullable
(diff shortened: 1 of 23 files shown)


dblock (approved):

java-codegen/src/main/java/org/opensearch/client/codegen/model/Type.java
```diff
@@ -12,12 +12,15 @@
 import static org.opensearch.client.codegen.model.Types.Java;
 
 import com.samskivert.mustache.Mustache;
-import java.util.Arrays;
 import java.util.Set;
-import java.util.stream.Collectors;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import org.checkerframework.checker.nullness.qual.NonNull;
```
reta, line 18: Probably wrong import, `javax.annotation.Nonnull` is the one to use?

java-codegen/src/main/java/org/opensearch/client/codegen/model/Type.java
```diff
@@ -44,37 +47,41 @@ public static Builder builder() {
         return new Builder();
     }
 
-    private final String pkg;
+    @Nullable
+    private final String package_;
```
reta, line 51:
```suggestion
    private final String packageName;
```

Or

```suggestion
    private final String package;
```

please

java-codegen/src/main/java/org/opensearch/client/codegen/model/Type.java
```diff
@@ -206,34 +213,43 @@ public Mustache.Lambda isDefined() {
         return new TypeIsDefinedLambda(this);
     }
 
-    public static final class Builder {
-        private String pkg;
+    public static final class Builder extends ObjectBuilderBase<Type, Builder> {
+        private String package_;
```
reta, line 217:
```suggestion
        private String packageName;
```

or 

```suggestion
        private String package;
```

please

java-codegen/src/main/java/org/opensearch/client/codegen/model/TypeParameterDefinition.java
```diff
@@ -0,0 +1,72 @@
...
+import javax.annotation.Nullable;
+import org.jetbrains.annotations.NotNull;
+import org.opensearch.client.codegen.utils.ObjectBuilderBase;
+import org.opensearch.client.codegen.utils.Strings;
+
+public final class TypeParameterDefinition {
+    @Nonnull
+    private final String name;
+    @Nullable
+    private final Type extends_;
```
reta, line 21: Same please

reta (approved):

PR #1201, opened 2024-09-19
Generate hashCode and equals

### LINKED ISSUE #312: [FEATURE] equals/hashCode on API objects
While developing unit tests for code that makes use of `opensearch-java` we noticed that the invocations are pretty hard to mock/verify (e.g., using Mockito) as the API objects do not implement equals/hashCode. This causes the effect that it is very hard to `verify` that a method has been invoked with a certain API object as its argument just uses `Object.equals` for its `equals` implementation.

Implement equals/hashCode on all API POJOs.

miguel-vila:
### Description

Generate `hashCode` and `equals` methods.

### Issues Resolved

#312


java-codegen/src/main/resources/org/opensearch/client/codegen/templates/ObjectShape/HashCode.mustache
```diff
@@ -0,0 +1,17 @@
+    public int hashCode() {
+        int result = 17;
+        {{#fields}}
```
Xtansia, line 3: This needs to also account for `super.hashCode()`

java-codegen/src/main/resources/org/opensearch/client/codegen/templates/ObjectShape/Equals.mustache
```diff
@@ -0,0 +1,21 @@
+    public boolean equals(Object o) {
+        {{#extendsOtherShape}}
+            if(!super.equals(o)) {
+                return false;
+            }
+        {{/extendsOtherShape}}
+        if(this == o) return true;
+        if(this.getClass() != o.getClass()) return false;
+        {{className}} other = ({{className}})o;
+        {{#fields}}
```
Xtansia, line 10: The if chains get quite long and verbose for classes with many fields. While not critical as these files aren't to be hand edited anyway, we could possibly collapse this to a `&&` chain.

Xtansia (commented):
@miguel-vila This is really cool to see, thank you for picking this up!

Have added a few comments.

java-codegen/src/main/resources/org/opensearch/client/codegen/templates/ObjectShape/Equals.mustache
miguel-vila: changed to a `&&` chain at [8a150f3](https://github.com/opensearch-project/opensearch-java/pull/1201/commits/8a150f31e2ea4a60f2ebdc48ecadefd6d592108f)

java-codegen/src/main/resources/org/opensearch/client/codegen/templates/ObjectShape/HashCode.mustache
miguel-vila: done as part of [8fd4694](https://github.com/opensearch-project/opensearch-java/pull/1201/commits/8fd4694c6f4952e1ba2b18453b8603e23e74b514)

java-codegen/src/main/resources/org/opensearch/client/codegen/templates/ObjectShape/HashCode.mustache
```diff
@@ -0,0 +1,17 @@
...
+            {{#isUnboxedPrimitive}}
+                result = 31 * result + {{getBoxedType}}.hashCode(this.{{name}});
+            {{/isUnboxedPrimitive}}
+            {{^isUnboxedPrimitive}}
+                {{#isRequired}}
+                    result = 31 * result + this.{{name}}.hashCode();
+                {{/isRequired}}
+                {{^isRequired}}
+                    result = 31 * result + ({{name}} != null ? this.{{name}}.hashCode() : 0);
+                {{/isRequired}}
```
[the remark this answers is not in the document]
miguel-vila: I think I have to configure some import in order for this to work, but I'm not sure how to do that. In the meantime I'm using the FQN

miguel-vila:
@Xtansia I have addressed most of your comments, the remaining one is regarding using an import instead of the fqn for `java.util.Objects`. Not sure how to do that. 

I'm also interested in these changes because I want to have these methods in classes like `TypeMapping` and `IndexSettings` but I see that they are not being codegenerated at the moment. What should be done for that?

java-codegen/src/main/resources/org/opensearch/client/codegen/templates/ObjectShape/Equals.mustache
```diff
@@ -0,0 +1,27 @@
...
+        {{/hasFieldsToSerialize}}
+        {{#hasFieldsToSerialize}}
+            {{className}} other = ({{className}})o;
+            return
+            {{#fields}}
+                {{#isUnboxedPrimitive}}
+                    this.{{name}}() == other.{{name}}()
+                {{/isUnboxedPrimitive}}
+                {{^isUnboxedPrimitive}}
+                    java.util.Objects.equals(this.{{name}}, other.{{name}})
```
Xtansia, line 20: The way to reference the defined type in `Types` and get it added to the imports is like so:
```suggestion
                    {{TYPES.Java.Util.Objects}}.equals(this.{{name}}, other.{{name}})
```

java-codegen/src/main/resources/org/opensearch/client/codegen/templates/ObjectShape/Equals.mustache
```diff
@@ -0,0 +1,27 @@
+    public boolean equals(Object o) {
+        {{#extendsOtherShape}}
+            if(!super.equals(o)) {
+                return false;
+            }
+        {{/extendsOtherShape}}
+        if(this == o) return true;
+        if(this.getClass() != o.getClass()) return false;
+        {{^hasFieldsToSerialize}}
```
Xtansia, line 9: This is the wrong conditional to use as it'll ignore query params in request shapes when there's no "body" fields, e.g in `tasks.ListRequest`: https://github.com/opensearch-project/opensearch-java/blob/a7a53e26ec5ad254c4ac325fbcaa84a2ca43bfdc/java-client/src/generated/java/org/opensearch/client/opensearch/tasks/ListRequest.java#L374-L390

Xtansia (changes requested):

Xtansia:
> I'm also interested in these changes because I want to have these methods in classes like `TypeMapping` and `IndexSettings` but I see that they are not being codegenerated at the moment. What should be done for that?

@miguel-vila Yeah, we're currently doing a bit of a piecemeal approach to get things to be generated. It's somewhat slow going as when we do generate new sections there are breaking changes and other differences with the current client implementation, which requires evaluating between the client, the spec & the server what the correct expectation is. 

So for some of the more complex structures it might be faster/easier to add them to your required types by hand. However if you're interested in helping out with moving even 1 or 2 requests to be generated that would be amazing.

miguel-vila:
@Xtansia I would prefer not to add those manually, but not sure what would be the changes needed to include them in the codegeneration. Locally, I modified the `OPERATION_MATCHER` constant in the `CodeGenerator` class to include `and(namespace("indices"), named("get"))` and I'm seeing some errors :

```
java.lang.UnsupportedOperationException: Can not get type name for oneOf: [org.opensearch.client.codegen.openapi.OpenApiSchema@134060ce[pointer=/components/schemas/_common.analysis:TokenFilter/oneOf/0], org.opensearch.client.codegen.openapi.OpenApiSchema@33443e63[pointer=/components/schemas/_common.analysis:TokenFilter/oneOf/1]]
        at org.opensearch.client.codegen.model.SpecTransformer.mapOneOf(SpecTransformer.java:467
...
```

Is it possible that these errors would require more changes than adding the methods manually?

Xtansia:
> @Xtansia I would prefer not to add those manually, but not sure what would be the changes needed to include them in the codegeneration. Locally, I modified the `OPERATION_MATCHER` constant in the `CodeGenerator` class to include `and(namespace("indices"), named("get"))` and I'm seeing some errors :
> 
> ```
> java.lang.UnsupportedOperationException: Can not get type name for oneOf: [org.opensearch.client.codegen.openapi.OpenApiSchema@134060ce[pointer=/components/schemas/_common.analysis:TokenFilter/oneOf/0], org.opensearch.client.codegen.openapi.OpenApiSchema@33443e63[pointer=/components/schemas/_common.analysis:TokenFilter/oneOf/1]]
>         at org.opensearch.client.codegen.model.SpecTransformer.mapOneOf(SpecTransformer.java:467
> ...
> ```
> 
> Is it possible that these errors would require more changes than adding the methods manually?

- For `TokenFilter` it'd require adding `title: ` properties to the `oneOf:` in the spec to be able to generate the tagged union with the correct "Kind"s (https://github.com/opensearch-project/opensearch-java/blob/main/java-client/src/main/java/org/opensearch/client/opensearch/_types/analysis/TokenFilter.java).
- `TokenFilterDefinition` will be more complicated though as support for the discriminated tagged unions isn't in the codegen yet (https://github.com/opensearch-project/opensearch-java/blob/main/java-client/src/main/java/org/opensearch/client/opensearch/_types/analysis/TokenFilterDefinition.java).

There's likely more edge cases that'll pop up with something as expansive as `indices.get`, but would also unblock a lot of things with getting them implemented. I'll give it an investigate and see if I can get it working.

reta (approved):

Xtansia (approved):

PR #1220, opened 2024-10-07
Begin generating index CRUD operations

Xtansia:
### Description
Begin generating indices.get, shortcutting a few of the more expansive types for property mapping and query types etc. Will make follow up PRs to generate them.


dblock (approved):

UPGRADING.md
```diff
@@ -51,4 +51,13 @@ After:
 - The `name` field, getter and builder method have been renamed to `repository`.
 
 ### CloneSnapshotRequest timeout removal
-- The `timeout` field, getter and builder method have been removed from `CloneSnapshotRequest` as it is not supported by OpenSearch.
\ No newline at end of file
+- The `timeout` field, getter and builder method have been removed from `CloneSnapshotRequest` as it is not supported by OpenSearch.
+
+### indices.Alias
+- The `indexRouting`, `routing` and `searchRouting` properties now accept a `List<String>` instead of a `String`.
```
dblock, line 57: For later, this could be patched to accept `String` as well, right?

UPGRADING.md
```diff
@@ -51,4 +51,13 @@ After:
...
+
+### InlineScript
+- The `lang` property now accepts a `ScriptLanguage` enum instead of a `String`.
```
dblock, line 60: This too could be made backwards compatible with a patch?

java-client/src/generated/java/org/opensearch/client/opensearch/_types/ExpandWildcard.java
```diff
@@ -30,32 +30,33 @@
...
 import org.opensearch.client.json.JsonpDeserializable;
 
+// typedef: _types.ExpandWildcard
+
 @JsonpDeserializable
+@Generated("org.opensearch.client.codegen.CodeGenerator")
 public enum ExpandWildcard implements JsonEnum {
     /**
-     * Match any data stream or index, including hidden ones.
+     * Match any index, including hidden ones.
```
reta, line 49: Could we bring back the previous comment? The note about data streams is important I think

java-client/src/generated/java/org/opensearch/client/opensearch/_types/Script.java
```diff
@@ -165,4 +169,18 @@ private static JsonpDeserializer<Script> buildScriptDeserializer() {
...
     public static final JsonpDeserializer<Script> _DESERIALIZER = JsonpDeserializer.lazy(Script::buildScriptDeserializer);
+
+    public int hashCode() {
+        int result = 17;
+        result = 31 * result + Objects.hashCode(this._kind);
+        result = 31 * result + Objects.hashCode(this._value);
+        return result;
+    }
+
+    public boolean equals(Object o) {
```
reta, line 180:
```suggestion
    @Override
    public boolean equals(Object o) {
```

java-client/src/generated/java/org/opensearch/client/opensearch/_types/Script.java
```diff
@@ -165,4 +169,18 @@ private static JsonpDeserializer<Script> buildScriptDeserializer() {
     }
 
     public static final JsonpDeserializer<Script> _DESERIALIZER = JsonpDeserializer.lazy(Script::buildScriptDeserializer);
+
+    public int hashCode() {
```
reta, line 173:
```suggestion
    @Override
    public int hashCode() {
```

java-client/src/generated/java/org/opensearch/client/opensearch/_types/ScriptBase.java
```diff
@@ -45,82 +52,109 @@
...
     }
 
     // ---------------------------------------------------------------------------------------------
-    protected static <BuilderT extends AbstractBuilder<BuilderT>> void setupScriptBaseDeserializer(ObjectDeserializer<BuilderT> op) {
 
+    protected static <BuilderT extends AbstractBuilder<BuilderT>> void setupScriptBaseDeserializer(ObjectDeserializer<BuilderT> op) {
         op.add(AbstractBuilder::params, JsonpDeserializer.stringMapDeserializer(JsonData._DESERIALIZER), "params");
+    }
 
+    public int hashCode() {
```
reta, line 148:
```suggestion
    @Override
    public int hashCode() {
```

java-client/src/generated/java/org/opensearch/client/opensearch/_types/ScriptBase.java
```diff
@@ -45,82 +52,109 @@
...
+        int result = 17;
+        result = 31 * result + Objects.hashCode(this.params);
+        return result;
     }
 
+    public boolean equals(Object o) {
```
reta, line 154:
```suggestion
    @Override
    public boolean equals(Object o) {
```

java-client/src/generated/java/org/opensearch/client/opensearch/_types/mapping/RuntimeField.java
```diff
@@ -0,0 +1,389 @@
...
+import org.opensearch.client.opensearch._types.Script;
+import org.opensearch.client.util.ApiTypeHelper;
+import org.opensearch.client.util.ObjectBuilder;
+import org.opensearch.client.util.ObjectBuilderBase;
+
+// typedef: _types.mapping.RuntimeField
+
+@JsonpDeserializable
+@Generated("org.opensearch.client.codegen.CodeGenerator")
+public class RuntimeField implements PlainJsonSerializable {
```
reta, line 61: We don't support RuntimeFields, the equivalent feature is derived fields (see please https://github.com/opensearch-project/OpenSearch/issues/12281)

java-client/src/generated/java/org/opensearch/client/opensearch/_types/mapping/SourceFieldMode.java
```diff
@@ -30,20 +30,36 @@
...
-    public static final ExistsResponse _INSTANCE = new ExistsResponse();
+@JsonpDeserializable
+@Generated("org.opensearch.client.codegen.CodeGenerator")
+public enum SourceFieldMode implements JsonEnum {
+    Disabled("disabled"),
 
-    public static final JsonpDeserializer<ExistsResponse> _DESERIALIZER = JsonpDeserializer.fixedValue(ExistsResponse._INSTANCE);
+    Stored("stored"),
 
+    Synthetic("synthetic");
```
reta, line 52: I think we don't support `synthetic` either

UPGRADING.md
reta: Why is that? I believe it does not match the server model (`AliasActions`) where this properties are `String`s

java-client/src/generated/java/org/opensearch/client/opensearch/indices/Alias.java
```diff
@@ -42,45 +50,46 @@
...
 @JsonpDeserializable
+@Generated("org.opensearch.client.codegen.CodeGenerator")
 public class Alias implements PlainJsonSerializable {
+
     @Nullable
     private final Query filter;
 
-    @Nullable
-    private final String indexRouting;
+    @Nonnull
```
reta, line 66: I believe this is not the shape of the API we support (https://www.elastic.co/guide/en/elasticsearch/reference/7.12/indices-aliases.html)

UPGRADING.md
Xtansia: Have undone this change / tweaked the spec. Was a case of a re-used schema between the query string params which accept a comma-separated list and the body fields which only accept/return strings.

java-client/src/generated/java/org/opensearch/client/opensearch/_types/ExpandWildcard.java
Xtansia: I'm unable to find any reference to expand wildcard in the context of data streams in any OS docs or in the source code. So I'm not sure if those comments are actually relevant to OS or if they're related to the expanded data stream APIs added in ES.

java-client/src/generated/java/org/opensearch/client/opensearch/_types/mapping/SourceFieldMode.java
Xtansia: Yeah, looked into it and this enum and the associated `mode` property on `SourceField` don't exist in OS.

Have made a PR removing it from the spec (https://github.com/opensearch-project/opensearch-api-specification/pull/652) and removed it here.

Xtansia:
@dblock @reta Could you please take another look at this? I think I've covered all the comments/concerns.

reta (approved):

dblock (approved):

PR #1224, opened 2024-10-09
Add bwc handling on non-string sort values on `Hit`s

### LINKED ISSUE #1128: [BUG] Sort values in the Hit type are wrongly always deserialized to string
### What is the bug?
Executing a search request that has sorting in it will produce SearchHits results where each Hit will have it's `sort` array field populated. However the `sort` array is always deserialized as string array and if one wants to use this value to repeat the request with `search_after` being set to the returned value of the `sort` this could result in an error from the server if the field type is not actually string. 

Create an index which has the following mapping
```
"creationDate": {
    "type": "date",
    "format": "date_time"
}
```

Java field definition
```
@Field(name = "creationDate", type = FieldType.Date, format = DateFormat.date_time)
private Date creationDate;

// Setters/Getters
```
Add a few documents that get indexed and then execute a search request followed by another search that utilizes `search_after`. For example
```
NativeQueryBuilder queryBuilder = NativeQuery.builder();
queryBuilder.withQuery(q -> q.matchAll(m -> m));
queryBuilder.withSort(List.of(
    (new SortOptions.Builder()).field(fb -> fb.field("creationDate").order(SortOrder.Asc))
        .build()));
queryBuilder.withPageable(PageRequest.of(0, 1));

List<Object> searchAfter = null;
boolean done = false;
while (!done) {
  queryBuilder.withSearchAfter(searchAfter);

  SearchHits<T> hits = operations.search(queryBuilder.build(), entityClass, getIndexCoordinates());
  List<SearchHit<T>> searchHits = hits.getSearchHits();

  boolean hasResults = searchHits != null && !searchHits.isEmpty();
  if (hasResults) {
    // This will always return List of strings
    searchAfter = searchHits.getLast().getSortValues();
  }

  if (!hasResults || searchAfter == null || searchAfter.isEmpty() || searchAfter.stream()
      .filter(Objects::nonNull)
      .toList()
      .isEmpty() || (searchHits.size() < 1)) {
    done = true;
  }
}
``` 

When executing the above code the server will return something like
```
{
    // metadata
    "hits": {
        // metadata
        "hits": [
            {
                // _source, _index, _id etc
                "sort": [
                    1720256355885
                ]
            }
        ]
    }
}
```

However during deserialization the client will deserialize the `sort` field to 
```
"sort": [
    "1720256355885"
]
```

Thus repeating the request with search after will produce an error. Something like
```
"root_cause": [
            {
                "type": "parse_exception",
                "reason": "failed to parse date field [1720256355885] with format [date_time]: [failed to parse date field [1720256355885] with format [date_time]]"
            }
        ]
```

The returned value types of the `sort` field must be maintained as returned from the server.

SpringBoot application

- The issue seems to be related with the internal Hit generated class and the deserializer associated with the `sort` member - check [here](https://github.com/opensearch-project/opensearch-java/blob/main/java-client/src/main/java/org/opensearch/client/opensearch/core/search/Hit.java#L781)
- Elastic java client seems to have addressed the same issue by changing the type of the `sort` member from `List<String>` to `List<FieldValue>`. Check [this](https://github.com/elastic/elasticsearch-java/commit/9b924bb376419d28738f9ae97d145ba77004115a#diff-2fde5bb5df450c5e85868a8d2c2d53a454caefda6814413f57c011d29fc5729aR512) commit

Xtansia:
### Description
Add bwc handling on non-string sort values on `Hit`s

### Issues Resolved
Fixes #1128 

### DIFF
--- a/java-client/src/main/java/org/opensearch/client/opensearch/core/search/Hit.java
+++ b/java-client/src/main/java/org/opensearch/client/opensearch/core/search/Hit.java
@@ -37,6 +37,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import org.opensearch.client.json.JsonData;
 import org.opensearch.client.json.JsonpDeserializer;
@@ -46,6 +47,7 @@ import org.opensearch.client.json.JsonpUtils;
 import org.opensearch.client.json.ObjectBuilderDeserializer;
 import org.opensearch.client.json.ObjectDeserializer;
 import org.opensearch.client.json.PlainJsonSerializable;
+import org.opensearch.client.opensearch._types.FieldValue;
 import org.opensearch.client.opensearch.core.explain.Explanation;
 import org.opensearch.client.util.ApiTypeHelper;
 import org.opensearch.client.util.ObjectBuilder;
@@ -98,7 +100,7 @@ public class Hit<TDocument> implements PlainJsonSerializable {
     @Nullable
     private final Long version;
 
-    private final List<String> sort;
+    private final List<FieldValue> sort;
 
     @Nullable
     private final JsonpSerializer<TDocument> tDocumentSerializer;
@@ -265,8 +267,19 @@ public class Hit<TDocument> implements PlainJsonSerializable {
 
     /**
      * API name: {@code sort}
+     *
+     * <p><b>NOTE: In version 3.0.0 of opensearch-java, this method will instead return a {@code List<FieldValue>}.</b></p>
      */
     public final List<String> sort() {
+        return this.sort.stream().map(FieldValue::_toJsonString).collect(Collectors.toList());
+    }
+
+    /**
+     * API name: {@code sort}
+     *
+     * <p><b>NOTE: In version 3.0.0 of opensearch-java, this method will be renamed to replace {@link #sort()}.</b></p>
+     */
+    public final List<FieldValue> sortVals() {
         return this.sort;
     }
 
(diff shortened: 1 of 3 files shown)


java-client/src/main/java/org/opensearch/client/opensearch/core/search/Hit.java
```diff
@@ -265,8 +267,19 @@ public final Long version() {
...
     public final List<String> sort() {
+        return this.sort.stream().map(FieldValue::_toJsonString).collect(Collectors.toList());
+    }
+
+    /**
+     * API name: {@code sort}
+     *
+     * <p><b>NOTE: In version 3.0.0 of opensearch-java, this method will be renamed to replace {@link #sort()}.</b></p>
+     */
+    public final List<FieldValue> sortVals() {
```
reta, line 282:
```suggestion
    public final List<FieldValue> sortFields() {
```

reta: For 3.x (as you mentioned) we could replace `sort()` directly

Xtansia: @reta I can change this naming, however it then won't match that already used to do this fix for `SearchRequest.searchAfter`: 
https://github.com/opensearch-project/opensearch-java/blob/2.x/java-client/src/main/java/org/opensearch/client/opensearch/core/SearchRequest.java#L741-L743

reta: @Xtansia ah, I see, sure, let's be consistent, thanks @Xtansia

reta (approved):

PR #1235, opened 2024-10-18
Add support for radial search on Neural query

### LINKED ISSUE #1226: Adding support for Radial Search in Neural Query
### What are you proposing?
Adding support for Radial Search in Neural Query

### What users have asked for this feature?
Users connecting to Opensearch using Java Client

### What problems are you trying to solve?
Currently the Java SDK only takes 'k' as part of Neural Query, whereas from Opensearch 2.15, we can pass min_score / max_distance as part of neural query. But I dont see these options with Java Client. 

### What is the developer experience going to be?
Devlopers can directly use that feature instead of moving to GenericClient.

#### Are there any security considerations? 
No

#### Are there any breaking changes to the API
No

### What is the user experience going to be?
Developers can build queries with ease.

#### Are there breaking changes to the User Experience?
No Breaking changes

### Why should it be built? Any reason not to?
To make it in sync with latest Radial search features in Opensearch 2.15 features.

### What will it take to execute?
Some enhancements to NeuralSearch query builders.

### Any remaining open questions?
No

Xtansia:
### Description
Add support for radial search on k-NN and Neural query types

### Issues Resolved
Fixes #1226 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 
 ### Added
 - Added `minScore` and `maxDistance` to `KnnQuery` ([#1166](https://github.com/opensearch-project/opensearch-java/pull/1166))
+- Added `minScore` and `maxDistance` to `NeuralQuery` ([#1235](https://github.com/opensearch-project/opensearch-java/pull/1235))
 
 ### Dependencies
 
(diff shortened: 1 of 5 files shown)


dblock:
I see the parameters present in the API spec. How far are we from getting this via the code generator?

Xtansia:
> I see the parameters present in the API spec. How far are we from getting this via the code generator?

The query types (plus a couple others) have special deserialization logic for handling the fact they're "single key dictionaries", keyed on field, and in Java (and .NET) the field value gets flattened into the same class as the other query properties. So requires tweaking the spec to be consistent in how they're represented to be able to add special handling to the generator. It is one of the next things I'm going to look into, and will need to make sure the spec changes work for the people working on protobuf representation of the search api.

dblock (approved):

PR #1256, opened 2024-11-01
AwsSdk2Transport throw exception when using ApacheHttpClient to make an unsupported DELETE/GET request with a body

### LINKED ISSUE #503: [Maintenance] Unit test coverage on AwsSdk2Transport
### What is the gap?
Paths in AwsSdk2Transport should be covered by unit tests that are more light weighted than integration tests.

### LINKED ISSUE #521: [BUG] Delete PIT and clear scroll request fails with sigv4 signature mismatch for java client 2.5.0
### What is the bug?
`deletePit` request to opensearch client fails with sigv4 signing mismatch

```
org.opensearch.client.opensearch._types.OpenSearchException: Request failed: [security_exception] The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details.

The Canonical String for this request should have been
'DELETE
/_search/point_in_time

... remaining request info...
```

Create an OpenSearchClient following instructions from here (https://opensearch.org/docs/2.7/clients/java/#connecting-to-amazon-opensearch-service) and then attempt to use the client's `deletePit` API.

I believe this is because it is a  DELETE request with a body. The `createPit` method works as intended. Sending the request results in a sigv4 error

The `deletePit` API call does not have a signature mismatch

Mac OS

```
implementation 'org.opensearch.client:opensearch-java:2.5.0'
implementation 'org.opensearch.client:opensearch-rest-client:2.7.0'
```

_Add any other context about the problem._

### LINKED ISSUE #712: [BUG] Clear scroll request fails with sigv4 signature mismatch for java client 2.8.0
### What is the bug?
`clearScroll` request to opensearch client fails with sigv4 signing mismatch:

org.opensearch.client.opensearch._types.OpenSearchException: Request failed: [security_exception] The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details.

The Canonical String for this request should have been
'DELETE
/_search/scroll

... remaining request info...

How can one reproduce the bug?

Create an OpenSearchClient following instructions from here (https://opensearch.org/docs/2.7/clients/java/#connecting-to-amazon-opensearch-service) and then attempt to use the client's `clearScroll` API.

I believe this is because it is a DELETE request with a body. The `search/scroll` method works as intended. Sending the request results in a sigv4 error

The `clearScroll` API call does not have a signature mismatch

Ubuntu 22.04
`implementation 'org.opensearch.client:opensearch-java:2.8.0'`

_Add any other context about the problem._

Xtansia:
### Description
The AWS SDK's ApacheHttpClient implementation does not send the request body on DELETE or GET requests, https://github.com/aws/aws-sdk-java-v2/blob/master/http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/impl/ApacheHttpRequestFactory.java#L118-L137.

Additionally moves to the supported `AwsV4HttpSigner` as `Aws4Signer` is now deprecated: https://github.com/aws/aws-sdk-java-v2/blob/88abec27e7d5d35b21545c7e05875a7cc3d0f46e/core/auth/src/main/java/software/amazon/awssdk/auth/signer/Aws4Signer.java

### Issues Resolved
Relates to #712 
Relates to #521 
Fixes #503 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 ### Dependencies
 
 ### Changed
+- Changed AwsSdk2Transport to pre-emptively throw an exception when using AWS SDK's ApacheHttpClient to make an unsupported DELETE/GET request with a body ([#1256](https://github.com/opensearch-project/opensearch-java/pull/1256))
 
 ### Deprecated
 
--- a/java-client/build.gradle.kts
+++ b/java-client/build.gradle.kts
@@ -229,6 +229,7 @@ dependencies {
     testImplementation("software.amazon.awssdk", "aws-crt-client", "[2.21,3.0)")
     testImplementation("software.amazon.awssdk", "apache-client", "[2.21,3.0)")
     testImplementation("software.amazon.awssdk", "netty-nio-client", "[2.21,3.0)")
+    testImplementation("software.amazon.awssdk", "url-connection-client", "[2.21,3.0)")
     testImplementation("software.amazon.awssdk", "sts", "[2.21,3.0)")
 
     testImplementation("org.apache.logging.log4j", "log4j-api","[2.17.1,3.0)")
--- a/java-client/build.gradle.kts
+++ b/java-client/build.gradle.kts
@@ -173,7 +173,6 @@ val integrationTest = task<Test>("integrationTest") {
 val opensearchVersion = "3.0.0-SNAPSHOT"
 
 dependencies {
-
     val jacksonVersion = "2.17.0"
     val jacksonDatabindVersion = "2.17.0"
 
@@ -210,21 +209,25 @@ dependencies {
     implementation("jakarta.annotation", "jakarta.annotation-api", "1.3.5")
 
     // Apache 2.0
-
     implementation("com.fasterxml.jackson.core", "jackson-core", jacksonVersion)
     implementation("com.fasterxml.jackson.core", "jackson-databind", jacksonDatabindVersion)
     testImplementation("com.fasterxml.jackson.datatype", "jackson-datatype-jakarta-jsonp", jacksonVersion)
 
     // For AwsSdk2Transport
-    "awsSdk2SupportCompileOnly"("software.amazon.awssdk","sdk-core","[2.15,3.0)")
-    "awsSdk2SupportCompileOnly"("software.amazon.awssdk","auth","[2.15,3.0)")
-    testImplementation("software.amazon.awssdk","sdk-core","[2.15,3.0)")
-    testImplementation("software.amazon.awssdk","auth","[2.15,3.0)")
-    testImplementation("software.amazon.awssdk","aws-crt-client","[2.15,3.0)")
-    testImplementation("software.amazon.awssdk","apache-client","[2.15,3.0)")
-    testImplementation("software.amazon.awssdk","sts","[2.15,3.0)")
+    "awsSdk2SupportCompileOnly"("software.amazon.awssdk", "sdk-core", "[2.21,3.0)")
+    "awsSdk2SupportCompileOnly"("software.amazon.awssdk", "auth", "[2.21,3.0)")
+    "awsSdk2SupportCompileOnly"("software.amazon.awssdk", "http-auth-aws", "[2.21,3.0)")
+    testImplementation("software.amazon.awssdk", "sdk-core", "[2.21,3.0)")
+    testImplementation("software.amazon.awssdk", "auth", "[2.21,3.0)")
+    testImplementation("software.amazon.awssdk", "http-auth-aws", "[2.21,3.0)")
+    testImplementation("software.amazon.awssdk", "aws-crt-client", "[2.21,3.0)")
+    testImplementation("software.amazon.awssdk", "apache-client", "[2.21,3.0)")
+    testImplementation("software.amazon.awssdk", "netty-nio-client", "[2.21,3.0)")
+    testImplementation("software.amazon.awssdk", "sts", "[2.21,3.0)")
+
     testImplementation("org.apache.logging.log4j", "log4j-api","[2.17.1,3.0)")
     testImplementation("org.apache.logging.log4j", "log4j-core","[2.17.1,3.0)")
+
     // EPL-2.0 OR BSD-3-Clause
     // https://eclipse-ee4j.github.io/yasson/
     implementation("org.eclipse", "yasson", "2.0.2")
@@ -236,6 +239,10 @@ dependencies {
     testImplementation("junit", "junit" , "4.13.2") {
         exclude(group = "org.hamcrest")
     }
+
+    // The Bouncy Castle License (MIT): https://www.bouncycastle.org/licence.html
+    testImplementation("org.bouncycastle", "bcprov-lts8on", "2.73.6")
+    testImplementation("org.bouncycastle", "bcpkix-lts8on", "2.73.6")
 }
 
 licenseReport {
--- a/java-client/build.gradle.kts
+++ b/java-client/build.gradle.kts
@@ -105,7 +105,13 @@ tasks.withType<ProcessResources> {
 
 tasks.withType<Javadoc>().configureEach{
     options {
+        this as StandardJavadocDocletOptions
         encoding = "UTF-8"
+        addMultilineStringsOption("tag").setValue(listOf(
+            "apiNote:a:API Note:",
+            "implSpec:a:Implementation Requirements:",
+            "implNote:a:Implementation Note:",
+        ))
     }
 }
 
--- a/java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
+++ b/java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
@@ -19,6 +19,7 @@ import java.net.SocketTimeoutException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URLEncoder;
+import java.time.Clock;
 import java.util.AbstractMap;
 import java.util.Collection;
 import java.util.Map;
@@ -26,7 +27,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
-import java.util.function.Supplier;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.zip.GZIPInputStream;
 import javax.annotation.CheckForNull;
@@ -36,6 +37,7 @@ import javax.net.ssl.SSLHandshakeException;
 import org.opensearch.client.json.JsonpDeserializer;
 import org.opensearch.client.json.JsonpMapper;
 import org.opensearch.client.json.jackson.JacksonJsonpMapper;
+import org.opensearch.client.opensearch.OpenSearchClient;
 import org.opensearch.client.opensearch._types.ErrorCause;
 import org.opensearch.client.opensearch._types.ErrorResponse;
 import org.opensearch.client.opensearch._types.OpenSearchException;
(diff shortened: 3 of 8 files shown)


java-client/build.gradle.kts
```diff
@@ -210,21 +209,25 @@ dependencies {
...
-    "awsSdk2SupportCompileOnly"("software.amazon.awssdk","auth","[2.15,3.0)")
-    testImplementation("software.amazon.awssdk","sdk-core","[2.15,3.0)")
-    testImplementation("software.amazon.awssdk","auth","[2.15,3.0)")
-    testImplementation("software.amazon.awssdk","aws-crt-client","[2.15,3.0)")
-    testImplementation("software.amazon.awssdk","apache-client","[2.15,3.0)")
-    testImplementation("software.amazon.awssdk","sts","[2.15,3.0)")
+    "awsSdk2SupportCompileOnly"("software.amazon.awssdk", "sdk-core", "[2.21,3.0)")
+    "awsSdk2SupportCompileOnly"("software.amazon.awssdk", "auth", "[2.21,3.0)")
+    "awsSdk2SupportCompileOnly"("software.amazon.awssdk", "http-auth-aws", "[2.21,3.0)")
+    testImplementation("software.amazon.awssdk", "sdk-core", "[2.21,3.0)")
```
reta, line 220: I believe we should be using `awsSdk2SupportTestImplementation` here:

```suggestion
    awsSdk2SupportTestImplementation("software.amazon.awssdk", "sdk-core", "[2.21,3.0)")
```

java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
```diff
@@ -85,6 +87,7 @@ public class AwsSdk2Transport implements OpenSearchTransport {
     private final String signingServiceName;
     private final Region signingRegion;
     private final JsonpMapper defaultMapper;
+    @Nonnull
```
reta, line 90: No particular take on that, but I believe all this fields are should not be null.

java-client/build.gradle.kts
```diff
@@ -348,40 +351,43 @@ publishing {
...
+    configurations[java11.implementationConfigurationName].extendsFrom(configurations.testImplementation.get())
+    configurations[java11.runtimeOnlyConfigurationName].extendsFrom(configurations.testRuntimeOnly.get())
+
+    dependencies {
+        testImplementation("org.opensearch.test", "framework", opensearchVersion) {
+            exclude(group = "org.hamcrest")
+        }
+
+        // Apache 2.0
+        testImplementation("org.wiremock", "wiremock", "3.9.2")
```
reta, line 371: Nothing against `wiremock`, but we already have all needed pieces:
 - `HttpServer` that comes from HttpCore 5 (example https://github.com/apache/httpcomponents-core/blob/master/httpcore5-reactive/src/test/java/org/apache/hc/core5/reactive/examples/ReactiveFullDuplexServerExample.java)
 - since we run test with JDK-11, we could use `com.sun.net.httpserver.HttpServer` that comes with standard library (see please https://docs.oracle.com/en/java/javase/11/docs/api/jdk.httpserver/com/sun/net/httpserver/HttpServer.html)

java-client/build.gradle.kts
Xtansia: Such a configuration doesn't exist, the `awsSdk2Support` feature/variant doesn't affect the test source set

java-client/build.gradle.kts
Xtansia: Have dropped wiremock and re-implemented the necessary features

dblock (dismissed):
Looks good. A lot going on that's somewhat unrelated to the signing fix, you might want to extract it first, but up to you.

CHANGELOG.md
```diff
@@ -51,9 +48,21 @@ This section is for maintaining a changelog for all breaking changes for the cli
 ### Removed
 
 ### Fixed
+- Fixed AWS SigV4 on delete requests when using AWS SDK's Apache client ([#1256](https://github.com/opensearch-project/opensearch-java/pull/1256))
 
 ### Security
 
+## [2.17.0] - 11/05/2024
```
reta, line 55: Nit, seems to be unrelated changes?

java-client/src/test/java/org/opensearch/client/transport/util/GeneratedCertificateSSLContext.java
```diff
@@ -0,0 +1,315 @@
...
+import sun.security.x509.KeyUsageExtension;
+import sun.security.x509.SerialNumber;
+import sun.security.x509.SubjectAlternativeNameExtension;
+import sun.security.x509.SubjectKeyIdentifierExtension;
+import sun.security.x509.X500Name;
+import sun.security.x509.X509CertImpl;
+import sun.security.x509.X509CertInfo;
+
+@SuppressWarnings("sunapi")
+public class GeneratedCertificateSSLContext {
```
reta, line 64: I think it would be much simpler to pre-generate certs, no? (see please https://github.com/opensearch-project/OpenSearch/tree/main/plugins/transport-reactor-netty4/src/test/resources).

reta: (at worst, the security plugin uses Bouncycastle to generate those https://github.com/opensearch-project/security/blob/main/src/test/java/org/opensearch/security/ssl/CertificatesRule.java)

Xtansia: Pre-generating and committing the certs unfortunately leads to a lot of false positive security alerts. I'd initially avoided using BC seeing as the changes were with the aim to avoid pulling in another dependency. But can swap over to BC

java-client/src/test/java/org/opensearch/client/transport/aws/AwsSdk2TransportTests.java
```diff
@@ -0,0 +1,374 @@
...
+        NETTY_NIO_ASYNC
+    }
+
+    private static final String[] TEST_SERVICE_NAMES = { "aoss", "es", "arbitrary" };
+    private static final Region TEST_REGION = Region.AP_SOUTHEAST_2;
+    private static final String TEST_INDEX = "sample-index1";
+    private static final SSLContext SSL_CONTEXT;
+    private static final TrustManager[] CLIENT_TRUST_MANAGERS;
+
+    static {
```
reta, line 89: Why not to use @BeforeClass here?

java-client/src/test/java/org/opensearch/client/transport/util/SelfSignedCertificateAuthority.java
```diff
@@ -0,0 +1,149 @@
...
+import org.bouncycastle.cert.jcajce.JcaX509CertificateConverter;
+import org.bouncycastle.cert.jcajce.JcaX509ExtensionUtils;
+import org.bouncycastle.cert.jcajce.JcaX509v3CertificateBuilder;
+import org.bouncycastle.crypto.CryptoServicesRegistrar;
+import org.bouncycastle.jce.provider.BouncyCastleProvider;
+import org.bouncycastle.operator.ContentSigner;
+import org.bouncycastle.operator.OperatorCreationException;
+import org.bouncycastle.operator.jcajce.JcaContentSignerBuilder;
+
+public class SelfSignedCertificateAuthority {
```
reta, line 43: :+1: TY

java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
```diff
@@ -315,46 +312,57 @@ private <RequestT> SdkHttpFullRequest prepareRequest(
         } catch (URISyntaxException e) {
             throw new IllegalArgumentException("Invalid request URI: " + url.toString());
         }
+
+        ContentStreamProvider bodyProvider = body != null ? ContentStreamProvider.fromInputStreamSupplier(body::getInputStream) : null;
```
reta, line 316: So the `body` is `OpenSearchRequestBodyBuffer` and is always backed by buffer, could we use `ContentStreamProvider::fromByteArrayUnsafe` (zero copy) instead of creating a wrapping input stream over the array?

Xtansia:
Unfortunately on further digging and expanding the tests a bit, AWS SDK's ApacheHttpClient does not send the request body at all for DELETE or GET requests, https://github.com/aws/aws-sdk-java-v2/blob/master/http-clients/apache-client/src/main/java/software/amazon/awssdk/http/apache/internal/impl/ApacheHttpRequestFactory.java#L118-L137

So this doesn't actually fix anything, as such I've pivoted to pre-emptively throwing an exception as the request would fail signature validation anyway and gives a clear reasoning.

reta:
> @Xtansia Probably worth opening an issue in that SDK, AFAIK nothing says `GET` or `DELETE` can't have a body.

@dblock indeed the spec [1] is not strict on that but the general recommendation (like [2]) is that GET should not have a body, DELETE though may [3]. Fe, https://www.rfc-editor.org/rfc/rfc9110.html#name-get says (similar for DELETE):


> Although request message framing is independent of the method used, content received in a GET request has no generally defined semantics, cannot alter the meaning or target of the request, and might lead some implementations to reject the request and close the connection because of its potential as a request smuggling attack ([Section 11.2](https://www.rfc-editor.org/rfc/rfc9112#section-11.2) of [[HTTP/1.1](https://www.rfc-editor.org/rfc/rfc9110.html#HTTP11)]).  A client SHOULD NOT generate content in a GET request unless it is made directly to an origin server that has previously indicated, in or out of band, that such a request has a purpose and will be adequately supported. An origin server SHOULD NOT rely on private agreements to receive content, since participants in HTTP communication are often unaware of intermediaries along the request chain.

[1] https://www.rfc-editor.org/rfc/rfc9110.html#name-changes-from-rfc-7231
[2] https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET
[3] https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/DELETE

java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
```diff
@@ -386,14 +424,14 @@ private void applyOptionsHeaders(SdkHttpFullRequest.Builder builder, TransportOp
...
+        SdkHttpRequest httpRequest,
+        OpenSearchRequestBodyBuffer requestBody,
         Endpoint<?, ResponseT, ?> endpoint,
         TransportOptions options
     ) throws IOException {
-
         HttpExecuteRequest.Builder executeRequest = HttpExecuteRequest.builder().request(httpRequest);
-        if (httpRequest.contentStreamProvider().isPresent()) {
-            executeRequest.contentStreamProvider(httpRequest.contentStreamProvider().get());
+        if (requestBody != null) {
```
reta, line 433: Why we cannot reuse the `contentStreamProvider` (as before) but have to pass `requestBody` and create a new `ContentStreamProvider` ?

java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
```diff
@@ -118,6 +122,10 @@ public AwsSdk2Transport(
      * @param options Options that apply to all requests. Can be null. Create with
      *                {@link AwsSdk2TransportOptions#builder()} and use these to specify non-default credentials,
      *                compression options, etc.
+     *
+     * @implNote Using {@code software.amazon.awssdk.http.apache.ApacheHttpClient} is discouraged as it does not support request bodies on GET or DELETE requests.
```
reta, line 126: :+1:

Xtansia:
I might just try PR'ing a fix into aws-sdk seeing as it's relatively simple change, and hopefully mean more likely to actually get fixed.

Xtansia:
Created https://github.com/aws/aws-sdk-java-v2/pull/5704

dblock:
> Created [aws/aws-sdk-java-v2#5704](https://github.com/aws/aws-sdk-java-v2/pull/5704)

Good work 👏

reta (approved):

Xtansia:
@houssain-barouni Yes, I have created #1333

PR #1296, opened 2024-11-15
Add support for disabling typed keys serialization

### LINKED ISSUE #741: Typed keys in aggregation results cannot be switched off
### What is the bug?
The client always adds a type prefix to aggregation keys in responses which breaks compatibility with other clients or with frontend applications that directly work on aggregation responses.

This is a breaking change compared to other clients' behavior.

The Elasticsearch client API has a switch for turning off that behavior and thus returning compatible results.
See [Serializing aggregations and suggestions without typed keys](https://www.elastic.co/guide/en/elasticsearch/client/java-api-client/current/serialize-without-typed-keys.html)

Each aggregation result contains aggregation names combined as `[type]#[name]`

There should be a compatibility setting that allows returning aggregation names without type prefix like the one provided by Elasticsearch.

OpenSearch 2.11.0
OpenSearch Java Client 2.8.1
Java 17

image

./.

Xtansia:
### Description
Add support for disabling typed keys serialization

### Issues Resolved
Fixes #741 

### DIFF
--- a/java-client/src/main/java/org/opensearch/client/json/JsonpMapperBase.java
+++ b/java-client/src/main/java/org/opensearch/client/json/JsonpMapperBase.java
@@ -39,6 +39,7 @@ import java.lang.reflect.Field;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 import javax.annotation.Nullable;
 
 public abstract class JsonpMapperBase implements JsonpMapper {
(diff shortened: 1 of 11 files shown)


java-client/src/main/java/org/opensearch/client/json/JsonpMapperBase.java
```diff
@@ -36,9 +36,45 @@
...
+    }
+
+    /**
+     * Updates attributes to a copy of the current ones with an additional key/value pair.
+     * Mutates the current mapper, intended to be used in implementations of {@link #withAttribute(String, Object)}
+     */
+    protected JsonpMapperBase addAttribute(String name, Object value) {
+        if (attributes == null) {
+            this.attributes = Collections.singletonMap(name, value);
+        } else {
```
reta, line 67: Just to skip attributes recreation if the one in question is set already?
```
} else if (!this.attributes.containsKey(name) || !Objects.equals(this.attributes.get(name), value) {
```

Xtansia: Have added handling of pre-existing values

reta (approved):

Issue #420, opened 2023-03-30
[BUG] Major inconsistencies between Java client and OpenSearch API


AmiStrn:
### What is the bug?
Many API's are broken and unusable using the client due to the fact that the auto-generated code is relying on a later Elasticsearch version where the API has changed since 7.10.2

### How can one reproduce the bug?
run
```
GetSnapshotResponse getSnapshotResponse = javaClient().snapshot().get(g -> g.repository(“test-repository”).snapshot(“test-snapshot-5"));
```

fails with:
```
error: Missing required property ‘GetSnapshotResponse.total’
```

what is this field `total`? 
see it here in the [elasticsearch docs version 8](https://www.elastic.co/guide/en/elasticsearch/reference/current/get-snapshot-api.html#get-snapshot-api-prereqs:~:text=fetch%20additional%20results.-,total,-(integer)%20The%20total): 
> total (integer) The total number of snapshots that match the request when ignoring size limit or after query parameter.

Notice that the param is missing from the response body in same [docs page in version 7.10](https://www.elastic.co/guide/en/elasticsearch/reference/7.10/get-snapshot-api.html#get-snapshot-api-prereqs)

### What is the expected behavior?
A valid response with no errors

This is causing the client to be impractical for use in production from our standpoint. as it seems to be auto-generated from an incompatible Elasticsearch client!

AmiStrn:
Thanks @MarinaRazumovsky for finding these issues 💪

reta:
PS: This is not just `opensearch-java` issue but covers all clients fleet.

AmiStrn:
Thank you @reta for the quick response, I was led here by the docs. And we wish to use the official client.

> fixing the APIs which people use and report inconsistencies with

Currently, that list is long. We can fix some, but it seems like a futile task since the base code needs to be taken from a client that is compatible with elasticsearch 7.10.  And then we would gladly build on top of that.

Xtansia:
The particular issue with the `GetSnapshotResponse` was fixed in v2.19.0.

Additionally re-generation efforts for v3.0.0 of the client should resolve a large number of other similar issues.

PR #1299, opened 2024-11-17
Generate GetSnapshot

hogesako:
### Description
Generate GetSnapshot
https://github.com/opensearch-project/opensearch-api-specification/issues/606

### Issues Resolved
GetSnapshot is just an example and may not be resolved.
https://github.com/opensearch-project/opensearch-java/issues/420 


Xtansia (changes requested):
This is great @hogesako, thank you for picking this up end-to-end! 🙌 

This is somewhat tricky from the breaking change perspective, as it's technically breaking, but it is just removing things that were never supported by OpenSearch and were stopping making GetSnapshot requests, so is also really a fix.

If we're wanting to backport this to 2.x, this needs an entry added to the CHANGELOG, otherwise if it's just going in the eventual 3.x it should be documented in UPGRADING.

@reta Do you have any opinions or reservations about backporting this to 2.x?

java-client/src/generated/java/org/opensearch/client/opensearch/snapshot/SnapshotInfo.java
```diff
@@ -50,55 +57,64 @@
...
 
     @Nullable
     private final ShardStatistics shards;
 
+    @Nonnull
+    private final String snapshot;
+
     @Nullable
-    private final Time startTime;
+    private final String startTime;
```
reta, line 109: Shouldn't the `startTime` stayed as `Time`? Or DateTime?

reta:
> @reta Do you have any opinions or reservations about backporting this to 2.x?

@Xtansia I think everything looks fine except `SnapshotInfo` that changes quite a few data types. We probably could manually backport it in a backward compatible way, what do you think?

java-client/src/generated/java/org/opensearch/client/opensearch/snapshot/SnapshotInfo.java
hogesako: It seems that common___DateTime is being converted to a String.
Should I override them individually in the Overrides class?
```
            .with(
                schema("snapshot._common", "SnapshotInfo"),
                so -> so.withProperties(
                    mb -> mb.with("start_time", po -> po.withMappedType(Types.Client.OpenSearch._Types.Time))
                        .with("end_time", po -> po.withMappedType(Types.Client.OpenSearch._Types.Time))
                )
            )
```

Xtansia: So the [`Time`](https://github.com/opensearch-project/opensearch-java/blob/main/java-client/src/main/java/org/opensearch/client/opensearch/_types/Time.java#L51) type is (confusingly) actually the Duration type, which is why it's used for `clusterManagerTimeout` etc. So `startTime` being `Time` was incorrect anyway. 

I hadn't added any special handling for `DateTime` as the only property I'd run across so far in my generation with that schema is [`OpenSearchVersionInfo.buildDate`](https://github.com/opensearch-project/opensearch-java/blob/dc41a99f1fd76fa59db8aa28c5466602cbcc6eb1/java-client/src/main/java/org/opensearch/client/opensearch/_types/OpenSearchVersionInfo.java#L49) which was already a String type.

reta: I see, thanks

Xtansia: Happy for us to add a special type mapping for DateTime, it's only `OpenSearchVersionInfo.buildDate` is a bit special/annoying because it can return the string `"unknown"` so it needs to be String where others can be whatever date type.

reta: Sounds great to me, thanks!

hogesako:
@Xtansia 
I have appended to CHANGELOG.
There is a data types change, is there any other work needed?

Xtansia (dismissed):

Xtansia (approved):

reta (approved):

Xtansia (approved):

reta (approved):

hogesako:
@Xtansia 
Will this PR be backported to 2.x?

Xtansia:
@hogesako It will need to be manually backported but tweaked to minimise breaking changes in 2.x

Issue #1171, opened 2024-08-29
[BUG] JsonException when SearchRequest contains a query of type range using .toJsonString function


mikeperello-scopely:
### What is the bug?
The `.toJsonString` function throws a `JsonException` when a `SearchRequest` includes a query of type `range`. 

Specifically, it shows the following error:
``
Cannot find a serializer for type <type>. Consider using a full-featured JsonpMapper.
``
### How can one reproduce the bug?
````
# Create a Query of type range
    Query rangeQuery = Query.of(q -> q.range(r -> r
            .field("time")
            .format("strict_date_optional_time")
            .gte(JsonData.of("2024-06-29T09:50:53Z"))
            .lte(JsonData.of("2024-06-29T10:50:53Z"))));
    
# Add it as part of a boolQuery
    BoolQuery boolQuery = BoolQuery.of(b -> b
            .filter(rangeQuery));

# Add the boolQuery to the SearchRequest
    SearchRequest searchRequest = new SearchRequest.Builder()
            .query(q -> q.bool(boolQuery))
            .build();

# Execute the toJsonString()
    System.out.println("query: " + searchRequest.toJsonString());

````

### What is the expected behavior?
The query should print without any issues.

### Do you have any additional context?

Other types of query, such as `matchPhrase` seem to work correctly. However, when adding a query of type range, it throws the error.
I have been investigating the error, and it looks it's related to the parsing of the `gte` and `lte` elements. Although they are declared as `JsonData` types, the value is extracted as String (in this case), leading to the JsonException. 

This, however, does not seem to affect the `SearchRequest` functionality, as I have been able to query sucessfully.

matthias-koch:
I have this issue with all aggregation searches.

dblock:
@mikeperello-scopely @matthias-koch Would one of you please care to help writing a (failing) unit test for this? Let's fix it.

mikeperello-scopely:
Sure, @dblock! Added.

Xtansia:
The fix for this was released in v2.19.0 of the client

PR #1309, opened 2024-11-18
Fix toJsonString default mapper serialize type

playLeo:
### Description
The toJsonString method of SearchRequest uses the default mapper provided by JsonpUtils.
This mapper only supports serializing instances of the JsonpSerializable type.
Using JsonData values in a range query cannot be serialized.

To resolve this, add support for serializing Integer, Double, and String types in the default mapper of JsonpUtils.

### Issues Resolved
This PR is related to [#1171](https://github.com/opensearch-project/opensearch-java/issues/1171)


java-client/src/main/java/org/opensearch/client/json/JsonpUtils.java
```diff
@@ -81,15 +81,30 @@ public <T> void serialize(T value, JsonGenerator generator) {
                 return;
             }
 
+            if (value instanceof String) {
```
dblock, line 84: Maybe some `else` here is better and avoids the early `return`?

dblock (changes requested):
Thanks. Fix CI, add to CHANGELOG?

java-client/src/main/java/org/opensearch/client/json/JsonpUtils.java
reta, line 84: @playLeo I don't really follow this change: the `JsonpUtils::serialize(T value, JsonGenerator generator)` is called by OpenSearch Java model classes, which have to implement `JsonpSerializable`. It is never called on arbitrary types (like String, ...), since the model classes use `JsonGenerator` for those instead.

reta (changes requested):

java-client/src/main/java/org/opensearch/client/json/JsonpUtils.java
reta: Oh, I see, this is `JsonData` class, which does serialize raw values, I think the change should be contained with `JsonDataImpl`  (as it does with JsonValue now).

java-client/src/main/java/org/opensearch/client/json/JsonDataImpl.java
```diff
@@ -109,6 +109,12 @@ public <T> T deserialize(JsonpDeserializer<T> deserializer, JsonpMapper mapper)
     public void serialize(JsonGenerator generator, JsonpMapper mapper) {
         if (value instanceof JsonValue) {
             generator.write((JsonValue) value);
+        } else if (value instanceof String) {
```
reta, line 112: :+1: thanks @playLeo , could you please just cover the missing types (Boolean, BigDecimal, Short, Long, Float, ...) thank you.

reta (approved):
Thanks @playLeo !

Xtansia (approved):

Xtansia:
@playLeo Could you please manually backport this PR to `2.x` following the above instructions?

playLeo:
@Xtansia I've created a PR, and issue #1199 can also be closed

PR #1336, opened 2024-12-04
[Backport 2.x] Generate GetSnapshot (#1299)

hogesako:
### Description
Backport #1299 to `2.x`
Please review to ensure compatibility is maintained.

### DIFF
--- a/java-client/src/main/java/org/opensearch/client/opensearch/snapshot/GetSnapshotResponse.java
+++ b/java-client/src/main/java/org/opensearch/client/opensearch/snapshot/GetSnapshotResponse.java
@@ -30,12 +30,18 @@
  * GitHub history for details.
  */
 
+//----------------------------------------------------
+// THIS CODE IS GENERATED. MANUAL EDITS WILL BE LOST.
+//----------------------------------------------------
+
 package org.opensearch.client.opensearch.snapshot;
 
 import jakarta.json.stream.JsonGenerator;
+import java.util.Collections;
 import java.util.List;
 import java.util.function.Function;
-import javax.annotation.Nullable;
+import javax.annotation.Generated;
+import javax.annotation.Nonnull;
 import org.opensearch.client.json.JsonpDeserializable;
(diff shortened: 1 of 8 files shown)


java-client/src/main/java/org/opensearch/client/opensearch/snapshot/GetSnapshotResponse.java
```diff
@@ -50,208 +57,130 @@
...
-     * Required - The total number of snapshots that match the request when ignoring
-     * size limit or after query parameter.
+     * * This method is left for compatibility. Always returns 0.
      * <p>
      * API name: {@code total}
      */
+    @Deprecated
     public final int total() {
-        return this.total;
+        return 0;
```
reta, line 99: May be:

```suggestion
        return snapshots().size();
```

Xtansia: Could also rather than removing the fields just make them optional, and just have the minor "breakage"/inconvenience of them going from `int` to `Integer`

Xtansia:
Please also run `./gradlew spotlessApply`

java-client/src/main/java/org/opensearch/client/opensearch/snapshot/SnapshotInfo.java
```diff
@@ -295,157 +323,134 @@ public final Long versionId() {
     }
 
     /**
-     * API name: {@code feature_states}
+     * This method is left for compatibility. Always returns an empty list.
      */
+    @Deprecated
     public final List<InfoFeatureState> featureStates() {
-        return this.featureStates;
+        return new ArrayList<>();
```
reta, line 330: Super nit:

```suggestion
        return Collections.emptyList();
```

reta: Or just delete is as well, since it was not supported anyway

reta (approved):

PR #1337, opened 2024-12-04
Generate query DSL types

Xtansia:
### Description
Generate Query DSL types.

There are some edge case query types that are excluded for now, and will be handled in a follow up PR as their schemas are a bit more complex to handle.


dblock:
Huge if true :)

reta (approved):

reta:
> Huge if true :)

It is but went through all, seems to be "easy" changes

PR #723, opened 2023-11-13
Fix PutTemplateRequest deserialization

Bfindlay:
### Description
Link to issue [417](https://github.com/opensearch-project/opensearch-java/issues/417)

This PR adds in the missing deserialization steps to the PutTemplateRequest _DESERIALIZER which would result in the error
`Missing required property 'PutTemplateRequest.name'` if you tried to deserialize a request object.

### Issues Resolved
https://github.com/opensearch-project/opensearch-java/issues/417

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1358, opened 2024-12-15
Generate more indices operations

Xtansia:
### Description
Generate more indices operations


java-client/src/generated/java/org/opensearch/client/opensearch/indices/AddBlockRequest.java
```diff
@@ -46,134 +53,145 @@
...
+
+    @Nonnull
     private final List<ExpandWildcard> expandWildcards;
 
     @Nullable
     private final Boolean ignoreUnavailable;
 
-    private final String index;
+    @Nonnull
+    private final List<String> index;
```
reta, line 85: Hm ... this seems to be out of sync, the name should be `indices` as per https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/admin/indices/readonly/AddIndexBlockRequest.java#L59 ?

java-client/src/generated/java/org/opensearch/client/opensearch/indices/AnalyzeRequest.java
```diff
@@ -483,17 +623,16 @@ public AnalyzeRequest build() {
...
 
     protected static void setupAnalyzeRequestDeserializer(ObjectDeserializer<AnalyzeRequest.Builder> op) {
-
         op.add(Builder::analyzer, JsonpDeserializer.stringDeserializer(), "analyzer");
         op.add(Builder::attributes, JsonpDeserializer.arrayDeserializer(JsonpDeserializer.stringDeserializer()), "attributes");
         op.add(Builder::charFilter, JsonpDeserializer.arrayDeserializer(CharFilter._DESERIALIZER), "char_filter");
         op.add(Builder::explain, JsonpDeserializer.booleanDeserializer(), "explain");
         op.add(Builder::field, JsonpDeserializer.stringDeserializer(), "field");
         op.add(Builder::filter, JsonpDeserializer.arrayDeserializer(TokenFilter._DESERIALIZER), "filter");
+        op.add(Builder::index, JsonpDeserializer.stringDeserializer(), "index");
```
reta, line 632: The index is not part of the JSON payload, is it? (it is URL path parameter)

java-client/src/generated/java/org/opensearch/client/opensearch/indices/ExistsAliasRequest.java
```diff
@@ -46,238 +53,312 @@
...
     private final Boolean allowNoIndices;
 
+    @Nonnull
     private final List<ExpandWildcard> expandWildcards;
 
     @Nullable
     private final Boolean ignoreUnavailable;
 
+    @Nonnull
     private final List<String> index;
```
reta, line 79: This seems to be also not in sync? Shouldn't be `indices`? https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/admin/indices/alias/get/GetAliasesRequest.java#L53

java-client/src/generated/java/org/opensearch/client/opensearch/indices/PutAliasRequest.java
```diff
@@ -52,25 +58,41 @@
...
+    private final String alias;
+
+    @Nullable
+    private final Time clusterManagerTimeout;
+
     @Nullable
     private final Query filter;
 
+    @Nonnull
     private final List<String> index;
```
reta, line 88: `indices` here as well? https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/admin/indices/alias/IndicesAliasesRequest.java#L106

java-client/src/generated/java/org/opensearch/client/opensearch/indices/AddBlockRequest.java
Xtansia: The spec copies the naming from the REST handler: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/rest/action/admin/indices/RestAddIndexBlockAction.java

Other operations which already accepted a list like [GetIndexRequest](https://github.com/opensearch-project/opensearch-java/blob/dc41a99f1fd76fa59db8aa28c5466602cbcc6eb1/java-client/src/main/java/org/opensearch/client/opensearch/indices/GetIndexRequest.java) still refer to it as `index`

java-client/src/generated/java/org/opensearch/client/opensearch/indices/AnalyzeRequest.java
Xtansia: Correct, we had a few request classes that'd been hand edited to allow deserializing the path params from JSON, so had incorporated it without to avoid breaking changes.  See https://github.com/opensearch-project/opensearch-java/pull/723

The problem we run into is that requests which have required path parameters are impossible to deserialize otherwise due to the required fields.

But can revert it everywhere if you think?

reta: > The index is not part of the JSON payload, is it? (it is URL path parameter)

I think the problem here is that this is not the JSON that will be sent to the server but I workaround to restore the state of the object. Yes, I would revert it everywhere, the path / query / etc params should not be part of the request payload (unless they are repeated for some reasons). Thank you.

Xtansia: They're only deserializable, they won't be serialized into the JSON sent to the server

reta: Exactly, so when they will be serialized into the JSON?

Xtansia: I believe that it's being used to construct the request shapes from externally generated JSON. 

If we're going to remove the ability to deserialize the path parameters, then we should completely remove deserialization from the request objects as it's unusable in the majority of cases otherwise. Or provide some other mechanism to pass the path (& query) params separate to the JSON payload for deserialization

reta: > Or provide some other mechanism to pass the path (& query) params separate to the JSON payload for deserialization

Yes, I think that would be the right option.

java-client/src/test/java/org/opensearch/client/opensearch/core/PutIndexTemplateRequestTest.java
```diff
@@ -16,13 +16,16 @@
...
+import org.junit.Ignore;
 import org.junit.Test;
 import org.opensearch.client.json.JsonpMapper;
 import org.opensearch.client.json.jsonb.JsonbJsonpMapper;
 import org.opensearch.client.opensearch.indices.PutIndexTemplateRequest;
 
 public class PutIndexTemplateRequestTest extends Assert {
 
+    // TODO: Allow constructing requests from JSON with required path params
+    @Ignore
```
reta, line 28: I think we could check that path params are not deserialized instead of ignoring

reta (dismissed):

reta (approved):

PR #1360, opened 2024-12-18
Generate majority of indices operations

Xtansia:
### Description
Generates all but one indices operation that requires some extra work/handling that will be done separately.


java-client/src/generated/java/org/opensearch/client/opensearch/_types/GetStats.java
```diff
@@ -193,61 +202,96 @@ protected void serializeInternal(JsonGenerator generator, JsonpMapper mapper) {
...
         generator.write(this.missingTimeInMillis);
 
         generator.writeKey("missing_total");
         generator.write(this.missingTotal);
 
         if (this.time != null) {
-            generator.writeKey("time");
-            generator.write(this.time);
-
+            generator.writeKey("getTime");
```
reta, line 215: The `getTime` does not look right. shouldn't be `time`?:

```suggestion
            generator.writeKey("time");
```

Xtansia: The field is actually `getTime`: https://github.com/opensearch-project/OpenSearch/blob/2.x/server/src/main/java/org/opensearch/index/get/GetStats.java#L158

Looks like this bug was introduced nearly twelve years ago: https://github.com/opensearch-project/OpenSearch/commit/cc83c2f848be69a77f1275fe1ff5363dcdd4c955#diff-f7abcc8f5eb3ec75bb9be11b7fe88baa0bcedf62c3c481552a0e98ceb7125a27R124

reta: Thanks @Xtansia

reta: Do you mind creating an issue on OpenSearch side?

reta (approved):

java-client/src/generated/java/org/opensearch/client/opensearch/_types/GetStats.java
Xtansia: Created https://github.com/opensearch-project/OpenSearch/issues/16894

PR #1407, opened 2025-02-03
Support KnnQuery's `method_parameters` and `rescore` properties

Xtansia:
### Description
Support KnnQuery's `method_parameters` and `rescore` properties. This is manually implemented feature to `2.x` as it can not be generated or auto-backported from `main` without introducing breaking changes. There will be follow up PRs to fix the spec and regenerate the class in `main`.


VijayanB:
Thanks for raising this PR. Can we also include this property too?
https://github.com/opensearch-project/k-NN/blob/d22f1aeee7a44f5e88c9693b84ffd81a7cc312ae/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java#L112

Xtansia:
> Thanks for raising this PR. Can we also include this property too? https://github.com/opensearch-project/k-NN/blob/d22f1aeee7a44f5e88c9693b84ffd81a7cc312ae/src/main/java/org/opensearch/knn/index/query/KNNQueryBuilder.java#L112

Have added it

VijayanB (approved):

PR #1443, opened 2025-02-18
[Backport 2.x] CodeGen updates

Xtansia:
### Description
Backport CodeGen updates to 2.x


java-client/src/main/java/org/opensearch/client/opensearch/cat/CatRequestBase.java
```diff
@@ -119,11 +137,40 @@ public final BuilderT headers(@Nullable String headers) {
          * API name: {@code s}
          * <p>
          */
-        public final BuilderT sort(@Nullable String sort) {
-            this.sort = sort;
+        @Nonnull
+        public final BuilderT sort(List<String> values) {
```
reta, line 141: So these changes are breaking (for 2.x release line) ... is it what we aiming for?

Xtansia: Have reverted these fields back to simply `String`. Forgot that `sort(String, String...)` wouldn't be binary compatible with `sort(String)` and also incompatible behavior of append vs overwrite.

reta (approved):

Issue #1463, opened 2025-02-26
[FEATURE] Change abstract builders visibility to public


cyvcloud:
### Is your feature request related to a problem?
When trying to use the query builder dsl in the Nashorn script engine certain builder methods are not accessible through Javascript because the abstract builder super classes are not public.

### What solution would you like?
Change abstract builders visibility to public. See also https://github.com/elastic/elasticsearch-java/commit/8260e83b2d117a024d8a8d17079427f93fd66ff6

### What alternatives have you considered?
No other way of using the query dsl from JavaScript with Nashorn seems possible.

### Do you have any additional context?
We could provide a pull request with the necessary changes. Would you be interested in accepting such a contribution?

Xtansia:
@cyvcloud Please feel free to raise a PR.

matthiasValuecloud:
@cyvcloud and me work at the same company and contributed with the above pull request.

Xtansia:
This was released as part of v2.22.0

PR #1464, opened 2025-02-27
Issues/1463

matthiasValuecloud:
### Description
Change the visiblity of builder from protected to public, so that nashorn engine can access inner classes.

### Issues Resolved
(https://github.com/opensearch-project/opensearch-java/issues/1463)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,7 +51,7 @@ This section is for maintaining a changelog for all breaking changes for the cli
 ### Dependencies
 
 ### Changed
-- Changed abstract builders visibility to public ([#000](https://github.com/opensearch-project/opensearch-java/pull/000))
+- Changed abstract builders visibility to public ([#1464](https://github.com/opensearch-project/opensearch-java/pull/1464))
 
 ### Deprecated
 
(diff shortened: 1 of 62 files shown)


Xtansia (changes requested):
Thanks for raising this PR. There are also some classes under the src/main source set that are not yet generated that will need to be changed as well

matthiasValuecloud:
> Thanks for raising this PR. There are also some classes under the src/main source set that are not yet generated that will need to be changed as well

There is now 1 class still missing :
>java-client/src/main/java/org/opensearch/client/transport/endpoints/DictionaryResponse.java
According to the equivalent Java class under elasticsearch, it is still protected. Therefore i skipped it so far, but we could also adjust this last class aswell.

Xtansia (approved):

Xtansia:
@matthiasValuecloud Could you please manually backport the changes to 2.x? It may be easier to recreate the changes from scratch against the `2.x` branch rather than trying to cherry pick per the above comment as 2.x has less generated classes.

matthiasValuecloud:
Yes i will do so in the next couple hours

matthiasValuecloud:
@Xtansia https://github.com/opensearch-project/opensearch-java/pull/1469
I think and hope it is okay for the added changelog to reference the "bigger" Pullrequest

PR #1483, opened 2025-03-14
Fix ClassCastException in extractAndWrapCause when handling Execution…

### LINKED ISSUE #1480: [BUG] ClassCastException occurs in extractAndWrapCause method of AwsSdk2Transport when ExecutionException's cause is not an Exception or Error
### What is the bug?
In the extractAndWrapCause method of AwsSdk2Transport, when handling an ExecutionException, the code assumes that executionException.getCause() is always an Exception or Error.

However, getCause() can return a Throwable that is neither an Exception nor an Error. 
If the cause is another type of Throwable (e.g., a custom subclass that is neither Error nor Exception), the following line will cause a ClassCastException:

`exception = (Exception)t;`

This results in the loss of the original exception details.

This issue can be reproduced when using NettyNioAsyncHttpClient as the SdkAsyncHttpClient with a low connectionAcquisitionTimeout value.

Steps to reproduce:
	1.	Set up an OpenSearchClient using AwsSdk2Transport with NettyNioAsyncHttpClient.
	2.	Configure the client with a very low connectionAcquisitionTimeout, such as 1 millisecond.
	3.	Attempt to make a request to OpenSearch
	4.	When a connectionAcquisitionTimeout occurs, it triggers an ExecutionException.
	5.	The extractAndWrapCause method processes the exception, but if the cause is a Throwable that is neither an Error nor an Exception, a ClassCastException is thrown.

This results in the loss of the original exception context, making it difficult to diagnose the actual timeout issue.
When a connectionAcquisitionTimeout occurs, the exception should be correctly propagated without causing a ClassCastException.
	•	If the cause of ExecutionException is an Error, it should be rethrown (which is already handled correctly in the current implementation).
	•	If the cause is an Exception, it should be processed as intended.
	•	If the cause is a Throwable that is neither an Error nor an Exception, the code should not attempt to cast it to an Exception, as this leads to a ClassCastException. Instead, it should be wrapped in a more appropriate exception (e.g., RuntimeException).

Mac OS Sonoma 14.6
opensearch-java:2.22.0
netty-nio-client:2.30.37
org.springframework.boot 3.2.7

N/A

If possible, I would be happy to create a PR to fix this issue.

wonthechan:
### Description
This fixes #1480 

### Issues Resolved
#1480 

### DIFF
--- a/java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
+++ b/java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
@@ -694,7 +694,11 @@ public class AwsSdk2Transport implements OpenSearchTransport {
             if (t instanceof Error) {
                 throw (Error) t;
             }
-            exception = (Exception) t;
+            if (t instanceof Exception) {
+                exception = (Exception) t;
+            } else {
+                exception = new RuntimeException("Unexpected throwable caught in execution", t);
+            }
         }
         if (exception instanceof SocketTimeoutException) {
             SocketTimeoutException e = new SocketTimeoutException(exception.getMessage());
--- a/java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
+++ b/java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
@@ -14,6 +14,7 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.UnsupportedEncodingException;
+import java.lang.reflect.UndeclaredThrowableException;
 import java.net.ConnectException;
 import java.net.SocketTimeoutException;
 import java.net.URI;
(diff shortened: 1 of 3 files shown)


java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
```diff
@@ -694,7 +694,11 @@ private static Exception extractAndWrapCause(Exception exception) {
             if (t instanceof Error) {
                 throw (Error) t;
             }
-            exception = (Exception) t;
+            if (t instanceof Exception) {
+                exception = (Exception) t;
+            } else {
+                exception = new RuntimeException("Unexpected throwable caught in execution", t);
```
reta, line 700: I am wondering if `UndeclaredThrowableException` would be an appropriate exception container in this case?

```suggestion
                exception = new UndeclaredThrowableException(t);
```

wonthechan: Thanks. It looks more appropriate to me too!

reta (approved):

reta:
Thank you @wonthechan !

PR #1524, opened 2025-04-14
Fix direct serialization of primitive types in `DEFAULT_JSONP_MAPPER`

### LINKED ISSUE #1503: [BUG] Minimal IndexRequest fails to serialize to JSON
### What is the bug?

`IndexRequest.toJsonString()` fails with `JsonException`.

Consider this minimal unit JUnit 5 test:

```java
package opensearchtest;

import static org.junit.jupiter.api.Assertions.assertNotNull;
import org.junit.jupiter.api.Test;
import org.opensearch.client.opensearch.core.IndexRequest;

public class OpenSearchSerializationTest {
    @Test
    void indexRequestToJson() {
        IndexRequest<Object> indexRequest = IndexRequest.of(i -> i
                .index("index")
                .document("document"));
        assertNotNull(indexRequest.toJsonString());
    }
}
```

It fails with:
```
jakarta.json.JsonException: Cannot find a serializer for type java.lang.String. Consider using a full-featured JsonpMapper.
	at org.opensearch.client.json.JsonpUtils$1.serialize(JsonpUtils.java:84)
	at org.opensearch.client.json.JsonpUtils.serialize(JsonpUtils.java:158)
	at org.opensearch.client.opensearch.core.IndexRequest.serialize(IndexRequest.java:288)
	at org.opensearch.client.json.PlainJsonSerializable.toJsonString(PlainJsonSerializable.java:29)
	at opensearchtest.OpenSearchSerializationTest.indexRequestToJson(OpenSearchSerializationTest.java:16)
```

Java 21, opensearch-java 2.22.0

This is an isolated test, but in the context of our application we use `toJsonString()` to log the content of various requests and responses. I can probably work around with a custom serialization utility, but I was expecting this to "just work". Why expose it otherwise?

Xtansia:
### Description
Fix direct serialization of primitive types in `DEFAULT_JSONP_MAPPER` as used by `toJsonString()`.

### Issues Resolved
Closes #1503 


java-client/src/main/java/org/opensearch/client/json/JsonpUtils.java
```diff
@@ -78,11 +80,31 @@ public JsonProvider jsonProvider() {
         public <T> void serialize(T value, JsonGenerator generator) {
             if (value instanceof JsonpSerializable) {
                 ((JsonpSerializable) value).serialize(generator, this);
-                return;
+            } else if (value instanceof JsonValue) {
```
reta, line 83: @Xtansia the DEFAULT_JSONP_MAPPER was not designed to be generic JSON mapper, it shouldn't be used with primitive types. Back to the issue: what is expectation of sending the index request (which is just a string) to OpenSearch?

bdelbosc:
Hi, any feedback on this PR? `SearchRequest#toJsonString` is not usable without this fix.

java-client/src/main/java/org/opensearch/client/json/JsonpUtils.java
Xtansia: Have switched to a more sensible example. [`ExtendedBounds` is generic and uses `JsonpUtils.serialize`](https://github.com/opensearch-project/opensearch-java/blob/main/java-client/src/generated/java/org/opensearch/client/opensearch/_types/aggregations/ExtendedBounds.java#L119), [Which](https://github.com/opensearch-project/opensearch-java/blob/main/java-client/src/main/java/org/opensearch/client/json/JsonpUtils.java#L152) directly passes through to `mapper.serialize` if no explicit serializer is passed or type isn't `JsonpSerializable`. Which is the same mechanism that `IndexRequest` uses, just `ExtendedBounds<Double>` is an actively in-use type within the API types.

I don't think it's unreasonable to directly handle the primitives that map directly to `JsonGenerator` calls. The alternative is moving this logic into `JsonpUtils.serialize`.

Issue #250, opened 2022-10-29
[FEATURE] Document how to use Java 8 Time (JSR 310) objects


dbwiddis:
### Is your feature request related to a problem?

As part of extensibility work in [`opensearch-sdk-java`](https://github.com/opensearch-project/opensearch-sdk-java) we are leveraging the OpenSearch Java Client to make REST calls to the OpenSearch cluster from extensions.  Our initial proof-of-concept feature is creating a detector as defined in the Anomaly Detection plugin, with work in [this draft PR](https://github.com/opensearch-project/anomaly-detection/pull/692).

The class being modeled, [AnomalyDetector](https://github.com/opensearch-project/anomaly-detection/blob/main/src/main/java/org/opensearch/ad/model/AnomalyDetector.java), includes classes from the [JSR-310 Time API](https://jcp.org/aboutJava/communityprocess/pfd/jsr310/JSR-310-guide.html), introduced with JDK-8, namely `Instant` and `Duration`.

The Java Client depends on Jackson's [ObjectMapper](https://fasterxml.github.io/jackson-databind/javadoc/2.7/com/fasterxml/jackson/databind/ObjectMapper.html) class.  Because Jackson 2.x versions still support Java 7 (and earlier) time functions, they do not include the JSR-310 API by default, and Java 8 support [must be added by one or more external modules](https://github.com/FasterXML/jackson-modules-java8).  Without this module included, attempts to serialize classes which use the JSR 310 API fail with this message:

> com.fasterxml.jackson.databind.exc.InvalidDefinitionException: 
> Java 8 date/time type `java.time.Instant` not supported by default: add Module \"com.fasterxml.jackson.datatype:jackson-datatype-jsr310\" to enable 

### What solution would you like?

1. Add a dependency to the specified module in the above exception message, e.g., 

```java
implementation "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.13.4"
```

2. Add the module in the instantiation of ObjectMapper in the JacksonJsonpMapper constructor here, by adding an additional call to `.registerModule(new JavaTimeModule())`:
https://github.com/opensearch-project/opensearch-java/blob/0ef1246d716482209b3cbb175b675c4dc16e86ae/java-client/src/main/java/org/opensearch/client/json/jackson/JacksonJsonpMapper.java#L60-L65

[...]

PR #1614, opened 2025-06-10
enable jackson `ObjectMapper` JDK8 & JSR310 modules + module autodetect

rursprung:
### Description
see the individual commit messages for further details.

short summary: this makes using core java types introduced 10+ years ago possible without having to resort to manual workarounds on the client side.

with jackson 3.x this will work out of the box since they've included the modules there. jackson 3.0 will probably be released soon, since RC5 is already out; though how/when `opensearch-java` will upgrade to this is unclear to me since it's part of the public API (and you just did a major release).

### Issues Resolved
n/a

though this relates to #250 & #251

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added support for Index Management plugin APIs ([#1604](https://github.com/opensearch-project/opensearch-java/pull/1604))
 - Added support for the security plugin APIs ([#1601](https://github.com/opensearch-project/opensearch-java/pull/1601))
 - Jackson `ObjectMapper` modules are now being auto-detected ([#1614](https://github.com/opensearch-project/opensearch-java/pull/1614))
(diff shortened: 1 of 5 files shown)


reta:
> if - after re-considering my arguments 😉 - you're still completely against adding the JDK8 & JSR310 modules as dependencies: would you at the very least accept adding the module autodetection to `opensearch-java`? then it'd be enough for consumers to just put them on the classpath without having to manually instantiate it.

Sorry for delay @rursprung , I think adding `findAndRegisterModules` would not harm (and users do have a choice to overrule this behavior by providing custom `ObjectMapper` instance). @Xtansia what do you think? Thank you.

rursprung:
i think with #1948 this is no longer that relevant since Jackson 3.x includes JDK8 & JSR310 support directly, there's no further need for external modules when using Jackson 3.x

so going forward consumers needing this have two options:
* use Jackson 2.x with the autodetect (added in 2.26.0) and add the modules to the classpath
* use Jackson 3.x (not yet possible since that PR just got merged and hasn't been released yet)

CC @reta: i guess i can go ahead and close this PR without merging it, do you agree? (you added Jackson 3.x support)

reta:
> CC @reta: i guess i can go ahead and close this PR without merging it, do you agree? (you added Jackson 3.x support)

I agree, thanks @rursprung

Issue #1581, opened 2025-05-28
Connect timeout is not triggering as expected when set in ConnectionConfig


001mertiya:
### What is the bug?
The connect timeout in the ConnectionConfig is not working as expected.

### How can one reproduce the bug?
Try creating the client by setting the connect timeout in ConnectionConfig as shown in the code snippet below, and use a very low value for the timeout (e.g., 1ms). Despite this, the connection does not fail as expected. However, if the connect timeout is set in RequestConfig, it works correctly.
```

ApacheHttpClient5TransportBuilder httpClientTransportBuilder = ApacheHttpClient5TransportBuilder.builder(hosts)
                .setHttpClientConfigCallback(httpClientBuilder -> {
                    BasicCredentialsProvider credentialsProvider = new BasicCredentialsProvider();
                    for (HttpHost host : hosts) {
                        credentialsProvider.setCredentials(new AuthScope(host),
                                new UsernamePasswordCredentials(openSearchClientConfigurations.getClusterUsername(), openSearchClientConfigurations.getClusterPassword().toCharArray()));
                    }

                    ConnectionConfig connectionConfig = ConnectionConfig.custom()
                            **.setConnectTimeout(Timeout.of(openSearchClientConfigurations.getConnectTimeout()))**
                            .setSocketTimeout(Timeout.of(openSearchClientConfigurations.getSocketTimeout()))
                            .setTimeToLive(TimeValue.of(openSearchClientConfigurations.getConnectionKeepAlive()))
                            .build();

                    PoolingAsyncClientConnectionManager connectionManager = PoolingAsyncClientConnectionManagerBuilder
                            .create()
                            .setMaxConnPerRoute(openSearchClientConfigurations.getMaxConnectionPerRoute())
                            .setMaxConnTotal(openSearchClientConfigurations.getMaxConnectionTotal())
                            .setDefaultConnectionConfig(**connectionConfig**)
                            .build();
                    return httpClientBuilder.setDefaultCredentialsProvider(credentialsProvider).setConnectionManager(**connectionManager**);
                })
                .setRequestConfigCallback(requestConfigBuilder -> requestConfigBuilder                    .setConnectionRequestTimeout(Timeout.of(openSearchClientConfigurations.getConnectionRequestTimeout())))
        }
        return httpClientTransportBuilder;

```
### What is the expected behavior?
The expected behavior is to throw a timeout exception when the specified connection timeout duration is exceeded.

### Do you have any additional context?
[...]

PR #1633, opened 2025-06-19
Fix default connectTimeout of transportBuilder

playLeo:
### Description
ConnectionConfig connectTimeout is used only when the connectTimeout in RequestConfig is null.
However, the builder is currently setting a value for connectTimeout in RequestConfig, which causes the value in ConnectionConfig to be ignored.

Therefore, it is safe to remove the default setting in RequestConfig.
Also, ConnectionConfig already defines a default value:
`Timeout DEFAULT_CONNECT_TIMEOUT = Timeout.ofMinutes(3)`

### Issues Resolved
This PR is related to [#1581](https://github.com/opensearch-project/opensearch-java/issues/1581)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Fixed
 - Fixed building instances of `Explanation` by making `details` optional ([#1620](https://github.com/opensearch-project/opensearch-java/pull/1620))
+- Fixed connectTimeout option of `ApacheHttpClient5TransportBuilder` ([#1633](https://github.com/opensearch-project/opensearch-java/pull/1633))
 
 ### Security
 
@@ -68,7 +69,6 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Fixed
 - Fixed ScoreCombination's `parameters` structure ([#1594](https://github.com/opensearch-project/opensearch-java/pull/1594))
-- Fixed default connectTimeout of transportBuilder ([#1633](https://github.com/opensearch-project/opensearch-java/pull/1633))
 
 ## [3.0.0] - 05/16/2025
 ### ⚠️ Breaking Changes ⚠️


001mertiya:
@reta  Why is there a discrepancy in the default connect timeout? The default value is 1 second in [ApacheHttpClient5TransportBuilder](https://github.com/opensearch-project/opensearch-java/blob/main/java-client/src/main/java/org/opensearch/client/transport/httpclient5/ApacheHttpClient5TransportBuilder.java#L49), but it’s 3 minutes by default in ConnectionConfig.

java-client/src/test/java/org/opensearch/client/transport/httpclient5/ApacheHttpClient5TransportBuilderTest.java
```diff
@@ -0,0 +1,46 @@
...
+        ApacheHttpClient5Transport apacheHttpClient5Transport = ApacheHttpClient5TransportBuilder.builder(new HttpHost("10.255.255.1", 9200))
+            .setHttpClientConfigCallback(httpClientBuilder -> {
+
+                ConnectionConfig connectionConfig = ConnectionConfig.custom()
+                    .setConnectTimeout(Timeout.ofMilliseconds(expectTime))
+                    .build();
+
+                PoolingAsyncClientConnectionManager connectionManager = PoolingAsyncClientConnectionManagerBuilder
+                    .create()
+                    .setDefaultConnectionConfig(connectionConfig)
```
reta, line 34: @001mertiya could you clarify please what exactly you are trying to achieve here? The `RequestConfig` and `ConnectionConfig` are different configuration settings. If you are not satisfied with the `RequestConfig`settings, use `requestConfigCallback` to set them to any value, including `null`

reta (changes requested):

java-client/src/test/java/org/opensearch/client/transport/httpclient5/ApacheHttpClient5TransportBuilderTest.java
playLeo: @reta 
The connectTimeout setting has been moved from `RequestConfig` to `ConnectionConfig`, and since the setting in `RequestConfig` has been deprecated, I think it should be removed. Also, the purpose of this test was to verify that the connectTimeout works correctly when a connection is established using the ConnectionConfig setting.

playLeo: here http5client commit [link](https://github.com/apache/httpcomponents-client/commit/f36637dc2f2905e52df76b3e95e71a9732a06982)

playLeo: Thank you for the detailed explanation. I understand. I have a question before I start working. It seems that you can configure `RequestConfig` and `ConnectionConfig` either by using `HttpClientConfigCallback` or by using `RequestConfigCallback` and `ConnectionConfigCallback`. Doesn’t having both ways to set these configurations create potential confusion? What is the benefit of using `ConnectionConfigCallback`?

playLeo: > we could (and probably should) reflect that by introducing `ConnectionConfigCallback` to allow configuring these timeouts.

I think I did not explain it clearly. If we create a `ConnectionConfigCallback` we would then have two ways to set the `ConnectionConfig` one through the existing `HttpClientConfigCallback` and one through the new `ConnectionConfigCallback` so I am wondering if the ConnectionConfigCallback is really necessary.

I also know that for `RequestConfig` there are two ways to configure it using `HttpClientConfigCallback` and `RequestConfigCallback` but I am wondering if the `RequestConfigCallback` is really needed in that case too.

reta: > If there are two ways to configure this, wouldn’t that be confusing? Are you saying it’s fine as long as it’s documented properly?

Yeah, so the `ConnectionConfigCallback` is only useful when the caller does not want to deal with connection manager (since it is part of it). In this case, providing just `ConnectionConfigCallback` and not `HttpClientConfigCallback` would work just fine. Thanks.

playLeo: Please take a look at the updated

001mertiya:
Hi @playLeo, can you pull the latest changes from main and add these test on top of that? There was an issue with socketTimeout as well, fixed in the latest main.

PR #1643, opened 2025-06-23
jackson `ObjectMapper`: auto-detect modules

rursprung:
jackson 2.x still supports java 7 and thus does not automatically support java 8 classes. this will change with jackson 3.x (see e.g. this [PR adding JSR310 date support] to 3.x).

PR #251 added documentation on how a custom `ObjectMapper` can be registered - but there's no reason why we can't just enable auto-detection centrally. the docs mention that performance should be considered, but this is IMHO not relevant here since we only construct it once and then keep the same `ObjectMapper` (which in turn the docs mention as best practices).

Signed-off-by: Ralph Ursprung <Ralph.Ursprung@avaloq.com>

[PR adding JSR310 date support]: https://github.com/FasterXML/jackson-databind/pull/5032

this is split from #1614. CC @reta & @Xtansia 

### Description

### Issues Resolved

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -99,11 +99,12 @@ OpenSearchTransport transport = new RestClientTransport(restClient, new JacksonJ
 OpenSearchClient client = new OpenSearchClient(transport);
 ```
 
-The `JacksonJsonpMapper` class (2.x versions) only supports Java 7 objects by default. [Java 8 modules](https://github.com/FasterXML/jackson-modules-java8) to support JDK8 classes such as the Date and Time API (JSR-310), `Optional`, and more can be used by including [the additional datatype dependency](https://github.com/FasterXML/jackson-modules-java8#usage) and adding the module.  For example, to include JSR-310 classes:
+The `JacksonJsonpMapper` class (2.x versions) only supports Java 7 objects by default. [Java 8 modules](https://github.com/FasterXML/jackson-modules-java8) to support JDK8 classes such as the Date and Time API (JSR-310), `Optional`, and more can be used by including [the additional datatype dependency](https://github.com/FasterXML/jackson-modules-java8#usage) and adding the module. Auto-detection for these modules is enabled, adding them to the classpath is enough.
+You can also provide your own `ObjectMapper` instance if you need to pre-configure it differently, for example:
 
 ```java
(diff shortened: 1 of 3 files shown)


USER_GUIDE.md
```diff
@@ -99,13 +99,7 @@ OpenSearchTransport transport = new RestClientTransport(restClient, new JacksonJ
 OpenSearchClient client = new OpenSearchClient(transport);
 ```
 
-The `JacksonJsonpMapper` class (2.x versions) only supports Java 7 objects by default. [Java 8 modules](https://github.com/FasterXML/jackson-modules-java8) to support JDK8 classes such as the Date and Time API (JSR-310), `Optional`, and more can be used by including [the additional datatype dependency](https://github.com/FasterXML/jackson-modules-java8#usage) and adding the module.  For example, to include JSR-310 classes:
-
-```java
```
reta, line 104: @rursprung could you please bring the example back? It demonstrates the usage of custom `ObjectMapper()`, the `registerModule` could be indeed removed, may be we could use `.configure(SerializationFeature.INDENT_OUTPUT, true)` as an example 

```
OpenSearchTransport transport = new RestClientTransport(restClient,
    new JacksonJsonpMapper(new ObjectMapper().registerModule(new JavaTimeModule()))); 
```

java-client/src/main/java/org/opensearch/client/json/jackson/JacksonJsonpMapper.java
```diff
@@ -63,7 +63,9 @@ public JacksonJsonpMapper(ObjectMapper objectMapper, JsonFactory jsonFactory) {
 
     public JacksonJsonpMapper() {
         this(
-            new ObjectMapper().configure(SerializationFeature.INDENT_OUTPUT, false).setSerializationInclusion(JsonInclude.Include.NON_NULL)
+            new ObjectMapper().configure(SerializationFeature.INDENT_OUTPUT, false)
```
reta, line 66: 👍

reta:
LGTM, one minor comment, thanks @rursprung

Xtansia (approved):
No concerns from me, thanks @rursprung

rursprung:
thanks everyone! can we backport this to 2.x as well?

PR #1644, opened 2025-06-23
Fix integer overflow issue in DocStatus by changing Integer to Long

donggyu04:
### Description

Since version 3.0.0 when calling `OpenSearchClient.nodes().stats()`, the function call fails on certain clusters with an error such as 
```
java.lang.RuntimeException: Jackson exception: Numeric value (5069719939) out of range of int (-2147483648 - 2147483647)
```
While debugging, it appears that the overflow error occurs because the `2xx` value in `doc_status` on a specific coordinator node exceeds the range of an integer.

image

To resolve this issue, the type of the `2xx` field was changed to `Long`.

### Issues Resolved

- https://github.com/opensearch-project/opensearch-java/issues/1645

### DIFF
--- a/java-client/src/generated/java/org/opensearch/client/opensearch/_types/DocStatus.java
+++ b/java-client/src/generated/java/org/opensearch/client/opensearch/_types/DocStatus.java
@@ -63,19 +63,19 @@ import org.opensearch.client.util.ToCopyableBuilder;
 public class DocStatus implements PlainJsonSerializable, ToCopyableBuilder<DocStatus.Builder, DocStatus> {
 
     @Nullable
-    private final Integer _1xx;
+    private final Long _1xx;
 
     @Nullable
-    private final Integer _2xx;
+    private final Long _2xx;
 
     @Nullable
-    private final Integer _3xx;
+    private final Long _3xx;
 
     @Nullable
-    private final Integer _4xx;
+    private final Long _4xx;
 
     @Nullable
-    private final Integer _5xx;
+    private final Long _5xx;
 
     // ---------------------------------------------------------------------------------------------
 
@@ -98,7 +98,7 @@ public class DocStatus implements PlainJsonSerializable, ToCopyableBuilder<DocSt
      * </p>
      */
     @Nullable
-    public final Integer _1xx() {
+    public final Long _1xx() {
         return this._1xx;
     }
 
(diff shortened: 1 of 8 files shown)


Xtansia (changes requested):
This should instead be corrected in the specification: https://github.com/opensearch-project/opensearch-api-specification/blob/main/spec/schemas/_common.yaml#L24-L47

It will then automatically be updated here via code generation.

java-client/src/generated/java/org/opensearch/client/opensearch/_types/DocStatus.java
```diff
@@ -66,7 +66,7 @@ public class DocStatus implements PlainJsonSerializable, ToCopyableBuilder<DocSt
     private final Integer _1xx;
 
     @Nullable
-    private final Integer _2xx;
+    private final Long _2xx;
```
reta, line 69: Nice catch, thank you @donggyu04 , could you please update all counters to be `Long` (to reflect server side APIs), thank you

donggyu04: @reta 

Hi~ I have created a [PR](https://github.com/opensearch-project/opensearch-api-specification/pull/933) to change the field type in the `opensearch-api-specification` repository :)

Xtansia:
@donggyu04 Now that the specification change has been merged, in your checked out copy of your branch please run `./gradlew :java-codegen:downloadLatestSpec :java-codegen:run` and then commit the changes. This will download the new spec and re-generate the DocStatus class with the new types.

Please also add an entry to the changelog under the [Changed heading](https://github.com/opensearch-project/opensearch-java/blob/main/CHANGELOG.md#changed).

Xtansia:
> @Xtansia
> 
> I have applied your request via [72f17ca](https://github.com/opensearch-project/opensearch-java/commit/72f17caa9dd4f957e510bb4b6a6ec4728bda176d) commit

Please commit all the changed files after running that command. Should be changes to `opensearch-openapi.yml` and a few other comment changes I believe.

donggyu04:
@Xtansia 

Okay, I fixed my PR via https://github.com/opensearch-project/opensearch-java/pull/1644/commits/85f4912427a6954828c4cd32bd1a313afe20eaca commit.

Issue #1717, opened 2025-08-23
[BUG] ISM API - Put Policies: `java.lang.RuntimeException: Jackson exception: Numeric value out of range of int (-2147483648 - 2147483647)`


byi-esri:
### What is the bug?

When attempting to make a `PutPolicyRequest`, the following exception is thrown:

```
java.lang.RuntimeException: Jackson exception: Numeric value (1755915614485) out of range of int (-2147483648 - 2147483647)
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 342]
```

However, the policy **is saved.** It appears that some numeric value, likely a timestamp, is triggering a validation error. I am currently not specifying any time values explicitly.

### How can one reproduce the bug?

I tried to use the ISM code to add a simply policy based on this example:
https://docs.opensearch.org/latest/im-plugin/ism/policies/#sample-policy-with-ism-template-for-auto-rollover

A snippet of my code is below:
```java
IsmTemplate getIsmTemplate() {
      return IsmTemplate.builder()
          .lastUpdatedTime(1)
          .indexPatterns(getIndexMatchingPattern())
          .build();
}

List<States> getStates() {
      var rollover = Action.builder()
          .rollover(ActionRollover.builder()
              .minDocCount(MAX_RECORDS)
              .build())
          .build();
      return List.of(
          States.builder()
              .name(IndexState.WRITEABLE.name)
              .actions(rollover)
              .build()
      );
}

/* ... */
var p = Policy.builder()
            .states(getStates())
          .policyId(<id>)
          .defaultState(IndexState.WRITEABLE.name)
          .ismTemplate(getIsmTemplate());

var policyReq = PutPoliciesRequest.builder()
          .policy(_p -> p)
          .policyId(r.getId())
          .build();
var policyRes = client.ism().putPolicies(policyReq); // THROWS HERE
```

### What is the expected behavior?
The expected behavior would be that `client.ism().putPolicies` should not throw.

### What is your host/environment?
Windows 11

### Do you have any screenshots?

Stacktrace of the exception showing that the call to `putPolicies` fails:

<img width="1533" height="183" alt="Image" src="https://github.com/user-attachments/assets/44e68155-2311-4ef4-bbe1-2b3f2880b7b5" />

### Do you have any additional context?

The number that is out of range seems to be a unix epoch description. Perhaps one of the fields in the request object needs to be changed to a type that can fit the number?

[...]

PR #1784, opened 2025-11-06
Fix NPE when null ObjectBuilder is accessed

moesterheld:
### Description
When an `ObjectBuilderDeserializer` is created for a property that is `null`, a NullPointerException is thrown when the deserializer tries to deserialize excpected nested fields due to the ObjectBuilder being null.
This adds a check that returns null as the property value if the builder is null.

### Issues Resolved
part of  https://github.com/opensearch-project/opensearch-java/issues/1717

### DIFF
--- a/java-client/src/main/java/org/opensearch/client/json/ObjectBuilderDeserializer.java
+++ b/java-client/src/main/java/org/opensearch/client/json/ObjectBuilderDeserializer.java
@@ -33,7 +33,6 @@
 package org.opensearch.client.json;
 
 import jakarta.json.stream.JsonParser;
-import java.util.Objects;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Supplier;


stvjbs:
Good day!
Could I ask you, if there is any plan to fix the issue by merging this PR soon?

As I understood, the only way to work with the ISM API is to use client.generic() function, am I right?

Thank you in advance for your answer :)

stvjbs:
@reta Thank you for the clarification, I see the issue. But the question is still existing :) 
 if there is any plan and term to fix that issue?
 May I contribute to api-specification project in order to have the fix faster?

reta:
> May I contribute to api-specification project in order to have the fix faster?

@stvjbs please go ahead, it would be much appreciated, thank you

moesterheld:
I added the timestamp long/int fix to the spec [here](https://github.com/opensearch-project/opensearch-api-specification/pull/1005)
I am now working on the remaining two issues

moesterheld:
Can anyone explain why Lists and Maps are always [generated](https://github.com/opensearch-project/opensearch-java/blob/897cc9bb34d05dd852b1a0b2fcd5621d9bd81628/java-codegen/src/main/resources/org/opensearch/client/codegen/templates/ObjectShape/FieldNullability.mustache#L2) as `@Nonnull` without checking if they are required properties?

In OpenApi, you can have lists (e.g. in `oneOf`) where the property is not included in `required` properties and the property can be left out in the json entirely.
In the case for this issue, `ism_template` defined [here](https://github.com/moesterheld/opensearch-api-specification/blob/79601ca618d1b0f21957e51c0fa34736e9b2eaf7/spec/schemas/ism._common.yaml#L51) is parsed correctly into a non-required shape, but due to the template, it is always set as `@Nonnull`

reta:
@moesterheld could you please backport to 3.x manually so I can merge it, thank you

Issue #1787, opened 2025-11-07
[BUG] ObjectMapper module resolution cannot fully be turned off


jamjar919:
### What is the bug?
When using the following snippet to create an OpenSearchClient:

```
return new OpenSearchClient(new AwsSdk2Transport(
            ApacheHttpClient.builder().build(),
            host,
            "es,
            region,
            AwsSdk2TransportOptions
                .builder()
                .setMapper(new JacksonJsonpMapper(new ObjectMapper()))
                .build()
        ));
```

You would expect no module resolution to be performed by Jackson, since we have manually created an `ObjectMapper` and `JacksonJsonpMapper` - however due to this line in `AwsSdk2Transport`:

```
        this.defaultMapper = Optional.ofNullable(options).map(AwsSdk2TransportOptions::mapper).orElse(new JacksonJsonpMapper());
```

A default `JacksonJsonpMapper()` is always created, which will call:

```
public JacksonJsonpMapper() {
        this(
            new ObjectMapper().configure(SerializationFeature.INDENT_OUTPUT, false)
                .setSerializationInclusion(JsonInclude.Include.NON_NULL)
                .findAndRegisterModules()
        );
    }
```

Which attempts to find and register Jackson modules. If any incompatible modules are discovered, then the initialisation of the client will fail.

It appears this was introduced during https://github.com/opensearch-project/opensearch-java/pull/1643 which introduced module resolution more generally.

### How can one reproduce the bug?
Create a client using the snippet above, and install an incompatible Jackson module. Running the code will produce an error like

```
Scala module 2.9.7 requires Jackson Databind version >= 2.9.0 and < 2.10.0
```

### What is the expected behavior?
When passing a mapper to `AwsSdk2Transport` we should only attempt to create a default mapper if one is not provided for us.  

### What is your host/environment?
Docker: `jdk21-hotspot-alma9`, version `3.3.0`

### Do you have any additional context?
This was found whilst creating a plugin to be in use by other projects - if those projects contain incompatible Jackjson modules then creating a client will always fail.

PR #1788, opened 2025-11-07
Only create the default mapper if one is not provided

jamjar919:
### Description
Changes the default mapper retrieval to an `orElseGet` in `AwsSdk2Transport`, which means if a mapper is supplied in the options, then one will not be created for the user unnecessarily. This means that modules won't be resolved at all if the user  chooses to pass in a custom mapper.

Signed-off-by: James Paterson <james@mjsquare.com>

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-java/issues/1787


andrross (approved):

reta (approved):

reta:
Thank you @jamjar919 !

Issue #1877, opened 2026-02-13
[BUG]  ShardFailure deserialization fails when OpenSearch omits the primary field in error responses


janhoy:
### What is the bug?                                      

This is a regression in version 3.5.0, introduced by [this commit](https://github.com/opensearch-project/opensearch-java/commit/27c8dc8c0da6689a493ac1d108d02b0a0b2dbc38), after fetching latest api-spec and doing code-gen.

In opensearch-java v3.4.0 this worked, but from v3.5.0, the client throws a `MissingRequiredPropertyException` when attempting to deserialize error responses that contain shard failures, since `primary` field is now required in the updated api-spec. This occurs during search requests when OpenSearch returns shard-level failures in the error response.

**Error:**
org.opensearch.client.util.MissingRequiredPropertyException: Missing required property 'ShardFailure.primary'

### How can one reproduce the bug?

1. Execute a search query that triggers a shard failure (e.g., malformed query, index with shard issues)
2. OpenSearch returns an error response containing `failed_shards` in the error metadata
3. The Java client attempts to deserialize the `ShardFailure` objects
4. Deserialization fails because the `primary` field is marked as required but OpenSearch omits it

**Example error response structure:**
```json
{
  "error": {
    "type": "search_phase_execution_exception",
    "reason": "all shards failed",
    "failed_shards": [
      {
        "shard": 0,
        "index": "my-index",
        "node": "node-id",
        "reason": {
          "type": "query_shard_exception",
          "reason": "Parse failure"
        }
      }
    ]
  }
}
```

Note: The `primary` field is missing from the failed_shards objects.

### What is the expected behavior?

The client should successfully deserialize error responses containing shard failures, regardless of whether the primary field is present or not. The primary field should be treated as optional since OpenSearch doesn't consistently include it in all error responses.

### What is your host/environment?

- OpenSearch Java Client version: 3.5

### Do you have any screenshots?

No

### Do you have any additional context?

Root Cause:
The ShardFailure class (generated from the [OpenAPI spec](https://github.com/opensearch-project/opensearch-api-specification/pull/981)) marks the primary field as required and uses a primitive boolean type, which cannot be null. However, OpenSearch's actual error responses often omit this field, causing a mismatch between the spec and real-world behavior.

Affected Code:
- java-client/src/generated/java/org/opensearch/client/opensearch/_types/ShardFailure.java

### Proposed Fix:
- 3.x: Make the `primary` attribute not required
[...]

PR #1894, opened 2026-02-22
[Backport] [3.x] Update Specification and Re-generate Code (#1890)

reta:
Backport of https://github.com/opensearch-project/opensearch-java/pull/1890 to 3.x


java-client/src/generated/java/org/opensearch/client/opensearch/_types/ShardStatistics.java
```diff
@@ -64,7 +64,7 @@ public class ShardStatistics implements PlainJsonSerializable, ToCopyableBuilder
     private final int failed;
 
     @Nonnull
-    private final List<ShardFailure> failures;
+    private final List<ShardSearchFailure> failures;
```
Xtansia: This is technically a breaking change for 3.x

reta: Thanks @Xtansia . Yes, it is, sadly, we have discussed it over here https://github.com/opensearch-project/opensearch-java/issues/1877#issuecomment-3917483841, I honestly unsure what would be the best option here:
 - revert ShardFailure changes for `primary` etc 
 - or adsorb this breaking change (since ShardSearchFailure will never have `primary` populated, it is never returned by the server)

Xtansia (approved):

PR #1977, opened 2026-05-09
fix(ism): change lastUpdatedTime from Integer to Long to prevent epoc…

### LINKED ISSUE #1898: [BUG] ISM types lastUpdatedTime is susceptible to int overflow
### What is the bug?

`Policy` and `IsmTemplate` have fields called `lastUpdatedTime`, stored as an `Integer`.
These fields get epoch millis as their values, which are larger than `MAX_INTEGER`.

https://github.com/opensearch-project/opensearch-java/blob/a903ddfde2c3530c61377dd6a8e17cc2a48358cf/java-client/src/generated/java/org/opensearch/client/opensearch/ism/Policy.java#L159

https://github.com/opensearch-project/opensearch-java/blob/a903ddfde2c3530c61377dd6a8e17cc2a48358cf/java-client/src/generated/java/org/opensearch/client/opensearch/ism/IsmTemplate.java#L103

Use `OpenSearchIsmClient` and try to `GET` a `policy`.

Successful GET, successful deserialization.

Windows 11, `opensearch-java` 3.6.0

N/A

If using a `JacksonJsonpMapper`, then I get an exception like this:
```
Caused by: com.fasterxml.jackson.core.exc.InputCoercionException: Numeric value (1772134315305) out of range of int (-2147483648 - 2147483647)
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 253]
```

Using Jackson config like `DeserializationFeature.USE_BIG_INTEGER_FOR_INTS` does not help, because the target type (eg `Policy`) explicitly has an an `Integer` type.

If using a `JsonbJsonpMapper`, then the value silently overflows and I get a negative timestamp.

Zahanturel:
### What does this change achieve?

`last_updated_time` in `Policy` and `IsmTemplate` carries epoch-millisecond timestamps. These values (e.g. `1700000000000`) exceed `Integer.MAX_VALUE` (2147483647), causing `InputCoercionException` with `JacksonJsonpMapper` and silent negative-value corruption with `JsonbJsonpMapper`.

Changes `format: int64` in the OpenAPI spec (the code generation source of truth) and updates the generated Java models accordingly. Verified against `TypeMapper.java` and the Mustache templates that the generated output is identical to what `./gradlew :java-codegen:run` would produce.

### Issues Resolved

Fixes #1898.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Detect AWS SDK `Apache5HttpClient` in `AwsSdk2Transport` body-method guardrail ([#1903](https://github.com/opensearch-project/opensearch-java/pull/1970))
 
 ### Fixed
+- Fix ISM `lastUpdatedTime` field type from `Integer` to `Long` to prevent epoch-millisecond overflow ([#1977](https://github.com/opensearch-project/opensearch-java/pull/1977))
 
 ## [Unreleased 3.x]
 ### Added
--- a/java-client/src/generated/java/org/opensearch/client/opensearch/ism/IsmTemplate.java
+++ b/java-client/src/generated/java/org/opensearch/client/opensearch/ism/IsmTemplate.java
@@ -65,7 +65,7 @@ public class IsmTemplate implements PlainJsonSerializable, ToCopyableBuilder<Ism
     private final List<String> indexPatterns;
 
     @Nullable
-    private final Integer lastUpdatedTime;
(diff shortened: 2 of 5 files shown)


reta (changes requested):
@Zahanturel thanks for the change, these files are generated, please submit the change to specification instead - https://github.com/opensearch-project/openSearch-api-specification

Zahanturel:
Thanks for the redirect. Will submit the fix to the API specification repo instead at opensearch-project/opensearch-api-specification.

PR #1948, opened 2026-04-05
[FEATURE] Support Jackson 3.x release line

reta:
### Description
Support Jackson 3.x release line,  aligning with https://github.com/opensearch-project/OpenSearch/pull/21029
Making Jackson 3.x default for `main` (4.x), the 3.x will keep 2.x as a default.

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-java/issues/1810

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1982, opened 2026-05-28
[Backport] [3.x] [FEATURE] Support Jackson 3.x release line (#1948)

reta:
Backport of https://github.com/opensearch-project/opensearch-java/pull/1948 to `3.x`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ## [Unreleased 3.x]
 ### Added
 - Detect AWS SDK `Apache5HttpClient` in `AwsSdk2Transport` body-method guardrail ([#1903](https://github.com/opensearch-project/opensearch-java/pull/1970))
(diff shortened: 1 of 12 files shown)


java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
```diff
@@ -41,7 +41,7 @@
 import javax.net.ssl.SSLHandshakeException;
 import org.opensearch.client.json.JsonpDeserializer;
 import org.opensearch.client.json.JsonpMapper;
-import org.opensearch.client.json.jackson.JacksonJsonpMapper;
+import org.opensearch.client.json.jackson3.JacksonJsonpMapper;
```
Xtansia: We shouldn't change the fallback here on 3.x

Xtansia (changes requested):

java-client/src/main/java/org/opensearch/client/transport/aws/AwsSdk2Transport.java
reta: Yeah, slipped somehow, fixed now, thanks!

Xtansia (approved):

PR #2002, opened 2026-06-12
Support sort _score/_doc order shorthand

russcam:
### Description
This PR adds support for the shorthand version of `_score` and `_doc` sort order of the form:

```json
{"_score": "desc"}
```

This currently throws a JSON parsing exception, but is a valid form. This form may be generated from another system and deserialized using the java client.

### Issues Resolved


reta (changes requested):
@russcam we would really need to stick to API specification (or update API specification with the shortcut form first), thank you

[1] https://github.com/opensearch-project/opensearch-api-specification/blob/main/spec/schemas/_common.yaml#L669

PR #2033, opened 2026-06-30
Use Testcontainers for integration tests

### LINKED ISSUE #1916: [FEATURE] use testcontainers for integration tests
when developing on this project you have to manually start an OpenSearch instance to run the integration tests against.

use `testcontainers` (most likely through [`opensearch-testcontainers`](https://github.com/opensearch-project/opensearch-testcontainers)) to run the containers automatically in the integration tests, so that everything is fully encapsulated and the OS instance has a well-known state for the tests.

n/a

other projects like [`spring-data-opensearch`](https://github.com/opensearch-project/spring-data-opensearch) have been using this for a long time.

lsh1215:
## Description

Adds Testcontainers-backed OpenSearch startup for Java client integration tests so developers and the released integration workflow no longer need to start OpenSearch with Docker Compose before running `integrationTest`.

The Testcontainers path is enabled by default for the `integrationTest` task. Existing externally managed clusters remain supported by setting `tests.rest.cluster` or disabling the test container with `-Dtests.opensearch.testcontainers.enabled=false`.

The Docker/Testcontainers startup path is limited to the `integrationTest` task. The released integration CI path remains Ubuntu-only, and the macOS/Windows build and unit-test jobs do not get a Docker runtime requirement.

Closes #1916.

## Changes

- Add a JVM-singleton `OpenSearchTestContainer` helper for integration tests.
- Start the container before `OpenSearchRestTestCase` initializes its REST client.
- Use HTTPS/basic auth and handle the OpenSearch 2.12+ initial admin password requirement.
- Handle ambiguous OpenSearch 2.x/custom image version tags conservatively and cover that compatibility logic with focused tests.
- Disable disk threshold checks only for the disposable Testcontainers-managed node to avoid Docker host disk watermark state applying `cluster.blocks.create_index=true` during tests.
- Replace released integration workflow Docker Compose setup with Gradle/Testcontainers startup.
- Keep unreleased/source-build integration workflow on its existing external cluster path by disabling Testcontainers there.
- Document the new default, version/image selection, and external-cluster opt-out.

## Verification

- `git diff --check`
- `./gradlew integrationTest --tests "org.opensearch.client.opensearch.integTest.OpenSearchTestContainerTests"`
- `./gradlew spotlessJavaCheck`
- `./gradlew integrationTest --tests "org.opensearch.client.opensearch.integTest.restclient.PingAndInfoIT" -Dtests.opensearch.version=3.2.0`
- GitHub Actions for PR #2033: all checks passed, including the released integration matrix.


java-client/build.gradle.kts
```diff
@@ -387,6 +398,7 @@ if (runtimeJavaVersion >= JavaVersion.VERSION_21) {
     testImplementation("org.opensearch.test", "framework", opensearchVersion) {
       exclude(group = "org.hamcrest")
     }
+    "java21Implementation"("org.testcontainers:testcontainers:2.0.5")
```
reta: Please let's use official https://github.com/opensearch-project/opensearch-testcontainers

reta (changes requested):

java-client/build.gradle.kts
lsh1215: Sorry, I missed the official module in the first pass. Thanks for pointing it out.

Updated in 728e3fe1b.

The integration test container setup now uses org.opensearch.testcontainers.OpenSearchContainer, and the local GenericContainer setup code has been removed.

The remaining Testcontainers dependency is compileOnly only, since OpenSearchContainer extends GenericContainer and javac needs that superclass API.

The current CI failures appear unrelated to the test changes; the failed jobs stopped during Gradle buildSrc dependency resolution.

java-client/build.gradle.kts
```diff
@@ -387,6 +398,12 @@ if (runtimeJavaVersion >= JavaVersion.VERSION_21) {
     testImplementation("org.opensearch.test", "framework", opensearchVersion) {
       exclude(group = "org.hamcrest")
     }
+    // opensearch-testcontainers exposes OpenSearchContainer as a GenericContainer subclass, but publishes
```
reta: Please just use:
```suggestion
     testImplementation("org.opensearch:opensearch-testcontainers:4.1.0")
     testImplementation"("org.testcontainers:testcontainers:2.0.4")
     testImplementation"("org.testcontainers:testcontainers-junit-jupiter:2.0.4")
```

java-client/src/test/java11/org/opensearch/client/opensearch/integTest/OpenSearchTestContainerTests.java
```diff
@@ -0,0 +1,41 @@
...
+ * compatible open source license.
+ */
+
+package org.opensearch.client.opensearch.integTest;
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
+
+public class OpenSearchTestContainerTests {
```
reta: I don't think we need these tests - opensearch-testcontainers has bunch of them

java-client/src/test/java11/org/opensearch/client/opensearch/integTest/OpenSearchTestContainer.java
```diff
@@ -0,0 +1,79 @@
...
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.client.opensearch.integTest;
+
+import org.opensearch.testcontainers.OpenSearchContainer;
+
+final class OpenSearchTestContainer {
```
reta: Please move under ` java-client/src/test/java21`. Also, we should be using JUnit Jupiter extentions mechanism with proper pre / post lifycycle (see please examples here https://www.baeldung.com/junit-5-registerextension-annotation)

java-client/src/test/java11/org/opensearch/client/opensearch/integTest/TestcontainersThreadFilter.java
```diff
@@ -0,0 +1,20 @@
...
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.client.opensearch.integTest;
+
+import com.carrotsearch.randomizedtesting.ThreadFilter;
+
+public final class TestcontainersThreadFilter implements ThreadFilter {
```
reta, line 13: Please remove, we should not see any leaking threads after https://github.com/opensearch-project/opensearch-java/pull/2033/changes#r3514648791 (the extension part)

lsh1215: Removing the filter surfaces thread leaks that RandomizedTesting flags at suite scope. These threads are JVM-global statics in the Testcontainers stack (duct-tape's `Timeouts` executor, the ryuk reaper, and the `pull-watchdog/wait pools`), so `container.stop()` doesn't end them. 
open issue: https://github.com/testcontainers/testcontainers-java/issues/9227

OpenSearch core keeps a `@ThreadLeakFilters` for the same threads: https://github.com/opensearch-project/OpenSearch/blob/main/plugins/ingestion-kafka/src/internalClusterTest/java/org/opensearch/plugin/kafka/TestContainerThreadLeakFilter.java

Can we keep a small filter like that instead? The alternatives (disabling ryuk, or reflecting into duct-tape's executor) are heavier and still miss `pull-watchdog/wait`.

java-client/src/test/java11/org/opensearch/client/opensearch/integTest/OpenSearchTestContainer.java
lsh1215: Moved the helper under `src/test/java21`. On the Jupiter extension part: these tests extend `OpenSearchRestTestCase`, so they run under `RandomizedRunner` (JUnit 4) and the Test task doesn't call useJUnitPlatform(). Jupiter's `@RegisterExtension / BeforeAllCallback` only run under the Jupiter engine, so they won't fire here.

The JUnit 4 equivalent is a `@ClassRule ExternalResource`, which the `opensearch-testcontainers` README documents: https://github.com/opensearch-project/opensearch-testcontainers#junit-4-integration

Would that work, or do you want to migrate these suites to the JUnit Platform (a larger change to the RandomizedRunner base classes)?

reta: Sorry, totally my bad,  indeed we still use JUnit 4,  `@ClassRule` would be the right equivalent (not extension), thank you

java-client/src/test/java11/org/opensearch/client/opensearch/integTest/TestcontainersThreadFilter.java
reta: > Can we keep a small filter like that instead? The alternatives (disabling ryuk, or reflecting into duct-tape's executor) are heavier and still miss pull-watchdog/wait.

I see, yeah, let's keep it for now, thank you for clarification

lsh1215:
Hi @reta, thanks for the review. Pushed 65bb540 with the changes we discussed:

- moved the helper to `src/test/java21` and wired it as a JUnit 4 `@ClassRule(ExternalResource)` that starts one container for the whole test JVM
- kept `TestcontainersThreadFilter` / `@ThreadLeakFilters`
- removed `OpenSearchTestContainerTests`

Two dependency notes:
- dropped `bootstrap.memory_lock=true` from the container env (Testcontainers doesn't set a matching memlock ulimit, so it was more likely to trip the bootstrap check than help; `disk.threshold_enabled=false` is still there)
- pinned `testcontainers` to 2.0.4 directly instead of the 2.0.3 that `opensearch-testcontainers:4.1.0` pulls in transitively (within its supported range, but happy to match 2.0.3 if you'd rather)

The failing CI checks don't look related to this change.

java-client/src/test/java21/org/opensearch/client/opensearch/integTest/OpenSearchTestContainerRule.java
```diff
@@ -0,0 +1,93 @@
...
+import org.opensearch.testcontainers.OpenSearchContainer;
+import org.opensearch.testcontainers.OpenSearchDockerImage;
+
+/**
+ * Starts one OpenSearch test container for the whole test JVM and points the integration tests at
+ * it through system properties. The container is shared by every test class, so it is left running
+ * after each class; Testcontainers' Ryuk reaper removes it once the JVM exits.
+ */
+final class OpenSearchTestContainerRule extends ExternalResource {
+    static final String ENABLED_PROPERTY = "tests.opensearch.testcontainers.enabled";
```
reta: I think we could make those just `private` as well?

java-client/src/test/java21/org/opensearch/client/opensearch/integTest/OpenSearchTestContainerRule.java
```diff
@@ -0,0 +1,93 @@
...
+    private static final String DEFAULT_ADMIN_PASSWORD = "admin";
+
+    private static OpenSearchContainer<?> container;
+
+    @Override
+    protected void before() {
+        startIfNeeded();
+    }
+
+    static synchronized void startIfNeeded() {
```
reta: I don't think we need synchronized - the JUnit will not initialize same class rule concurrently

reta:
> Hi @reta, thanks for the review. Pushed https://github.com/opensearch-project/opensearch-java/commit/65bb540678214d9d47c4028e87568cd3c59dce29 with the changes we discussed:

Thanks a lot @lsh1215 , it looks great, I have few minor nits but LGTM otherwise.

> The failing CI checks don't look related to this change.

Fixed that here https://github.com/opensearch-project/opensearch-java/pull/2042, we would need to rebase / merge right after sadly.

lsh1215:
Thanks! Done in d460616. Constants are private now and the synchronized is gone. Will rebase on #2042 once it's merged.

reta (approved):

reta:
> The backport to `3.x` failed. Please backport manually. See failed workflow run: https://github.com/opensearch-project/opensearch-java/actions/runs/28753546085

@lsh1215 may I ask you please to send a pull request against 3.x branch, thank you

PR #2034, opened 2026-06-30
Fix Apache HttpClient 5 reactor recovery

### LINKED ISSUE #1969: [BUG]ApacheHttpClient5Transport enters unrecoverable state after IOReactorShutdownException
**Description**
We are encountering a persistent IOReactorShutdownException ("I/O reactor has been shut down") when using ApacheHttpClient5Transport.
 
Once the transport enters this state—typically triggered under high load or following an OOM event—all subsequent requests fail consistently. The client does not recover automatically, and the only workaround is to restart the process or manually recreate the client/transport.
 
This effectively puts the application into a permanently degraded state.
 
**Context & Comparison**
This behavior appears similar to an issue previously reported and addressed in the Elasticsearch Java client:
 
**Issue**: [elastic/elasticsearch-java#1046](https://github.com/elastic/elasticsearch-java/issues/1046)
**Fix/Mitigation**: [elastic/elasticsearch-java#1049](https://github.com/elastic/elasticsearch-java/pull/1049)
 
The fix there ensures the transport can recover instead of remaining in a failed state.
 
**Expected Behavior**
The OpenSearch Java Client should provide similar resilience, allowing the transport layer to recover from IOReactorShutdownException without requiring a full application restart or manual client reinitialization.
 
**Current Environment**
 
Library: [OpenSearch Java Client](https://github.com/opensearch-project/opensearch-java)
Version: 3.4.0
Transport: ApacheHttpClient5Transport 5.5.2

sharp-pixel:
Fixes #1969

## Problem

`ApacheHttpClient5Transport` could enter an unrecoverable state after Apache HttpClient 5 reported `IOReactorShutdownException`. Once the underlying I/O reactor was shut down, the transport continued using the same dead `CloseableHttpAsyncClient`, so every later request failed until the application recreated the transport.

One common trigger is memory pressure while buffering large or highly concurrent responses. The existing per-response buffer limit capped a single response, but it did not cap aggregate heap usage across concurrent responses, so many in-flight responses could still exhaust the heap and kill the reactor.

## Solution

This change makes `ApacheHttpClient5Transport` recover from reactor shutdowns when it owns the client lifecycle:

- Rebuilds and starts a fresh Apache HC5 async client when `IOReactorShutdownException` is detected.
- Handles both synchronous `execute(...)` failures and callback-delivered reactor shutdown failures.
- Retries the affected request once on the rebuilt client.
- Adds rebuild backoff to avoid rebuild storms if the reactor keeps dying.
- Prevents recovery after the transport has been explicitly closed.
- Keeps externally supplied clients unchanged; recovery is disabled when the transport does not own the client lifecycle.

It also reduces the chance of reactor-killing memory failures:

- Adds an optional aggregate response-buffer memory budget.
- Adds bounded response-consumer capacity increments when a budget is active.
- Converts `OutOfMemoryError` during response buffering into a request-level `IOException` so the reactor thread is not killed.
- Fails fast without node denylisting/retry for deterministic client-side failures such as shared-budget exhaustion, oversized single responses, and OOM-caused buffering failures.
- Resets leaked response-buffer budget reservations when the client is rebuilt, including request-specific budgets.

## Testing

Ran:

```bash
./gradlew :java-client:test \
  --tests org.opensearch.client.transport.httpclient5.ApacheHttpClient5TransportRecoveryTest \
  --tests org.opensearch.client.transport.httpclient5.internal.ResponseMemoryBudgetTest \
  --tests org.opensearch.client.transport.httpclient5.internal.HeapBufferedAsyncEntityConsumerTest

./gradlew :java-client:test \
  --tests 'org.opensearch.client.transport.httpclient5.*' \
  --tests 'org.opensearch.client.transport.httpclient5.internal.*'

./gradlew :java-client:spotlessJavaCheck

git diff --check
```

Also verified the locally published modified client in a standalone sample app using the normal existing transport builder path against OpenSearch 3.7.0 in Docker:

- ping and info/version
- create index
- index and search a document
- concurrent indexing/search smoke test with one client instance

### DIFF
--- a/java-client/src/main/java/org/opensearch/client/transport/httpclient5/ApacheHttpClient5Transport.java
+++ b/java-client/src/main/java/org/opensearch/client/transport/httpclient5/ApacheHttpClient5Transport.java
@@ -41,7 +41,12 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.LongSupplier;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.zip.GZIPOutputStream;
 import javax.annotation.Nullable;
@@ -78,7 +83,9 @@ import org.apache.hc.core5.http.message.RequestLine;
 import org.apache.hc.core5.http.message.StatusLine;
 import org.apache.hc.core5.http.nio.AsyncRequestProducer;
 import org.apache.hc.core5.http.nio.AsyncResponseConsumer;
+import org.apache.hc.core5.io.CloseMode;
 import org.apache.hc.core5.net.URIBuilder;
+import org.apache.hc.core5.reactor.IOReactorShutdownException;
 import org.apache.hc.core5.util.Args;
 import org.opensearch.client.json.JsonpDeserializer;
 import org.opensearch.client.json.JsonpMapper;
@@ -107,8 +114,29 @@ public class ApacheHttpClient5Transport implements OpenSearchTransport {
     private static final Log logger = LogFactory.getLog(ApacheHttpClient5Transport.class);
     static final ContentType JsonContentType = ContentType.APPLICATION_JSON;
 
+    /**
+     * Default minimum delay between two consecutive attempts to rebuild the underlying client after an I/O reactor
+     * shutdown. The effective delay grows exponentially (up to {@link #MAX_REBUILD_BACKOFF_NANOS}) while the reactor
+     * keeps dying, and resets once it has been stable.
+     */
+    static final long DEFAULT_REBUILD_BACKOFF_MILLIS = 1000L;
+
+    /** Upper bound for the exponentially-growing rebuild back-off. */
+    private static final long MAX_REBUILD_BACKOFF_NANOS = 30_000L * 1_000_000L;
+
     private final JsonpMapper mapper;
-    private final CloseableHttpAsyncClient client;
+    private final AtomicReference<CloseableHttpAsyncClient> clientRef;
+    @Nullable
+    private final Supplier<CloseableHttpAsyncClient> clientFactory;
+    private final ReentrantLock rebuildLock = new ReentrantLock();
+    private final AtomicBoolean closed = new AtomicBoolean(false);
+    private final long rebuildBackoffBaseNanos;
+    // The following three fields are guarded by rebuildLock.
+    private boolean rebuiltOnce = false;
+    private long lastRebuildNanos = 0L;
+    private long currentBackoffNanos = 0L;
+    // Monotonic nanosecond clock; overridable for deterministic back-off testing.
+    private volatile LongSupplier nanoClock = System::nanoTime;
     private final ApacheHttpClient5Options transportOptions;
     private final ConcurrentMap<HttpHost, DeadHostState> denylist = new ConcurrentHashMap<>();
     private final AtomicInteger lastNodeIndex = new AtomicInteger(0);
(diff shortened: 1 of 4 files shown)


java-client/src/main/java/org/opensearch/client/transport/httpclient5/internal/HeapBufferedAsyncEntityConsumer.java
```diff
@@ -84,34 +117,127 @@ protected void data(final ByteBuffer src, final boolean endOfStream) throws IOEx
...
+     * @param initialCapacity the initial capacity of the buffer
+     * @return a new {@link ByteArrayBuffer}
+     */
+    protected ByteArrayBuffer createBuffer(final int initialCapacity) {
+        return new ByteArrayBuffer(initialCapacity);
+    }
+
+    /**
+     * Releases the (potentially very large) response buffer and converts a fatal {@link OutOfMemoryError} into a
+     * recoverable {@link IOException}.
```
andrross: I'm not sure it is safe to do this. The contract of java.lang.Error generally is that it "indicates serious problems that a reasonable application should not try to catch". I don't think you can assume the JVM is in a workable state after observing an error and the right thing to do is let the process crash. In this case, you don't actually know if the response buffer in question here caused the OOM, versus something else in the application unrelated to this client that consumed all the heap.

andrross:
Would it make sense to break this into 2 PRs? One with the fix to allow recovery from reactor shutdowns, and another PR with the changes to reduce the chance of reactor-killing memory failures.

java-client/src/main/java/org/opensearch/client/transport/httpclient5/ApacheHttpClient5Transport.java
```diff
@@ -218,61 +302,246 @@ private void performRequestAsync(
...
+        try {
+            future = client.execute(context.requestProducer, context.asyncResponseConsumer, context.context, callback);
+        } catch (final RuntimeException runtimeFailure) {
+            if (isRecoverableReactorFailure(runtimeFailure, client) == false) {
+                throw runtimeFailure;
+            }
+            // The I/O reactor has been shut down. Try to recover by rebuilding the client, then retry this request once
+            // on the fresh client.
+            retryAfterReactorFailure(nodeTuple, options, request, warningsHandler, listener, node, client, runtimeFailure, allowRecovery);
+            return;
```
andrross, line 392: Nitpick, but easier to follow if you invert the logic:


```suggestion
            if (isRecoverableReactorFailure(runtimeFailure, client)) {
                // The I/O reactor has been shut down. Try to recover by rebuilding the client, then retry this request once
                // on the fresh client.
                retryAfterReactorFailure(nodeTuple, options, request, warningsHandler, listener, node, client, runtimeFailure, allowRecovery);
            
            } else {
                throw runtimeFailure;
            }
```

andrross (approved):
Looks good to me. @reta can you take a look?

@sharp-pixel The PR description should be updated since this no longer has all the original changes.

reta:
> Looks good to me. @reta can you take a look?

Yeah I will take a look in a few days, thanks @andrross and @sharp-pixel

reta:
@sharp-pixel few things from my side (thanks for working on it first of all):
 - I was under impression that much smaller scoped change will be ported (from https://github.com/elastic/elasticsearch-java/pull/1049), I think that one helps to eliminate the cause of reactor shutdowns
 - Having the same behavior across both clients is really helpful (since other projects could support OpenSearch and Elasticsearch with minimal efforts)

In general, I think there is nothing fundamentally wrong to ask the caller to recreate the client in case of fatal failures (like I/O reactor one). The current fix is changing a lot of internals of the ApacheHttpClient5Transport, but we should not forget that we also have AwsSdk2Transport (that internally uses Apache5HttpClient) and it would be reasonable to expect the same client recovery there (since it is the same AHC5 client under the hood, with all the same issues).

May be we should take a step back and:
 - port the fix from `elasticsearch-java`
 - have a wrapper around `OpenSearchTransport` (something like `RecoverableOpenSearchTransport`) that will accept a builder and would recreate the underlying transport in case of recoverable failures? (that may require some additional hooks / listeners but that should not be too difficult I believe)
 
Thank you.

andrross:
> I was under impression that much smaller scoped change will be ported (from https://github.com/elastic/elasticsearch-java/pull/1049), I think that one helps to eliminate the cause of reactor shutdowns

@reta That change causes certain code paths to convert OutOfMemoryErrors into RuntimeExceptions and allow the JVM to continue operating. Maybe I'm paranoid but that seems like a bad idea? If I'm wrong about that then it's probably fine to take the smaller-scoped change.

reta:
> > I was under impression that much smaller scoped change will be ported (from [elastic/elasticsearch-java#1049](https://github.com/elastic/elasticsearch-java/pull/1049)), I think that one helps to eliminate the cause of reactor shutdowns
> 
> @reta That change causes certain code paths to convert OutOfMemoryErrors into RuntimeExceptions and allow the JVM to continue operating. Maybe I'm paranoid but that seems like a bad idea? If I'm wrong about that then it's probably fine to take the smaller-scoped change.

@andrross It does indeed, but it has been applied to widely used software (elasticsearch-java), so may be we both are overly paranoid?

andrross:
> > > I was under impression that much smaller scoped change will be ported (from [elastic/elasticsearch-java#1049](https://github.com/elastic/elasticsearch-java/pull/1049)), I think that one helps to eliminate the cause of reactor shutdowns
> > 
> > 
> > @reta That change causes certain code paths to convert OutOfMemoryErrors into RuntimeExceptions and allow the JVM to continue operating. Maybe I'm paranoid but that seems like a bad idea? If I'm wrong about that then it's probably fine to take the smaller-scoped change.
> 
> @andrross It does indeed, but it has been applied to widely used software (elasticsearch-java), so may be we both are overly paranoid?

@reta Without the fix, does the OOM get swallowed anyway, just with the side effect that the netty reactor is shut down?

reta:
> > > > I was under impression that much smaller scoped change will be ported (from [elastic/elasticsearch-java#1049](https://github.com/elastic/elasticsearch-java/pull/1049)), I think that one helps to eliminate the cause of reactor shutdowns
> > > 
> > > 
> > > @reta That change causes certain code paths to convert OutOfMemoryErrors into RuntimeExceptions and allow the JVM to continue operating. Maybe I'm paranoid but that seems like a bad idea? If I'm wrong about that then it's probably fine to take the smaller-scoped change.
> > 
> > 
> > @andrross It does indeed, but it has been applied to widely used software (elasticsearch-java), so may be we both are overly paranoid?
> 
> @reta Without the fix, does the OOM get swallowed anyway, just with the side effect that the netty reactor is shut down?

I am not 100% sure, I will look closely over the weekend, I believe that either a) reactor is shutdown cleanly b) reactor is not shutdown, but the thread(s) dies?

andrross:
@reta If the existing behavior is that Netty prevents the OOM exception from propagating to the application, then the fix to prevent OOMs from shutting down the reactor seems much less risky.

reta:
> @reta If the existing behavior is that Netty prevents the OOM exception from propagating to the application, then the fix to prevent OOMs from shutting down the reactor seems much less risky.

Thanks @andrross , I will double check that, thanks!

reta:
> > @reta If the existing behavior is that Netty prevents the OOM exception from propagating to the application, then the fix to prevent OOMs from shutting down the reactor seems much less risky.
> 
> Thanks @andrross , I will double check that, thanks!

@andrross so I did check the XxxError handling that the patch applies (from https://github.com/elastic/elasticsearch-java/pull/1049) prevents the reactor shutdown - basically, there is no need to recreate the client

PR #2039, opened 2026-07-05
Fix ISM and other timestamp fields typed as integer instead of int64 causing numeric overflow

### LINKED ISSUE #1717: [BUG] ISM API - Put Policies: `java.lang.RuntimeException: Jackson exception: Numeric value out of range of int (-2147483648 - 2147483647)`

### LINKED ISSUE #1898: [BUG] ISM types lastUpdatedTime is susceptible to int overflow

gingeekrishna:
Fixes #1717, Fixes #1898

Epoch-millisecond timestamps were typed as plain `integer` in the spec with no `format: int64`, so the codegen emitted `Integer` (max ~2.1B) instead of `Long`. Any real timestamp value (e.g. `1755915614485`) overflows `int` and Jackson raises:
```
Numeric value out of range of int (-2147483648 - 2147483647)
```

**Fix:** Add `format: int64` to affected fields — `IsmTemplate.last_updated_time`, `Policy.last_updated_time`, and the same fields in `flow_framework` and SM — and update the two generated ISM Java classes from `Integer`/`integerDeserializer` to `Long`/`longDeserializer`.

### DIFF
--- a/java-codegen/opensearch-openapi.yaml
+++ b/java-codegen/opensearch-openapi.yaml
@@ -56686,9 +56686,11 @@ components:
           $ref: '#/components/schemas/flow_framework.common___user'
         created_time:
           type: integer
+          format: int64
           description: When the workflow was created.
         last_updated_time:
           type: integer
+          format: int64
           description: When the workflow was last updated.
         last_provisioned_time:
           type: number
@@ -61772,6 +61774,7 @@ components:
           description: The priority of the ISM template.
         last_updated_time:
           type: integer
+          format: int64
           description: When the ISM template was last updated.
     ism._common___Metadata:
       type: object
@@ -61797,6 +61800,7 @@ components:
           description: The description of the policy.
         last_updated_time:
           type: integer
+          format: int64
           description: When the policy was last updated.
         schema_version:
           type: number
@@ -70265,6 +70269,7 @@ components:
           type: boolean
         last_updated_time:
           type: integer
+          format: int64
           description: When the policy was last modified.
         enabled_time:
           type: integer
--- a/java-codegen/opensearch-openapi.yaml
+++ b/java-codegen/opensearch-openapi.yaml
@@ -70273,6 +70273,7 @@ components:
           description: When the policy was last modified.
         enabled_time:
           type: integer
+          format: int64
           description: When the policy was last enabled.
       required:
         - creation
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Run model tests against both JSON mappers instead of picking one at random ([#2085](https://github.com/opensearch-project/opensearch-java/pull/2085))
 - Fix currentSize calculation in BulkIngester ([#2113](https://github.com/opensearch-project/opensearch-java/pull/2113))
 - Use protocol version from response (rather than request) ([#2118](https://github.com/opensearch-project/opensearch-java/pull/2118))
+- Fix ISM timestamp fields typed as `Integer` instead of `Long` causing numeric overflow for epoch-millisecond values ([#2039](https://github.com/opensearch-project/opensearch-java/pull/2039))
 
 ### Changed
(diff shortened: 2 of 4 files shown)


java-codegen/opensearch-openapi.yaml
reta, line 1: Please submit a change against https://github.com/opensearch-project/opensearch-api-specification/blob/main/spec/schemas/ism._common.yaml, thank you

reta (changes requested):

java-codegen/opensearch-openapi.yaml
gingeekrishna: Submitted the spec fix upstream: https://github.com/opensearch-project/opensearch-api-specification/pull/1196

gingeekrishna:
Submitted the spec fix upstream: https://github.com/opensearch-project/opensearch-api-specification/pull/1196

PR #2040, opened 2026-07-05
Fix ShardProfile.fetch typed as single FetchProfile instead of List<FetchProfile>

### LINKED ISSUE #1965: [BUG] Client is unable to deserialize search response with profile API enabled
### What is the bug?
When enabling` .profile(true)` on a search query, the opensearch-java client crashes with an `UnexpectedJsonEventException: Unexpected JSON event 'START_ARRAY' instead of '[START_OBJECT, KEY_NAME]'.`
This occurs because the OpenSearch server returns the fetch profile metrics as a JSON Array `("fetch": [ ... ]).` However, the OpenAPI specification (and thus the auto-generated Java client) incorrectly defines fetch as a single object (FetchProfile) rather than a list of objects (List<FetchProfile>). Because of this mismatch, the Java deserializer expects a { but encounters a [ and immediately fails.

Execute a standard Search Request with profiling enabled.

The opensearch-java client should successfully deserialize the profile response without crashing.

Java Client Version: opensearch-java v3.4.0 (tried on latest as well)
OpenSearch Server Version: 3.3

```
"aggregations": [],
                "fetch": [
                    {
                        "type": "fetch",
                        "description": "fetch",
                        "time_in_nanos": 177937,
                        "breakdown": {
                            "build_sub_phase_processors": 10339,
                            "build_sub_phase_processors_count": 1,
                            "create_stored_fields_visitor": 2642,
                            "create_stored_fields_visitor_count": 1,
                            "get_next_reader": 9412,
                            "get_next_reader_count": 2,
                            "load_source": 1732,
                            "load_source_count": 2,
                            "load_stored_fields": 151041,
                            "load_stored_fields_count": 2
                        },
                        "children": [
                            {
                                "type": "FetchSourcePhase",
                                "description": "FetchSourcePhase",
                                "time_in_nanos": 2771,
                                "breakdown": {
                                    "process": 1509,
                                    "process_count": 2,
                                    "set_next_reader": 1262,
                                    "set_next_reader_count": 2
                                }
                            }
                        ]
```

gingeekrishna:
Fixes #1965

OpenSearch returns `fetch` as a JSON array in profile responses, but the spec defined it as a single `FetchProfile` object. This caused:
```
UnexpectedJsonEventException: Unexpected JSON event 'START_ARRAY' instead of '[START_OBJECT, KEY_NAME]'
```

**Fix:** Change the spec to `type: array / items: $ref: FetchProfile` and update the generated `ShardProfile.java` to use `List<FetchProfile>` with array serialization and `arrayDeserializer`, matching the pattern already used by `aggregations` and `searches` in the same class.

### DIFF
--- a/java-codegen/opensearch-openapi.yaml
+++ b/java-codegen/opensearch-openapi.yaml
@@ -52512,7 +52512,9 @@ components:
           items:
             $ref: '#/components/schemas/_core.search___SearchProfile'
         fetch:
-          $ref: '#/components/schemas/_core.search___FetchProfile'
+          type: array
+          items:
+            $ref: '#/components/schemas/_core.search___FetchProfile'
       required:
         - aggregations
         - id
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Run model tests against both JSON mappers instead of picking one at random ([#2085](https://github.com/opensearch-project/opensearch-java/pull/2085))
 - Fix currentSize calculation in BulkIngester ([#2113](https://github.com/opensearch-project/opensearch-java/pull/2113))
 - Use protocol version from response (rather than request) ([#2118](https://github.com/opensearch-project/opensearch-java/pull/2118))
+- Fix `ShardProfile.fetch` typed as single `FetchProfile` instead of `List<FetchProfile>` causing deserialization failure with profile API ([#2040](https://github.com/opensearch-project/opensearch-java/pull/2040))
 
 ### Changed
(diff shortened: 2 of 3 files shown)


java-codegen/opensearch-openapi.yaml
reta, line 1: Please submit a change against https://github.com/opensearch-project/opensearch-api-specification/blob/main/spec/schemas/_core.search.yaml#L201, thank you

reta (changes requested):

java-codegen/opensearch-openapi.yaml
gingeekrishna: Submitted the spec fix upstream: https://github.com/opensearch-project/opensearch-api-specification/pull/1197

gingeekrishna:
Submitted the spec fix upstream: https://github.com/opensearch-project/opensearch-api-specification/pull/1197

PR #2041, opened 2026-07-05
Fix NPE in _listAddAll when server returns JSON null for a list field

### LINKED ISSUE #1813: [BUG] `required` spec not respected for maps and lists
### What is the bug?
In the spec, a null value can be allowed for lists, e.g. here:
https://github.com/opensearch-project/opensearch-api-specification/blob/963f90293d480015deaf352d1987bec2f18916c2/spec/schemas/ism._common.yaml#L50-L57

Deserializing these will fail with a NPE due to the `Objects.requireNonNull` in `_listAddAll`

ISM Policies can be created without nested ISM Templates. As an example, see the create policy request/response in the [docs](https://docs.opensearch.org/latest/im-plugin/ism/api/#example-request)

However, reading a policy created like this without the client fails due to the list being `null`
```
java.lang.NullPointerException
	at java.base/java.util.Objects.requireNonNull(Objects.java:233)
	at org.opensearch.client.util.ObjectBuilderBase._listAddAll(ObjectBuilderBase.java:101)
	at org.opensearch.client.opensearch.ism.Policy$Builder.ismTemplate(Policy.java:379)
	at org.opensearch.client.json.ObjectDeserializer$FieldObjectDeserializer.deserialize(ObjectDeserializer.java:82)
	at org.opensearch.client.json.ObjectDeserializer.deserialize(ObjectDeserializer.java:185)
	at org.opensearch.client.json.ObjectDeserializer.deserialize(ObjectDeserializer.java:146)
	at org.opensearch.client.json.JsonpDeserializer.deserialize(JsonpDeserializer.java:87)
	at org.opensearch.client.json.ObjectBuilderDeserializer.deserialize(ObjectBuilderDeserializer.java:91)
```

lists and maps should respect `required` like all other properties. If this is not possible, deserialization should take care of initializing them if they are missing from the JSON (however, the two representations would not be equal then)

OS java client 3.2/3.3

I wonder why they are conditionally excluded from the required check. What is the reason for that?

gingeekrishna:
Fixes #1813

`_listAddAll(list, values)` threw `NullPointerException` via `Objects.requireNonNull(values)` when `values` was `null`. This happens when OpenSearch returns JSON `null` for an optional list field (e.g. `ism_template` in ISM Policy), causing deserialization to fail:
```
NullPointerException at ObjectBuilderBase._listAddAll
  at Policy$Builder.ismTemplate(Policy.java:379)
```

**Fix:** Guard against null `values` by returning the existing list unchanged when `values` is null. Same logic applied to `_mapPutAll` would follow the same pattern if similar issues arise for maps.

### DIFF
--- a/java-codegen/opensearch-openapi.yaml
+++ b/java-codegen/opensearch-openapi.yaml
@@ -52512,7 +52512,9 @@ components:
           items:
             $ref: '#/components/schemas/_core.search___SearchProfile'
         fetch:
-          $ref: '#/components/schemas/_core.search___FetchProfile'
+          type: array
+          items:
+            $ref: '#/components/schemas/_core.search___FetchProfile'
       required:
         - aggregations
         - id
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Fix currentSize calculation in BulkIngester ([#2113](https://github.com/opensearch-project/opensearch-java/pull/2113))
 - Use protocol version from response (rather than request) ([#2118](https://github.com/opensearch-project/opensearch-java/pull/2118))
 - Fix `ShardProfile.fetch` typed as single `FetchProfile` instead of `List<FetchProfile>` causing deserialization failure with profile API ([#2040](https://github.com/opensearch-project/opensearch-java/pull/2040))
+- Fix `NullPointerException` in `_listAddAll` when server returns JSON `null` for a list field ([#2041](https://github.com/opensearch-project/opensearch-java/pull/2041))
 
 ### Changed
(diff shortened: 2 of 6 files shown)


java-codegen/opensearch-openapi.yaml
```diff
@@ -52512,7 +52512,9 @@ components:
           items:
             $ref: '#/components/schemas/_core.search___SearchProfile'
         fetch:
-          $ref: '#/components/schemas/_core.search___FetchProfile'
+          type: array
+          items:
+            $ref: '#/components/schemas/_core.search___FetchProfile'
```
Hailong-am, line 52517: should we update https://github.com/opensearch-project/opensearch-api-specification/blob/5691b4a54b7cf485a7c5d1e1b53c1987348f4e13/spec/schemas/_core.search.yaml#L201-L202 here and sync from it?

reta: Correct, this is the way we generate Java client, thank you @Hailong-am

Issue #1093, opened 2024-07-19
[RELEASE] Release v. Next


jordanpadams:
Can we get a new tagged release of the library per bug fix #1068? This is a blocker for our AOSS migration efforts.

If this is a lot of effort, [fixing this bug with publishing this SNAPSHOT](https://github.com/opensearch-project/opensearch-java/issues/1090) would at least allow us to get out an interim build. Thanks!

dblock:
Tagging @Xtansia for a release, or I can take care of it early next week.

@jordanpadams want to try and help with the fix for the SNAPSHOT workflow?

jordanpadams:
@dblock the 403 forbidden indicates an issue with the login information, which I do not have access to. It needs to be updated either with the correct username/password, or it may need to be a generated set of tokens generated through Sonatype: https://central.sonatype.org/publish/generate-token/

dblock:
2.12.0 was released, https://github.com/opensearch-project/opensearch-java/releases/tag/v2.12.0

tloubrieu-jpl:
Hi @dblock ,

Something does not seems right with the release.

The bug https://github.com/opensearch-project/opensearch-java/issues/1068 we wanted to fix is not fixed in the release (according to my tests).

And, randomly, I also noticed that the following commit https://github.com/opensearch-project/opensearch-java/commit/373802683b0380985d6f246218d2e1b051c8f394 is not in the release.

Do you have an idea why that is ?

Thanks

Xtansia:
@tloubrieu-jpl I can confirm from decompiling the 2.12.0 jar from Maven that it does contain the changes from PR #1072. Can you please confirm your POM/Gradle/etc. configuration and which version of the dependency you are taking?

It will not contain commit 3738026 as that is only going in the next major release (3.0.0) as it is a breaking change.

tloubrieu-jpl:
Thanks @Xtansia , I captured the maven dependencies of my project there:

```
 +- org.opensearch.client:opensearch-java:jar:2.12.0:compile
[INFO] |  +- com.google.code.findbugs:jsr305:jar:3.0.2:compile
[INFO] |  +- org.eclipse.parsson:parsson:jar:1.1.6:compile
[INFO] |  |  \- jakarta.json:jakarta.json-api:jar:2.1.3:compile
[INFO] |  +- org.opensearch.client:opensearch-rest-client:jar:2.12.0:runtime
[...]

Issue #1089, opened 2024-07-17
[BUG]  FieldSort not implement SortOptionsVariant


jbrun:
### What is the bug?
When we want to sort a request, FieldSort doesn’t implement SortOptionVariant, so we cannot use it in SearchRequest with SortOptionsBuilders

### How can one reproduce the bug?
`SearchRequest.Builder searchRequest = new SearchRequest.Builder()
.sort(SortOptionsBuilders.field().field(“lastName”).order(SortOrder.Desc).build()._toSortOptions());`

or

`SearchRequest.Builder searchRequest = new SearchRequest.Builder()
.sort(FieldSort.of(s → s.field(“lastName”).order(ortOrder.Desc))._toSortOptions());`

._toSortOptions() not available

However ScoreSort, GeoDistanceSort and ScriptSort implement SortOptionVariant

### What is the expected behavior?
package org.opensearch.client.opensearch._types;
public class FieldSort implements **SortOptionsVariant**, JsonpSerializable

### What is your host/environment?
opensearch-java:2.10.4

dblock:
[Catch All Triage - [1](https://github.com/dblock), [2](https://github.com/krisfreedain), [3](https://github.com/amistrn)]

PR #2062, opened 2026-07-15
gRPC support for opensearch-java

dayana-j:
### Description
This PR adds a transparent gRPC support to the opensearch-java as a separate `java-client-grpc` module with a translation layer, transport layers, connection management, and security implementation for gRPC such as TLS/basic auth, SigV4, and JWT. Bulk operations are automatically routed over gRPC for improved performance; lower latency, higher throughput, and smaller payloads via HTTP/2 and Protocol Buffers, while all other operations fall back to REST seamlessly. Users configure the transport once and use the standard `OpenSearchClient` API without any protobuf imports, channel management, or code changes.

OpenSearch 3.5+ exposes a gRPC endpoint (default port 9400) alongside REST for Bulk and k-NN operations. This implementation bridges the gap so that existing client code benefits from gRPC performance without migration effort.
The gRPC transport lives in its own module to isolate gRPC dependencies (`grpc-netty-shaded`, `protobuf-java`) from the core client, preventing classpath conflicts with the OpenSearch test framework.

Users add the gRPC module as an optional dependency:
```
dependencies {
    implementation 'org.opensearch.client:opensearch-java:3.x'
    implementation 'org.opensearch.client:opensearch-java-grpc:3.x'  // optional
}
```
### How It Works
The `HybridTransport` wraps both a `GrpcTransport` (for bulk) and a standard REST transport (for everything else). When `client.bulk()` is called, the request is converted from the client's Java types to protobuf, sent over gRPC, and the response is converted back — all transparently. If gRPC is unavailable or the conversion fails, the request automatically retries via REST.

```
var restTransport = ApacheHttpClient5TransportBuilder.builder(new HttpHost("http", "localhost", 9200)).build();

var grpcTransport = GrpcTransport.builder("localhost", 9400)
    .jsonpMapper(new JacksonJsonpMapper())
    .basicAuth("admin", "admin")
    .build();

var client = new OpenSearchClient(new HybridTransport(grpcTransport, restTransport));

client.bulk(bulkRequest);   // → gRPC (fast path)
client.search(searchReq);   // → REST (automatic fallback)
```

### Authentication
All three authentication methods supported by OpenSearch are implemented as gRPC `ClientInterceptor`s that attach credentials to every outgoing call's metadata.

**TLS + Basic Auth:**
```
var grpcTransport = GrpcTransport.builder("localhost", 9400)
    .jsonpMapper(new JacksonJsonpMapper())
    .tls(GrpcTlsConfig.builder()
        .trustCertificatePath("/path/to/ca.pem")
        .build())
    .basicAuth("admin", "admin")
    .build();
```

**AWS SigV4** signs each gRPC call using AWS SDK v2 credential providers. The interceptor constructs a synthetic URL from the gRPC method path, signs it with `AwsV4HttpSigner`, and attaches the Authorization headers as gRPC metadata. Body-aware payload signing is supported for full SigV4 compliance.
```
var grpcTransport = GrpcTransport.builder("domain.us-east-1.es.amazonaws.com", 9400)
    .jsonpMapper(new JacksonJsonpMapper())
    .tls(GrpcTlsConfig.builder().build())
    .sigV4(GrpcSigV4Config.builder()
        .region(Region.US_EAST_1)
        .service("es")
        .credentialsProvider(DefaultCredentialsProvider.create())
        .build())
    .build();
```

**JWT** uses a token supplier called on every request for automatic token rotation:
`.jwtAuth(() -> myOidcProvider.getAccessToken())`

### Connection Management
The gRPC `ManagedChannel` provides built-in connection health management that's more capable than REST's passive dead-node tracking. The channel automatically reconnects on failure with exponential backoff, and the `GrpcChannelHealthMonitor` exposes the connectivity state machine (`IDLE → CONNECTING → READY → TRANSIENT_FAILURE`) so callers can check readiness or register state change callbacks.

```
// Warm up the connection at startup
transport.healthMonitor().waitForReady(5, TimeUnit.SECONDS);

// Check before sending
if (transport.healthMonitor().isReady()) { /* connected */ }
```
Retry logic is configurable via `GrpcTransportOptions` — max retries, backoff interval, keepalive pings, idle timeout, and max inbound message size all have sensible defaults that match the OpenSearch server's gRPC settings.

### Translation Layer
`BulkRequestConverter` walks the typed `BulkOperation` list, serializes document bodies to JSON bytes via `JsonpMapper`, and builds the protobuf `BulkRequest`. `BulkResponseConverter` converts each protobuf `ResponseItem` back to the client's `BulkResponseItem` with all fields. Status code mapping follows the OpenSearch server's `RestToGrpcStatusConverter`.

### Tests
The implementation includes 112 unit tests across 6 test files covering all conversion logic, transport routing, authentication interceptors, channel management, and TLS configuration. Three condensed integration tests validate end-to-end behavior against a real OpenSearch 3.5+ cluster using the project's existing test framework (`OpenSearchJavaClientTestCase` + Testcontainers). Tests are version-gated via `assumeTrue` and skip automatically on older clusters.

### CI Configuration
The Docker configuration in `.ci/opensearch/` enables gRPC transport (port 9400) in the test container. A standalone `validate-grpc-integration.sh` script is included for local validation. gRPC integration tests run via `./gradlew :java-client-grpc:test`.

### Dependencies
gRPC support is configured as an optional feature (`grpcSupport`) with zero impact on users who don't enable it. Dependencies are `io.grpc` (1.68.0), `com.google.protobuf` (3.25.5), and `org.opensearch:protobufs` (1.2.0).

### Related
This is the Java counterpart to the opensearch-py gRPC implementation, following the same design pattern:
- [opensearch-py #1058](https://github.com/opensearch-project/opensearch-py/pull/1058) — gRPC translation layer
- [opensearch-py #1078](https://github.com/opensearch-project/opensearch-py/pull/1078) — gRPC transport and connection management
- [opensearch-py #1087](https://github.com/opensearch-project/opensearch-py/pull/1087) — gRPC TLS & Basic auth
- [opensearch-py #1089](https://github.com/opensearch-project/opensearch-py/pull/1089) — gRPC SigV4
- [opensearch-py #1093](https://github.com/opensearch-project/opensearch-py/pull/1093) — gRPC JWT

OpenSearch gRPC API documentation: https://docs.opensearch.org/latest/api-reference/grpc-apis/index/


finnegancarroll:
>If gRPC is unavailable or the conversion fails, the request automatically retries via REST.

I think if we encounter some gRPC error we should explicitly propagate this back to the user instead of silently falling back to REST here.

java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/translation/FieldMappingUtil.java
```diff
@@ -0,0 +1,176 @@
...
+import org.opensearch.protobufs.OpType;
+import org.opensearch.protobufs.Refresh;
+import org.opensearch.protobufs.VersionType;
+import org.opensearch.protobufs.WaitForActiveShardOptions;
+import org.opensearch.protobufs.WaitForActiveShards;
+
+/**
+ * Maps opensearch-java client enum types to their protobuf equivalents.
+ */
+public final class FieldMappingUtil {
```
reta: Let's keep this one package private if possible:

```suggestion
final class FieldMappingUtil {
```

java-client-grpc/build.gradle.kts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+repositories {
+    mavenLocal()
+    maven(url = "https://ci.opensearch.org/ci/dbc/snapshots/maven/")
+    mavenCentral()
+}
+
+java {
+    targetCompatibility = JavaVersion.VERSION_11
```
reta: This is a bummer (somewhat), the opensearch-java baseline is JDK 8, any specific reasons to use JDK 11 baseline here?

java-client-grpc/build.gradle.kts
```diff
@@ -0,0 +1,95 @@
...
+    // For AwsSdk2 SigV4 support (optional, compile-only)
+    compileOnly("software.amazon.awssdk", "sdk-core", "[2.21,3.0)")
+    compileOnly("software.amazon.awssdk", "auth", "[2.21,3.0)")
+    compileOnly("software.amazon.awssdk", "http-auth-aws", "[2.21,3.0)")
+
+    // Test dependencies
+    testImplementation("io.grpc", "grpc-testing", grpcVersion)
+    testImplementation("junit", "junit", "4.13.2")
+    testImplementation("org.opensearch.client", "opensearch-rest-client", opensearchVersion)
+    testImplementation("tools.jackson.core", "jackson-core", "3.1.1")
```
reta: I believe you don't need those (tools.jackson.core), `opensearch-java` has hard dependency on Jackson 3.x (`main` branch), it will be brought up transitively.

java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/BasicAuthInterceptor.java
```diff
@@ -0,0 +1,96 @@
...
+ * <pre>{@code
+ * BasicAuthInterceptor interceptor = new BasicAuthInterceptor("admin", "admin");
+ * ManagedChannel channel = ManagedChannelBuilder.forAddress("localhost", 9400)
+ *     .intercept(interceptor)
+ *     .build();
+ * }</pre>
+ *
+ * @see <a href="https://docs.opensearch.org/latest/security/authentication-backends/basic-authc/">OpenSearch Basic Auth</a>
+ */
+public class BasicAuthInterceptor implements ClientInterceptor {
```
reta: Let's make is package private please (since it is used by the transport only):

```suggestion
class BasicAuthInterceptor implements ClientInterceptor {
```

java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/translation/GrpcStatusConverter.java
```diff
@@ -0,0 +1,217 @@
...
+        // utility class
+    }
+
+    /**
+     * Converts a gRPC {@link StatusRuntimeException} to the appropriate opensearch-java exception.
+     *
+     * @param e the gRPC exception
+     * @return the converted exception (TransportException, OpenSearchException, or UnsupportedOperationException)
+     */
+    public static Exception convert(StatusRuntimeException e) {
```
reta: Let's make is package private please (since it is used by this class only):

```suggestion
    static Exception convert(StatusRuntimeException e) {
```

java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/translation/GrpcStatusConverter.java
```diff
@@ -0,0 +1,217 @@
...
+        return new OpenSearchException(errorResponse);
+    }
+
+    /**
+     * Returns the HTTP status code equivalent for a gRPC status code.
+     *
+     * @param code the gRPC status code
+     * @return HTTP status code equivalent
+     */
+    public static int toHttpStatus(Status.Code code) {
```
reta: This is not used anywhere, right? If such - please remove

java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/BasicAuthInterceptor.java
```diff
@@ -0,0 +1,96 @@
...
+    private final String authHeaderValue;
+
+    /**
+     * Creates a basic auth interceptor with the given credentials.
+     *
+     * @param username the username
+     * @param password the password
+     * @throws IllegalArgumentException if username or password is null
+     */
+    public BasicAuthInterceptor(String username, String password) {
```
reta:
```suggestion
    BasicAuthInterceptor(String username, String password) {
```

java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/GrpcChannelFactory.java
```diff
@@ -0,0 +1,204 @@
...
+ * Uses gRPC-Netty's shaded SSL context for TLS channels, supporting:
+ * <ul>
+ *   <li>System default trust store (JVM cacerts)</li>
+ *   <li>Custom CA certificate (PEM file)</li>
+ *   <li>Custom trust store (JKS/PKCS12)</li>
+ *   <li>Mutual TLS with client certificate</li>
+ *   <li>Insecure mode (trust all — development only)</li>
+ * </ul>
+ */
+public final class GrpcChannelFactory {
```
reta: Let's make is package private please (since it is used by the transport only):

```suggestion
final class GrpcChannelFactory {
```

java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/GrpcChannelFactory.java
```diff
@@ -0,0 +1,204 @@
...
+    /**
+     * Creates a plaintext (unencrypted) gRPC channel.
+     *
+     * @param host       the server hostname
+     * @param port       the server port
+     * @param options    channel options (keepalive, max message size, etc.)
+     * @param interceptors optional client interceptors (e.g., auth)
+     * @return a configured ManagedChannel
+     */
+    public static ManagedChannel createPlaintextChannel(
```
reta:
```suggestion
    private static ManagedChannel createPlaintextChannel(
```

java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/GrpcChannelFactory.java
```diff
@@ -0,0 +1,204 @@
...
+     *
+     * @param host       the server hostname
+     * @param port       the server port
+     * @param tlsConfig  TLS configuration (certs, trust store, insecure mode)
+     * @param options    channel options (keepalive, max message size, etc.)
+     * @param interceptors optional client interceptors (e.g., auth)
+     * @return a configured ManagedChannel with TLS
+     * @throws IOException if SSL context creation fails
+     */
+    public static ManagedChannel createTlsChannel(
```
reta:
```suggestion
    private static ManagedChannel createTlsChannel(
```

java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/GrpcChannelFactory.java
```diff
@@ -0,0 +1,204 @@
...
+     *
+     * @param host         the server hostname
+     * @param port         the server port
+     * @param tlsConfig    TLS configuration, or null for plaintext
+     * @param options      channel options
+     * @param interceptors client interceptors
+     * @return a configured ManagedChannel
+     * @throws IOException if TLS context creation fails
+     */
+    public static ManagedChannel createChannel(
```
reta:
```suggestion
    static ManagedChannel createChannel(
```

java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/GrpcTransport.java
```diff
@@ -0,0 +1,436 @@
...
+ * <pre>{@code
+ * GrpcTransport transport = GrpcTransport.builder("localhost", 9400)
+ *     .jsonpMapper(mapper)
+ *     .build();
+ *
+ * OpenSearchClient client = new OpenSearchClient(transport);
+ * client.bulk(bulkRequest); // goes over gRPC
+ * }</pre>
+ */
+public class GrpcTransport implements OpenSearchTransport {
```
reta, line 46: So for `opensearch-java` client, we have made deliberate split between general and AWS specific transport (sigv4), could we please follow the same pattern? Fe `GrpcTransport`  and `AwsGrpcTransport`, thank you

java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/GrpcTransport.java
```diff
@@ -0,0 +1,436 @@
...
+
+    // ─── Supported Endpoints Registry ────────────────────────────────────────────
+
+    private static final java.util.Set<Endpoint<?, ?, ?>> SUPPORTED_ENDPOINTS;
+
+    static {
+        java.util.Set<Endpoint<?, ?, ?>> endpoints = new java.util.HashSet<>();
+        endpoints.add(BulkRequest._ENDPOINT);
```
reta, line 54: We also support `k-nn` endpoints here (at least some of), you also mentioned that but implementation is omitted? Why is that?

java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/GrpcTransportOptions.java
```diff
@@ -0,0 +1,166 @@
...
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static GrpcTransportOptions defaults() {
+        return new Builder().build();
+    }
+
+    public static final class Builder {
+        private int maxInboundMessageSize = 10 * 1024 * 1024; // 10MB
```
reta, line 113: Just wondering, 10MB is generally large but not for _bulk requests (those could be much larger), is this limit coming from protobuf constraints dealing large messages?

java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/HybridTransport.java
```diff
@@ -0,0 +1,175 @@
...
+ * client.search(searchReq);  // → REST
+ * }</pre>
+ */
+public class HybridTransport implements OpenSearchTransport {
+
+    private static final Logger logger = Logger.getLogger(HybridTransport.class.getName());
+
+    private final GrpcTransport grpcTransport;
+    private final OpenSearchTransport restTransport;
+    private final boolean fallbackOnError;
```
reta: I would probably agree wth https://github.com/opensearch-project/opensearch-java/pull/2062#issuecomment-4996678818, the fallback part is really confusing. If the intent is to use GRPC, it is better be GRPC, right? (I mean for endpoints that supposed to be backed by GRPC)

java-client-grpc/build.gradle.kts
```diff
@@ -0,0 +1,95 @@
...
+            pom {
+                name.set("OpenSearch Java Client - gRPC Transport")
+                packaging = "jar"
+                artifactId = "opensearch-java-grpc"
+                description.set("gRPC transport for the OpenSearch Java Client.")
+                url.set("https://github.com/opensearch-project/opensearch-java/")
+            }
+        }
+    }
+}
```
reta, line 152: Its seems likest integration tests are not being run:

```
val runtimeJavaVersion = (System.getProperty("runtime.java")?.toInt())?.let(JavaVersion::toVersion) ?: JavaVersion.current()
if (runtimeJavaVersion >= JavaVersion.VERSION_21) {
  val java21: SourceSet = sourceSets.create("java21") {
    java {
      compileClasspath += sourceSets.main.get().output + sourceSets.test.get().output
      runtimeClasspath += sourceSets.main.get().output + sourceSets.test.get().output
      srcDir("src/test/java11")
    }
  }

  configurations[java21.implementationConfigurationName].extendsFrom(configurations.testImplementation.get())
  configurations[java21.runtimeOnlyConfigurationName].extendsFrom(configurations.testRuntimeOnly.get())

  dependencies {
    testImplementation("org.opensearch.test", "framework", opensearchVersion) {
      exclude(group = "org.hamcrest")
    }
    testImplementation("org.opensearch:opensearch-testcontainers:4.1.0")
    testImplementation("org.testcontainers:testcontainers:2.0.5")
  }

  tasks.named<JavaCompile>("compileJava21Java") {
    targetCompatibility = JavaVersion.VERSION_21.toString()
    sourceCompatibility = JavaVersion.VERSION_21.toString()
  }

  tasks.named<JavaCompile>("compileTestJava") {
    targetCompatibility = JavaVersion.VERSION_21.toString()
    sourceCompatibility = JavaVersion.VERSION_21.toString()
 }
 
 tasks.named<Test>("integrationTest") {
    testClassesDirs += java21.output.classesDirs
    classpath = sourceSets["java21"].runtimeClasspath
 }
  
  tasks.named<Test>("unitTest") {
    testClassesDirs += java21.output.classesDirs
    classpath = sourceSets["java21"].runtimeClasspath
 }
}
```

java-client-grpc/src/test/java11/org/opensearch/client/opensearch/integTest/grpc/AbstractGrpcIT.java
```diff
@@ -0,0 +1,118 @@
...
+ * <ul>
+ *   <li>{@link #grpcClient()} — OpenSearchClient using HybridTransport (bulk → gRPC, rest → REST)</li>
+ *   <li>{@link #getGrpcHost()} — gRPC hostname</li>
+ *   <li>{@link #getGrpcPort()} — gRPC mapped port</li>
+ * </ul>
+ * <p>
+ * All tests automatically skip if the server version is below 3.5.0 (gRPC not available).
+ */
+@ThreadLeakFilters(filters = TestcontainersThreadFilter.class)
+public abstract class AbstractGrpcIT extends OpenSearchJavaClientTestCase {
```
reta: This will not work unfortunately: `OpenSearchJavaClientTestCase`, `TestcontainersThreadFilter` are part of the `opensearch-java` tests, these classes are not exposed and are not available for use.

reta:
Thanks @dayana-j , good one, did a first pass, look very solid in general! Could you please also add OpenSearch 3.5.0 (or 3.6.0) to the test matrix here `.github/workflows/test-integration.yml` ? Thank you.

java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/GrpcTransport.java
dayana-j: Yes I will, thank you for pointing it out to me

java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/GrpcTransport.java
dayana-j: The k-nn endpoints were omitted because in this version, we are doing the Bulk implementation first and then bring in the implementation for k-nn and search queries in later.

java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/GrpcTransportOptions.java
dayana-j: The 10MB default matches the OpenSearch server's grpc.netty.max_msg_size default. The limit is coming from what I saw on the documentation page for the gRPC API: [https://docs.opensearch.org/latest/api-reference/grpc-apis/index/](https://docs.opensearch.org/latest/api-reference/grpc-apis/index/)

java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/HybridTransport.java
dayana-j: Agreed. I'll change the default behavior so that gRPC errors are propagated for supported endpoints (no silent fallback). The HybridTransport will still route unsupported operations to REST, but if a gRPC supported endpoint like Bulk fails due to connection issues, the error will be surfaced to the user. This way customers know immediately when their gRPC setup isn't working rather than unknowingly running on REST.

java-client-grpc/src/test/java11/org/opensearch/client/opensearch/integTest/grpc/AbstractGrpcIT.java
dayana-j: Thank you for point it out.

java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/GrpcChannelFactory.java
reta: @dayana-j apologies, why this comment was resolved  without being addressed? The general context here is to limit the public API surface to minimal required subset, This class is only used by `GrpcTransport`s implementations internally

java-client-grpc/src/test/java11/org/opensearch/client/opensearch/integTest/grpc/GrpcTestContainerRule.java
```diff
@@ -0,0 +1,135 @@
...
+ * </ul>
+ * <p>
+ * Sets system properties so tests can discover the cluster:
+ * <ul>
+ *   <li>{@code tests.rest.cluster} — host:port for REST (9200)</li>
+ *   <li>{@code tests.grpc.cluster} — host:port for gRPC (9400)</li>
+ *   <li>{@code https} — false (security disabled)</li>
+ * </ul>
+ */
+final class GrpcTestContainerRule extends ExternalResource {
```
reta:
```suggestion
final class OpenSearchGrpcTestContainerRule extends ExternalResource {
```

java-client-grpc/src/test/java11/org/opensearch/client/opensearch/integTest/grpc/OpenSearchGrpcTestContainerRule.java
```diff
@@ -0,0 +1,135 @@
...
+    private static final String CLUSTER_PROPERTY = "tests.rest.cluster";
+    private static final String GRPC_CLUSTER_PROPERTY = "tests.grpc.cluster";
+    private static final String HTTPS_PROPERTY = "https";
+
+    static final int GRPC_PORT = 9400;
+
+    private static OpenSearchContainer<?> container;
+
+    @Override
+    protected void before() {
```
reta, line 48: I believe we should be able to use `assumeThat` for skipping tests:

```suggestion
    protected void before() {
        assumeThat("OpenSearch should support gRPC", supportsGrpc(version));
        ....
    }
```

java-client-grpc/src/test/java11/org/opensearch/client/opensearch/integTest/grpc/GrpcTestContainerRule.java
```diff
@@ -0,0 +1,135 @@
...
+        String version = System.getProperty(VERSION_PROPERTY, "");
+        OpenSearchContainer<?> openSearch = hasText(image)
+            ? new OpenSearchContainer<>(image)
+            : new OpenSearchContainer<>(OpenSearchDockerImage.ofVersion(requiredVersion()));
+
+        // Disable security for plaintext testing
+        openSearch.withEnv("plugins.security.disabled", "true");
+
+        // Only enable gRPC transport for versions that support it (3.5.0+)
+        if (supportsGrpc(version)) {
```
reta: I believe we always need OpenSearch version that supports gRPC, right? We could easily skip the tests for older version using `assumeThat`  (see please https://github.com/opensearch-project/opensearch-java/pull/2062/changes#r3626125320)

java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/translation/FieldMappingUtil.java
```diff
@@ -0,0 +1,176 @@
...
+     * <p>
+     * For OK (0), the result field disambiguates:
+     * - "created" → 201
+     * - "updated", "deleted", "noop" → 200
+     *
+     * @param grpcCode the gRPC status code (0-16)
+     * @param result   the result string from the response item, may be null
+     * @return the equivalent HTTP status code
+     */
+    public static int grpcStatusToHttpStatus(int grpcCode, String result) {
```
reta: Please remove this method, it is used only in tests and nowhere else

reta:
Thank you @dayana-j , we are getting close, a few comments / leftovers

java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/GrpcChannelFactory.java
```diff
@@ -54,7 +54,7 @@ private GrpcChannelFactory() {
      * @param interceptors optional client interceptors (e.g., auth)
      * @return a configured ManagedChannel
      */
-    public static ManagedChannel createPlaintextChannel(
+    private static ManagedChannel createPlaintextChannel(
```
reta: Some tests broke (sorry missed that):

> Error: /home/runner/work/opensearch-java/opensearch-java/java-client-grpc/src/test/java/org/opensearch/client/transport/grpc/GrpcChannelTest.java:111: error: createTlsChannel(String,int,GrpcTlsConfig,GrpcTransportOptions,List<ClientInterceptor>) has private access in GrpcChannelFactory
        channel = GrpcChannelFactory.createTlsChannel(

 I think we could relax to package private:

```suggestion
static ManagedChannel createPlaintextChannel(
```

java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/GrpcChannelFactory.java
```diff
@@ -79,7 +79,7 @@ public static ManagedChannel createPlaintextChannel(
      * @return a configured ManagedChannel with TLS
      * @throws IOException if SSL context creation fails
      */
-    public static ManagedChannel createTlsChannel(
+    private static ManagedChannel createTlsChannel(
```
reta:
```suggestion
static ManagedChannel createTlsChannel(
```

reta (approved):
Thank you @dayana-j !

dayana-j:
Thank you @reta

PR #2071, opened 2026-07-29
Add gRPC supports for Search foundation; match-all query

dayana-j:
### Description
This PR adds gRPC support for Search for the opensearch-java client, starting with the match_all query type. It follows the same architecture pattern used for the Bulk gRPC transport.

The SearchRequestConverter converts the opensearch-java SearchRequest to the protobuf SearchRequest used by the gRPC SearchService. Currently only match_all is implemented:
  ```
private static org.opensearch.protobufs.MatchAllQuery convertMatchAll(MatchAllQuery matchAll) {
      org.opensearch.protobufs.MatchAllQuery.Builder builder =
          org.opensearch.protobufs.MatchAllQuery.newBuilder();
      if (matchAll.boost() != null) {
          builder.setBoost(matchAll.boost().floatValue());
      }
      return builder.build();
  }
```
Other query types supported by the server (term, match, bool, range, k-NN, etc.) are not yet supported in this version and will throw an UnsupportedOperationException directing users to REST:
 ```
 throw new UnsupportedOperationException(
      "Query type '" + query._kind() + "' is not yet supported for gRPC transport in this version. "
          + "Use REST transport for this query type."
  );
```
The SearchResponseConverter converts the protobuf SearchResponse back to the opensearch-java SearchResponse<TDocument>. This includes full response metadata (took, timed_out, _shards), hits metadata (total, max_score), and per-hit fields. Document _source is deserialized from raw bytes to the user's target class via the JsonpMapper:
  ```
static <TDocument> TDocument deserializeSource(
      byte[] sourceBytes, JsonpMapper jsonpMapper, Class<TDocument> tDocumentClass
  ) {
      InputStream stream = new ByteArrayInputStream(sourceBytes);
      jakarta.json.stream.JsonParser parser = jsonpMapper.jsonProvider().createParser(stream);
      parser.next();
      return jsonpMapper.deserialize(parser, tDocumentClass);
  }
```
Additionally, the endpoint support was enhanced with a request-aware check (GrpcTransport.isEndpointSupported(endpoint, request)) as suggested by maintainer feedback. The HybridTransport now inspects the request content before routing: unsupported query types (term, match, bool, etc.) are automatically routed to REST instead of throwing an error, while match_all queries go over gRPC.
 ```
 // match_all → routed to gRPC
  client.search(s -> s.index("idx").query(q -> q.matchAll(m -> m)), Doc.class);

  // term query → routed to REST automatically
  client.search(s -> s.index("idx").query(q -> q.term(t -> t.field("status").value("active"))), Doc.class);
```

Unit tests cover the request conversion for match_all with various parameters, verify that unsupported query types are correctly identified by the request-aware routing, and confirm proper routing behavior. Integration tests index documents via gRPC bulk, then search with match_all over gRPC and verify the full response.


java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/GrpcTransport.java
```diff
@@ -52,7 +55,8 @@ public class GrpcTransport implements OpenSearchTransport {
     static {
         java.util.Set<Endpoint<?, ?, ?>> endpoints = new java.util.HashSet<>();
         endpoints.add(BulkRequest._ENDPOINT);
-        // Future: SearchRequest._ENDPOINT, KnnSearchRequest._ENDPOINT
+        endpoints.add(SearchRequest._ENDPOINT);
```
reta, line 58: Thanks @dayana-j , I think this change needs a bit more thoughts: 
 - the GRPC transport support is strictly endpoint base (for now), however the `_search` is limited to  only `match_all` query (it is not enough to filter endpoint, we need params or/and payload)
 - the hybrid transport does internal routing based on endpoints as well, so we basically end up with the situation that user could only invoke `match_all` like queries even if REST based transport supports everything else

That would significantly limit its applicability I believe. The suggestion I have is to enhance endpoint support with `request` argument `GrpcTransport.isEndpointSupported(endpoint, request)` so the GRPC transport could check endpoint and perform the request conversion (yes, could be a bit wasteful work but at least we will properly route to REST transport till all queries are covered). 

Curious WDYT? Thank you.

dayana-j: Thanks @reta, I did not think about adding in enhanced endpoint support to help do search. Your comment allowed me to think more about it and made me realize that params will be needed for search requests. I added the enhance endpoint support to check both the endpoint and the request content before routing since only the match-all query will be supported now.

reta (approved):

dayana-j:
Thanks @reta

Issue #540, opened 2023-06-24
[FEATURE] Support for composing query variants from 3rd party plugins


jhinch:
### Is your feature request related to a problem?

Some plugins such as the [OpenSearch LTR plugin](https://github.com/opensearch-project/opensearch-learning-to-rank-base) add additional query variants https://elasticsearch-learning-to-rank.readthedocs.io/en/latest/searching-with-your-model.html#searching-with-ltr.

opensearch-java current cannot support construction of a query containing such a query variant as [QueryVariant](https://github.com/opensearch-project/opensearch-java/blob/main/java-client/src/main/java/org/opensearch/client/opensearch/_types/query_dsl/QueryVariant.java#L44) interface requiring a kind being specified and [Query.Kind](https://github.com/opensearch-project/opensearch-java/blob/main/java-client/src/main/java/org/opensearch/client/opensearch/_types/query_dsl/Query.java#L69) being a enum, not allowing for variants to be specified in a 3rd party client.

### What solution would you like?

There would be a couple possible solutions to this problem. Converting the enum to a type which allows for open extension might be one. Add a `custom` variant to the enum which itself allows for a flexible query construction might be another.

### What alternatives have you considered?

See above

### Do you have any additional context?

N/A

wbeckler:
Converting enum to another type would be a breaking change, right? Do you want to flesh out the "custom" idea a little?

rtiruveedulaarkin:
We have developed custom plugin which works with rest high level client library. We are unable to use this plugin with opensearch-java client library due to Aggregate.Kind/Aggregation.Kind being enum types and they are not extensible. Please add support for extending Aggregate.Kind/Aggregation.Kind types.

wumpus3k-if:
### Use Case

We also would like to use a custom, plugin-defined query type in a search with the OpenSearch Java client. That would allow us to move the corresponding code from the high-level REST client towards the Java client.

### Possible Solution

The Elasticsearch Java client supports this in the following way:
[...]

PR #2072, opened 2026-07-29
Feat/add support for custom kinds

acote-coveo:
### Description

This PR adds a `_Custom` variant kind to all discriminated tagged unions (e.g. `Aggregate`, `Suggest`, `Query`, `Property`, `Processor`, `Analyzer`, `TokenFilterDefinition`) so that plugin-provided types are captured as raw JSON instead of causing deserialization failures.

When the server returns a discriminator value that isn't natively modeled, the client stores it as `Kind._Custom` with the original type name accessible via `_customKind()` and the raw JSON body via `_custom()`. This enables users to work with plugin-defined aggregations, queries, processors, etc. without waiting for a client release that models the new type.

### How It Works

**Reading a custom variant:**
```java
for (Map.Entry<String, Aggregate> entry : response.aggregations().entrySet()) {
    Aggregate agg = entry.getValue();
    if (agg._isCustom()) {
        String typeName = agg._customKind();   // e.g. "my_plugin_agg"
        JsonData rawData = agg._custom();      // raw JSON body
    }
}
```

**Building a custom variant:**
```java
Aggregate custom = new Aggregate.Builder()
    ._custom("my_plugin_agg", JsonData.of(Map.of("score", 42)))
    .build();
```

The same `_isCustom()`, `_customKind()`, `_custom()`, and `Builder._custom(type, data)` methods are available on every discriminated tagged union type.

### Changes

**Infrastructure (`java-client/src/main/java`):**
- `JsonData._DESERIALIZER`: Fix three-arg `deserialize` overload to correctly handle a pre-consumed parser event.
- `TaggedUnion`: Add default `_customKind()` method returning null for built-in kinds.
- `ExternallyTaggedUnion`: Use `_customKind()` fallback in typed-keys serialization when `jsonValue()` is null; accept `unknownVariantCtor` for graceful deserialization of unknown types.

**Code generation templates (`java-codegen/`):**
- `TaggedUnionShape.mustache`: Add `Kind._Custom(null)`, `_customKind` field, accessors, `CustomVariant` inner class with equals/hashCode, null-validated builder method, and updated equals/hashCode on the union itself.
- `Deserialize.mustache`: Wire `unknownFieldHandler` (internally-tagged) and `unknownVariantCtor` (externally-tagged).
- `Serialize.mustache`: Use `_customKind` when `Kind._Custom` for the JSON key.

**Documentation:** Added "Custom (Plugin-Provided) Variant Types" section to `guides/json.md`.

**Tests:** Added `CustomVariantTest` covering deserialization, typed-keys round-trip serialization, internally-tagged round-trip, builder API, and equality semantics.

### Related

This approach aligns with the Elasticsearch Java client's `OpenTaggedUnion` pattern, adapted for OpenSearch's typed variant interfaces.

On ElasticSearch, they specify whether an Tagged Union should be open or not. I took the approach to extend everything that is discriminated to offer as much flexibility as allowed by the plugins. 

### Issues Resolved

Closes:

- https://github.com/opensearch-project/opensearch-java/issues/540

### DIFF
--- a/java-client/src/generated/java/org/opensearch/client/opensearch/_types/aggregations/Aggregate.java
+++ b/java-client/src/generated/java/org/opensearch/client/opensearch/_types/aggregations/Aggregate.java
@@ -44,6 +44,7 @@ import java.util.function.Function;
 import javax.annotation.Generated;
 import javax.annotation.Nonnull;
 import org.opensearch.client.json.ExternallyTaggedUnion;
+import org.opensearch.client.json.JsonData;
 import org.opensearch.client.json.JsonEnum;
 import org.opensearch.client.json.JsonpDeserializer;
 import org.opensearch.client.json.JsonpMapper;
@@ -122,7 +123,9 @@ public class Aggregate implements TaggedUnion<Aggregate.Kind, AggregateVariant>,
         Umterms("umterms"),
         ValueCount("value_count"),
         VariableWidthHistogram("variable_width_histogram"),
-        WeightedAvg("weighted_avg");
+        WeightedAvg("weighted_avg"),
+        /** A custom variant type not natively supported by this client. */
+        _Custom(null);
 
         private final String jsonValue;
 
@@ -138,6 +141,7 @@ public class Aggregate implements TaggedUnion<Aggregate.Kind, AggregateVariant>,
 
     private final Kind _kind;
     private final AggregateVariant _value;
+    private final String _customKind;
 
     @Override
     public final Kind _kind() {
@@ -149,14 +153,23 @@ public class Aggregate implements TaggedUnion<Aggregate.Kind, AggregateVariant>,
         return _value;
     }
 
+    /**
+     * Returns the actual type name when {@code _kind() == Kind._Custom}, otherwise {@code null}.
+     */
+    public final String _customKind() {
+        return _customKind;
+    }
+
     public Aggregate(AggregateVariant value) {
         this._kind = ApiTypeHelper.requireNonNull(value._aggregateKind(), this, "<variant kind>");
         this._value = ApiTypeHelper.requireNonNull(value, this, "<variant value>");
+        this._customKind = null;
     }
 
     private Aggregate(Builder builder) {
         this._kind = ApiTypeHelper.requireNonNull(builder._kind, builder, "<variant kind>");
         this._value = ApiTypeHelper.requireNonNull(builder._value, builder, "<variant value>");
+        this._customKind = builder._customKind;
     }
 
     public static Aggregate of(Function<Aggregate.Builder, ObjectBuilder<Aggregate>> fn) {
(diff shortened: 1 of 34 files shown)


java-client/src/generated/java/org/opensearch/client/opensearch/_types/aggregations/Aggregate.java
```diff
@@ -122,7 +123,9 @@ public enum Kind implements JsonEnum {
         Umterms("umterms"),
         ValueCount("value_count"),
         VariableWidthHistogram("variable_width_histogram"),
-        WeightedAvg("weighted_avg");
+        WeightedAvg("weighted_avg"),
+        /** A custom variant type not natively supported by this client. */
+        _Custom(null);
```
reta, line 128: Thanks for the pull request @acote-coveo , mostly all code for the Java client is being generated (this is why every source file has `@Generated("org.opensearch.client.codegen.CodeGenerator")` annotation and the comment on top).

Please submit the change to the OpenSearch OpenAPI specification so it could be picked up automatically: https://github.com/opensearch-project/opensearch-api-specification/blob/main/spec/schemas/_common.aggregations.yaml#L9

reta (changes requested):

java-client/src/generated/java/org/opensearch/client/opensearch/_types/aggregations/Aggregate.java
acote-coveo: @reta Thanks for the quick feedback. I just wanted to mention that I didn't explicitly modify the generated, I instead customized the generator template because I assumed that this kind of customization should not be part of the API specification. To me, the API specification should describe what OpenSearch supports. The custom kind that I added is a client-side concern to allow customization over the typed DSL. 

For example: https://github.com/opensearch-project/opensearch-java/pull/2072/changes#diff-93745660477f5c12415912b6c105074b530c83c722b2771390de7f4e9b458474R6-R8

Let me know with my answer if you still think this should be document at the API level.

acote-coveo: If I were to add this to the API Specification, I would probably add a new custom annotation to mark the variants that are non-exhaustive and from the generator, I would look for this tag to know if I should generate the _custom variant code I just added. The main benefits would be that:

1. We could control which variants are open or not.
2. Other generators would have the same signals to generate exhaustive vs non-exhaustive variants.

reta: > @reta Thanks for the quick feedback. I just wanted to mention that I didn't explicitly modify the generated, I instead customized the generator template because I assumed that this kind of customization should not be part of the API specification.

My apologies  @acote-coveo , jumped into quick conclusion. The support of custom types (queries, aggregations, etc) is a long standing one indeed. However, not every component does support such extensibility, I think we could combine specification approach (fe add `X-Extensible: true/false` attribute to the schema) and the generator changes that you are are suggesting,. It will at least bring clarify what could be customized, hope it make sense,. 

Thank you.

acote-coveo: @reta Created a new pr here: https://github.com/opensearch-project/opensearch-api-specification/pull/1182/changes

How it would be used in this repo: https://github.com/opensearch-project/opensearch-java/pull/2083

I'll close this one.

PR #2083, opened 2026-08-03
Feat/add support for non exhaustive union types

acote-coveo:
### Description

OpenSearch plugins can introduce aggregation types, query types, processors, analyzers, and mapping properties that aren't modeled in the Java client's generated code. Today, encountering one of these at deserialization time throws a `JsonException`, making the client unusable against clusters with such plugins installed.

This PR introduces a `Kind._Custom` enum member on tagged unions marked `x-non-exhaustive: true` in the OpenAPI spec. When the deserializer encounters an unknown discriminator value it captures the raw JSON as `JsonData` under the `_Custom` kind instead of failing. The original type name is preserved via `_customKind()` and the data round-trips through serialization.

### Changes

**Bug-fix (independent, first commit):**

- `JsonData._DESERIALIZER` — the lambda-based deserializer silently ignored the pre-consumed `event` argument in the three-arg `deserialize` overload. Replaced with an explicit `JsonpDeserializerBase` that delegates to `parser.getValue()` when the event is already consumed.

**Feature:**

| Area | What changed |
|------|--------------|
| OpenAPI spec | Added `x-non-exhaustive: true` to 10 union schemas (Aggregate, Query, SpanQuery, Property, Processor, Analyzer, CharFilterDefinition, TokenFilterDefinition, TokenizerDefinition, FieldSuggester) |
| Codegen model | `OpenApiSchema` parses the extension; `SpecTransformer` propagates it; `TaggedUnionShape` gains a `nonExhaustive` flag and a `_customKind` field |
| Codegen templates | `TaggedUnionShape.mustache` emits `Kind._Custom`, `_customKind()`, `_isCustom()`, `_custom()`, `Builder._custom(type, data)`, and an inner `CustomVariant` wrapper class. Deserialize/Serialize templates handle the fallback path |
| Runtime | `TaggedUnion` interface adds a default `_customKind()` method; `ExternallyTaggedUnion` uses it for typed-key serialization |
| Generated code | 9 union classes regenerated with the new support |

### Usage

```java
// Reading an unknown aggregation type returned by a plugin
Aggregate agg = response.aggregations().get("foo");
if (agg._isCustom()) {
    String typeName = agg._customKind();   // e.g. "my_plugin_agg"
    JsonData raw = agg._custom();           // full JSON body
}

// Building a custom variant
Aggregate custom = new Aggregate.Builder()
    ._custom("my_plugin_agg", JsonData.of(Map.of("score", 42)))
    .build();
```

### Related changes

https://github.com/opensearch-project/opensearch-api-specification/pull/1182

### Checklist

- [x] New functionality is covered by tests
- [x] `CHANGELOG.md` updated
- [x] User guide updated (`guides/json.md`)
- [ ] CI passes

### DIFF
--- a/java-client/src/generated/java/org/opensearch/client/opensearch/_types/query_dsl/Query.java
+++ b/java-client/src/generated/java/org/opensearch/client/opensearch/_types/query_dsl/Query.java
@@ -125,7 +125,9 @@ public class Query implements TaggedUnion<Query.Kind, Object>, AggregationVarian
         Type("type"),
         Wildcard("wildcard"),
         Wrapper("wrapper"),
-        XyShape("xy_shape");
+        XyShape("xy_shape"),
+        /** A custom variant type not natively supported by this client. */
+        _Custom(null);
 
         private final String jsonValue;
 
@@ -149,6 +151,7 @@ public class Query implements TaggedUnion<Query.Kind, Object>, AggregationVarian
 
     private final Kind _kind;
     private final Object _value;
+    private final String _customKind;
 
     @Override
     public final Kind _kind() {
@@ -160,14 +163,23 @@ public class Query implements TaggedUnion<Query.Kind, Object>, AggregationVarian
         return _value;
     }
 
+    /**
+     * Returns the actual type name when {@code _kind() == Kind._Custom}, otherwise {@code null}.
+     */
+    public final String _customKind() {
+        return _customKind;
+    }
+
     public Query(QueryVariant value) {
         this._kind = ApiTypeHelper.requireNonNull(value._queryKind(), this, "<variant kind>");
         this._value = ApiTypeHelper.requireNonNull(value, this, "<variant value>");
+        this._customKind = null;
     }
 
     private Query(Builder builder) {
         this._kind = ApiTypeHelper.requireNonNull(builder._kind, builder, "<variant kind>");
         this._value = ApiTypeHelper.requireNonNull(builder._value, builder, "<variant value>");
+        this._customKind = builder._customKind;
     }
 
     public static Query of(Function<Query.Builder, ObjectBuilder<Query>> fn) {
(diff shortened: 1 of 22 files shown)


java-client/src/generated/java/org/opensearch/client/opensearch/_types/query_dsl/Query.java
```diff
@@ -1720,6 +1753,19 @@ public ObjectBuilder<Query> xyShape(Function<XyShapeQuery.Builder, ObjectBuilder
             return this.xyShape(fn.apply(new XyShapeQuery.Builder()).build());
         }
 
+        /**
+         * Set a custom (plugin-provided) variant.
+         *
+         * @param type the variant type name as returned by the server
+         * @param data the raw JSON body of the variant result
+         */
+        public ObjectBuilder<Query> _custom(String type, JsonData data) {
```
russcam, line 1762: `JsonData` is a pretty wide type to expose in the client/SDK.

In each case where a custom type is supported, is it possible to constrain the type(s) to the base type from which built-in components inherit/implement? For example here, constrain this to `extends QueryBase & QueryVariant`? In the case of a query, it means it still has a `name()` and `boost()`

I see there is some challenge to resolve in accessing the deserializer for the implementation in doing this.

acote-coveo: I went with `JsonData` because:

1. It offers the most flexibility
2. It integrates well with the serializer/deserializer mechanisms.
3. It's the pattern used for ElasticSearch clients (e.g: https://github.com/elastic/elasticsearch-java/blob/8be02330d48758e2708a00f18cc87587df2a2520/java-client/src/main/java/co/elastic/clients/elasticsearch/_types/aggregations/Aggregate.java#L282)

What are the concerns that you see with this approach versus limit to `QueryBase & QueryVariant` instance for example?

---

I'm open to rework the approach and have more restrictive types if that's what the maintainers prefers but I would like to have some consensus before spending further time on this. 

Supporting `QueryBase & QueryVariant` would require to open the deserializers configurations to allow registering new types.

PR #2085, opened 2026-08-03
Run model tests against both JSON mappers instead of picking one at r…

### LINKED ISSUE #770: [PROPOSAL] ModelTestCase.mapper improvements
### What are you proposing?
- Modify the ModelTestCase model's mapper to use only one fixedly or both to use both fixedly.
image

### What problems are you trying to solve?
- I don't think it's good code if a test passes or breaks randomly. It seems like a good idea to organize the reasons for using two mappers and improve it so that they are always used when both are needed.

### What is the developer experience going to be?
- No more cases of partially passing or failing tests

loggk:
### Description

`ModelTestCase` chooses between `JsonbJsonpMapper` and `JacksonJsonpMapper` with a single `Random` coin flip per JVM, so each test run exercises one serialization backend and never the other, chosen at random. A bug affecting only one mapper passes CI about half the time.

### Changes

- Replace the random mapper selection in `ModelTestCase` with JUnit 4 `Parameterized`, running every test against both mappers. `@RunWith` is inherited and the mapper is injected by field, so none of the 104 subclasses need changes.

Verified with `./gradlew :java-client:unitTest`: 426 tests become 604, with exactly 178 per mapper and 0 failures. Runs are now deterministic, and failures report which mapper broke (`testName[json-b]` / `testName[jackson]`). `spotlessJavaCheck` passes.

### Issues Resolved

Resolves #770

### Check List

- [x] All tests pass
- [x] Commits are signed per the DCO using `--signoff`
- [x] Changelog updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Fixed
 - Fix `unitTest` task not running the tests in the `test` source set ([#2074](https://github.com/opensearch-project/opensearch-java/pull/2074))
+- Run model tests against both JSON mappers instead of picking one at random ([#2085](https://github.com/opensearch-project/opensearch-java/pull/2085))
 
 ## [Unreleased 3.x]
 ### Added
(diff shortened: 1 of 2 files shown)


reta (commented):
Thanks @loggk !

reta (approved):

PR #2086, opened 2026-08-03
Add setRetryStrategy to ApacheHttpClient5TransportBuilder

### LINKED ISSUE #1738: [BUG] Retries disabled in HTTP client builder.
### What is the bug?
HTTP client transport builder specifically [disables retries](https://github.com/opensearch-project/opensearch-java/blob/2.15/java-client/src/main/java/org/opensearch/client/transport/httpclient5/ApacheHttpClient5TransportBuilder.java#L360). This requires workaround to be created in the business layer, which may not always be the most elegant way to perform these retries. Allowing the client builder the ability to provide a custom retry strategy will provide a more robust and efficient client. Not everybody is going to use retry strategies just because there is an ability to do so - so I think it should be left to the discretion of the software architect on whether to use a retry strategy and how it should be defined, rather than outright disabling the ability to retry at all.

Create a new HTTP transport.
Set a custom or default retry strategy in the builder.
None of the overridden retry request methods ever get called.

Allow retry strategies to be defined if they are deemed as required. 

Code running in an opensearch container. Running version 2.15.0.

_Add any other context about the problem._

loggk:
### Description
The builder unconditionally calls `disableAutomaticRetries()` on the underlying `HttpAsyncClientBuilder`. In HttpClient 5 this flag cannot be undone, so a retry strategy set through `HttpClientConfigCallback` is silently ignored and there is no supported way to enable retries.

This adds a `setRetryStrategy(HttpRequestRetryStrategy)` option to the builder. When set, the strategy is passed through to the http client; when not set, retries stay disabled as before, keeping the default behavior on par with the 4.x http client.

A new `RetryStrategyTest` verifies both behaviors against a local socket server that fails the first request with a 503: with a retry strategy the request is transparently retried and succeeds, and by default the failure surfaces without a retry.

### Issues Resolved
Fixes #1738 

- [x] New functionality includes testing
- [x] API changes companion pull request created, if applicable
- [x] Commits are signed per the DCO using --signoff
- [x] Public documentation issue/PR created, if applicable

### DIFF
--- a/java-client/src/main/java/org/opensearch/client/transport/httpclient5/ApacheHttpClient5TransportBuilder.java
+++ b/java-client/src/main/java/org/opensearch/client/transport/httpclient5/ApacheHttpClient5TransportBuilder.java
@@ -395,7 +395,7 @@ public class ApacheHttpClient5TransportBuilder {
                 .setTargetAuthenticationStrategy(DefaultAuthenticationStrategy.INSTANCE);
             if (automaticRetriesDisabled) {
                 // Keep behavior on par with the 4.x http client, which had no automatic retries
-                httpClientBuilder =  httpClientBuilder.disableAutomaticRetries();
+                httpClientBuilder = httpClientBuilder.disableAutomaticRetries();
             }
             if (httpClientConfigCallback != null) {
                 httpClientBuilder = httpClientConfigCallback.customizeHttpClient(httpClientBuilder);
--- a/java-client/src/main/java/org/opensearch/client/transport/httpclient5/ApacheHttpClient5TransportBuilder.java
+++ b/java-client/src/main/java/org/opensearch/client/transport/httpclient5/ApacheHttpClient5TransportBuilder.java
@@ -395,7 +395,7 @@ public class ApacheHttpClient5TransportBuilder {
                 .setTargetAuthenticationStrategy(DefaultAuthenticationStrategy.INSTANCE);
             if (automaticRetriesDisabled) {
                 // Keep behavior on par with the 4.x http client, which had no automatic retries
-              httpClientBuilder =  httpClientBuilder.disableAutomaticRetries();
+                httpClientBuilder =  httpClientBuilder.disableAutomaticRetries();
             }
             if (httpClientConfigCallback != null) {
                 httpClientBuilder = httpClientConfigCallback.customizeHttpClient(httpClientBuilder);
--- a/java-client/src/main/java/org/opensearch/client/transport/httpclient5/ApacheHttpClient5TransportBuilder.java
+++ b/java-client/src/main/java/org/opensearch/client/transport/httpclient5/ApacheHttpClient5TransportBuilder.java
@@ -395,7 +395,7 @@ public class ApacheHttpClient5TransportBuilder {
                 .setTargetAuthenticationStrategy(DefaultAuthenticationStrategy.INSTANCE);
             if (automaticRetriesDisabled) {
                 // Keep behavior on par with the 4.x http client, which had no automatic retries
-                httpClientBuilder.disableAutomaticRetries();
+              httpClientBuilder =  httpClientBuilder.disableAutomaticRetries();
             }
             if (httpClientConfigCallback != null) {
                 httpClientBuilder = httpClientConfigCallback.customizeHttpClient(httpClientBuilder);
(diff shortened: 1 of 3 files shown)


java-client/src/main/java/org/opensearch/client/transport/httpclient5/ApacheHttpClient5TransportBuilder.java
```diff
@@ -72,6 +73,7 @@ public class ApacheHttpClient5TransportBuilder {
     private HttpClientConfigCallback httpClientConfigCallback;
     private RequestConfigCallback requestConfigCallback;
     private ConnectionConfigCallback connectionConfigCallback;
+    private HttpRequestRetryStrategy retryStrategy;
```
reta: I think our best option here is to mirror `HttpAsyncClientBuilder` (but with different default):

```suggestion
    private boolean automaticRetriesDisabled = true;
```

And call `disableAutomaticRetries()` depending on this flag. Than `HttpClientConfigCallback` could be used to set whatever retry policy is required (or keep the default).

reta (changes requested):

reta (commented):
Thanks @loggk !

java-client/src/main/java/org/opensearch/client/transport/httpclient5/ApacheHttpClient5TransportBuilder.java
```diff
@@ -378,8 +392,11 @@ public TlsDetails create(final SSLEngine sslEngine) {
             HttpAsyncClientBuilder httpClientBuilder = HttpAsyncClientBuilder.create()
                 .setDefaultRequestConfig(requestConfigBuilder.build())
                 .setConnectionManager(connectionManager)
-                .setTargetAuthenticationStrategy(DefaultAuthenticationStrategy.INSTANCE)
-                .disableAutomaticRetries();
+                .setTargetAuthenticationStrategy(DefaultAuthenticationStrategy.INSTANCE);
+            if (automaticRetriesDisabled) {
+                // Keep behavior on par with the 4.x http client, which had no automatic retries
+                httpClientBuilder.disableAutomaticRetries();
```
reta:
```suggestion
              httpClientBuilder =  httpClientBuilder.disableAutomaticRetries();
```

java-client/src/main/java/org/opensearch/client/transport/httpclient5/ApacheHttpClient5TransportBuilder.java
```diff
@@ -378,8 +392,11 @@ public TlsDetails create(final SSLEngine sslEngine) {
             HttpAsyncClientBuilder httpClientBuilder = HttpAsyncClientBuilder.create()
                 .setDefaultRequestConfig(requestConfigBuilder.build())
                 .setConnectionManager(connectionManager)
-                .setTargetAuthenticationStrategy(DefaultAuthenticationStrategy.INSTANCE)
-                .disableAutomaticRetries();
+                .setTargetAuthenticationStrategy(DefaultAuthenticationStrategy.INSTANCE);
+            if (automaticRetriesDisabled) {
+                // Keep behavior on par with the 4.x http client, which had no automatic retries
+              httpClientBuilder =  httpClientBuilder.disableAutomaticRetries();
```
reta:
```suggestion
                httpClientBuilder =  httpClientBuilder.disableAutomaticRetries();
```

reta (dismissed):

java-client/src/main/java/org/opensearch/client/transport/httpclient5/ApacheHttpClient5TransportBuilder.java
```diff
@@ -378,8 +392,11 @@ public TlsDetails create(final SSLEngine sslEngine) {
             HttpAsyncClientBuilder httpClientBuilder = HttpAsyncClientBuilder.create()
                 .setDefaultRequestConfig(requestConfigBuilder.build())
                 .setConnectionManager(connectionManager)
-                .setTargetAuthenticationStrategy(DefaultAuthenticationStrategy.INSTANCE)
-                .disableAutomaticRetries();
+                .setTargetAuthenticationStrategy(DefaultAuthenticationStrategy.INSTANCE);
+            if (automaticRetriesDisabled) {
+                // Keep behavior on par with the 4.x http client, which had no automatic retries
+                httpClientBuilder =  httpClientBuilder.disableAutomaticRetries();
```
reta:
```suggestion
                httpClientBuilder = httpClientBuilder.disableAutomaticRetries();
```

reta (approved):

PR #2088, opened 2026-08-07
fix: allow indexed_shape-only geo/xy shape queries

### LINKED ISSUE #2011: [BUG] "shape" should not be required in GeoShapeQueryField
### What is the bug?
The current opesearch-openapi.yaml contias following definition:

```
    _common.query_dsl___GeoShapeQueryField:
      type: object
      properties:
        indexed_shape:
          $ref: '#/components/schemas/_common.query_dsl___FieldLookup'
        shape:
          $ref: '#/components/schemas/_common.query_dsl___GeoShape'
        relation:
          $ref: '#/components/schemas/_common___GeoShapeRelation'
      required:
        - shape
```

This is not correct to require "shape", it's not required in the API - https://docs.opensearch.org/latest/query-dsl/geo-and-xy/geoshape/#using-a-pre-indexed-shape-definition

This makes it impossible to do pre-indexed shape queries right now using query DSL generated classes.

```java
new Query.Builder()
    .geoShape(GeoShapeQuery.Builder()
        .field("some")
        .shape(shape -> shape.indexedShape(i -> i.index("some").id("id").path("path")))
        .build()
    )
    .build();
```

I expect it to be possible to do pre-indexed shape queries.

Any

N/A

N/A

arimu1:
### Description

Fixes #2011.

`GeoShapeQueryField` (and the sibling `XyShapeQueryField`) currently mark `shape` as required. OpenSearch documents that a geo/xy shape query may use either an inline `shape` **or** a pre-indexed shape via `indexed_shape` (not both). Because the generated builder enforces `shape`, callers cannot construct valid pre-indexed shape queries:

```java
new Query.Builder()
    .geoShape(g -> g
        .field("location")
        .shape(s -> s.indexedShape(i -> i.index("shapes").id("id").path("location")))
    )
    .build();
```

### Changes

- Remove `required: [shape]` from `_common.query_dsl___GeoShapeQueryField` and `_common.query_dsl___XyShapeQueryField` in `java-codegen/opensearch-openapi.yaml`
- Update generated `GeoShapeQueryField` / `XyShapeQueryField` so `shape` is `@Nullable` (no `ApiTypeHelper.requireNonNull`, conditional serialization, null-safe equals/hashCode)
- Add unit coverage for `indexed_shape`-only construction/serialization
- CHANGELOG entry under Unreleased 4.x Fixed

### Related

- Upstream API-spec discussion: opensearch-project/opensearch-api-specification#1160 / #1161 (same schema correction). This client PR unblocks users of the generated DSL until the next full spec refresh.

### Testing

```
./gradlew :java-client:test \
  --tests org.opensearch.client.opensearch._types.query_dsl.GeoShapeQueryFieldTest \
  --tests org.opensearch.client.opensearch._types.query_dsl.XyShapeQueryFieldTest
```

Temurin 21 / macOS — 4 tests, 0 failures (`toBuilder` + `indexedShapeOnly` for both classes).

### Check List

- [x] New functionality includes testing
- [x] New functionality has been documented if needed (N/A — aligns client with existing API docs)
- [x] API changes companion issue/PR for [opensearch-api-specification](https://github.com/opensearch-project/opensearch-api-specification) (prior attempt #1161)
- [x] Commits are signed per DCO with `Signed-off-by`

### DIFF
--- a/java-codegen/opensearch-openapi.yaml
+++ b/java-codegen/opensearch-openapi.yaml
@@ -48620,8 +48620,6 @@ components:
           $ref: '#/components/schemas/_common.query_dsl___GeoShape'
         relation:
           $ref: '#/components/schemas/_common___GeoShapeRelation'
-      required:
-        - shape
     _common.query_dsl___GeoValidationMethod:
       type: string
       enum:
@@ -50318,8 +50316,6 @@ components:
           $ref: '#/components/schemas/_common.query_dsl___XyShape'
         relation:
           $ref: '#/components/schemas/_common___GeoShapeRelation'
-      required:
-        - shape
     _common.query_dsl___ZeroTermsQuery:
       type: string
       enum:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Fixed
 - Fix `unitTest` task not running the tests in the `test` source set ([#2074](https://github.com/opensearch-project/opensearch-java/pull/2074))
+- Allow pre-indexed shape queries by making `shape` optional on `GeoShapeQueryField` and `XyShapeQueryField` ([#2011](https://github.com/opensearch-project/opensearch-java/issues/2011))
 
 ## [Unreleased 3.x]
 ### Added
--- a/java-client/src/generated/java/org/opensearch/client/opensearch/_types/query_dsl/GeoShapeQueryField.java
+++ b/java-client/src/generated/java/org/opensearch/client/opensearch/_types/query_dsl/GeoShapeQueryField.java
@@ -49,7 +49,6 @@ import org.opensearch.client.json.ObjectBuilderDeserializer;
 import org.opensearch.client.json.ObjectDeserializer;
 import org.opensearch.client.json.PlainJsonSerializable;
 import org.opensearch.client.opensearch._types.GeoShapeRelation;
-import org.opensearch.client.util.ApiTypeHelper;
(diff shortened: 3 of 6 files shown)


java-codegen/opensearch-openapi.yaml
```diff
@@ -48620,8 +48620,6 @@ components:
           $ref: '#/components/schemas/_common.query_dsl___GeoShape'
         relation:
           $ref: '#/components/schemas/_common___GeoShapeRelation'
-      required:
```
reta, line 48623: Thanks @arimu1 , please submit a pull request against https://github.com/OpenSearch-project/openSearch-api-specification

reta (changes requested):

arimu1:
@reta Thanks for the review direction — understood that `indexed_shape` / required-`shape` must be fixed in the OpenAPI source of truth first ([OpenSearch-project/opensearch-api-specification](https://github.com/OpenSearch-project/opensearch-api-specification)), not as a hand-edit of generated client sources.

I'll open a PR against the API specification for the geo/xy shape oneOf (inline `shape` **or** `indexed_shape`) and follow the client regen path after that lands. Leaving this PR open as a temporary client-side illustration until the spec PR is up; happy to close #2088 if you prefer that sequencing.

reta:
> happy to close #2088 if you prefer that sequencing.

Thanks @arimu1 , you could close the pull request, the change in schema will be picked up

arimu1:
Closing as requested — the schema fix needs to land in [opensearch-api-specification](https://github.com/opensearch-project/opensearch-api-specification) first, then the client can pick it up via regeneration.

Happy to open a follow-up once the OpenAPI change is in.

PR #2098, opened 2026-08-11
Support named query scores in Hit.matchedQueries()

### LINKED ISSUE #1805: [FEATURE] Support `include_named_queries_score` to parsed scores of named queries

The OpenSearch Search API supports an `include_named_queries_score` parameter (introduced in OpenSearch 3.0 and backported to 2.x). When enabled, the `matched_queries` field in the response changes from a `List<String>` to a `Map<String, Double>`.

The Java client currently has no way to request this parameter, nor does `Hit.java` have a field to store the resulting scores. Furthermore, because `Hit.matched_queries` is strictly typed as `List<String>`, enabling this parameter manually via a raw request throws deserialization errors.

I would like to add support for requesting and accessing these scores without breaking backward compatibility for existing users.

The solution should:
- [ ] Enable setting the `include_named_queries_score` URL parameter.
- [ ] Enable setting the `include_named_queries_score` request body attribute.
- [ ] Add a new property to `Hit.java`: `@Nullable Map<String, Double> matched_query_scores`.

**Implementation Note:**
To maintain backward compatibility, the deserialization logic for `Hit` will need to handle the polymorphic nature of the `matched_queries` JSON field:
1. If JSON is `List<String>`: Populate existing `matched_queries` list; `matched_query_scores` remains null.
2. If JSON is `Map<String, Double>`: Populate the new `matched_query_scores` map **AND** populate the existing `matched_queries` list using the map keys. This ensures legacy code relying on the List continues to work seamlessly.

#### [Breaking change] Change `matched_queries` to `Map<String, Double>`

The cleanest API solution would be to change the type of the existing field. However, `Hit.matched_queries` is [currently exposed](https://github.com/opensearch-project/opensearch-java/blob/2ae7bf3aa6411aee3b865c39b5f848c6dcfc2b15/java-client/src/generated/java/org/opensearch/client/opensearch/core/search/Hit.java#L225) as a `List<String>`. Changing this would break binary and source compatibility for all users, which is likely unacceptable outside of a major version release.

#### Wrapper class

I considered introducing a wrapper class (e.g., `MatchedQueries`) to encapsulate the polymorphism. However, replacing the existing `List<String>` field with this wrapper is also a breaking change (same as above). Introducing a wrapper alongside the existing list seems overly complex compared to simply adding a Map field.

*   OpenSearch server support added in: https://github.com/opensearch-project/OpenSearch/pull/11626
*   Related discussion in `elasticsearch-java` (where they opted for a breaking change in a new major version): https://github.com/elastic/elasticsearch-java/issues/634

Given that OpenSearch 3.0 was released recently, waiting for a new major version of the client to introduce this feature seems undesirable. The proposed additive solution allows immediate support.

lawofcycles:
### Description
This change lets the client retrieve named query scores through the typed API. Hit.matchedQueries() is now generated as Map<String, Double> mapping each named query to its score, and include_named_queries_score is sent as a query parameter so the server returns the scored map.

On the response side, a schema driven rule in the code generator detects the oneOf of a string array and a scalar value map and maps it to a typed map. The deserializer reads the object form as name to score and the array form as names with null values. This replaces the generic JsonData fallback that the polymorphic matched_queries schema previously produced. The detection is generic and, against the current specification, applies only to Hit.matched_queries.

On the request side, include_named_queries_score is emitted as a query parameter. The request body property is ignored in the code generator on purpose, because the server enables the scored map rendering only from the query parameter. Sending the flag in the request body returns names without scores. See https://github.com/opensearch-project/OpenSearch/issues/22689 for that server behavior.

matched_queries is typed directly instead of adding a separate scores field. This is a breaking change scoped to the 4.0 line, where matched_queries was already exposed as JsonData rather than List<String>, so no existing typed usage breaks.

### Issues Resolved
Closes #1805


java-client/src/generated/java/org/opensearch/client/opensearch/core/search/Hit.java
```diff
@@ -516,7 +529,7 @@ public static class Builder<TDocument> extends ObjectBuilderBase implements Copy
         @Nullable
         private Map<String, InnerHitsResult> innerHits;
         @Nullable
-        private JsonData matchedQueries;
+        private Map<String, Double> matchedQueries;
```
reta: Thanks @lawofcycles , to be fair, I have some doubts this is the approach we should take: `matchedQueries` is either a map (when named) or array, however we just flush it out as a map with `null` scores - not very intuitive and asking for bug reports. I will spend a bit more time looking into other options, if any, thank you.

lawofcycles: @reta  Fair concern on the null valued map. I see two typed shapes that avoid it.

One is a TaggedUnion of the array and map forms, in the spirit of SourceConfig or FieldValue, bound through withMappedType. The other keeps matchedQueries() as List<String> and adds a separate matchedQueryScores() as Map<String, Double> that is filled only when scores are requested.

The split field reads a little more naturally for callers, but the union fits the existing code generation much better, since it reuses the withMappedType path with no changes to the object templates, while the split field would need one wire key to populate two fields on both read and write. I am leaning toward the union for that reason. Let me know which direction you prefer, or if you have another shape in mind.

reta: > I am leaning toward the union for that reason. Let me know which direction you prefer, or if you have another shape in mind.

Thanks @lawofcycles , I would agree with you that TaggedUnion would feel more natural (and in spirit of other types).

lawofcycles: @reta Reworked matched_queries into a MatchedQueries tagged union with Names (List<String>) and Scores (Map<String, Double>) variants, bound through Overrides.PropertyOverride.withMappedType and using UnionDeserializer for the array/object split, following the SourceConfig and FieldValue pattern.

reta (approved):

lawofcycles:
Thank you for reviewing!

PR #2100, opened 2026-08-16
[Backport 3.x] Support named query scores in Hit.matchedQueries()

### LINKED ISSUE #1805: [FEATURE] Support `include_named_queries_score` to parsed scores of named queries

lawofcycles:
### Description
Adds typed access to named query scores on `3.x`. `Hit.matched_queries` becomes a `MatchedQueries` tagged union carrying either names (`List<String>`) or scores (`Map<String, Double>`), and `include_named_queries_score` is sent as a query parameter so the server actually returns the scored map (previously it was placed in the request body, which the server does not honor: https://github.com/opensearch-project/OpenSearch/issues/22689).

Source-breaking on `3.x`, since `matched_queries()` now returns `MatchedQueries` instead of `List<String>`.

### Issues Resolved
#1805 (backport of #2098 to `3.x`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Bump `org.apache.httpcomponents.client5:httpclient5` from 5.6 to 5.6.1 ([#1967](https://github.com/opensearch-project/opensearch-java/pull/1967))
 
 ### Changed
+- Breaking: Change typed shape of `Hit.matched_queries` to a `MatchedQueries` tagged union that carries either a `List<String>` (names) or a `Map<String, Double>` (scores), and send `include_named_queries_score` as a query parameter ([#2098](https://github.com/opensearch-project/opensearch-java/pull/2098))
 
 ### Deprecated
 
--- a/java-client/src/generated/java/org/opensearch/client/opensearch/core/search/Hit.java
+++ b/java-client/src/generated/java/org/opensearch/client/opensearch/core/search/Hit.java
@@ -90,8 +90,8 @@ public class Hit<TDocument> implements PlainJsonSerializable, ToCopyableBuilder<
     @Nonnull
     private final Map<String, InnerHitsResult> innerHits;
 
-    @Nonnull
-    private final List<String> matchedQueries;
+    @Nullable
+    private final MatchedQueries matchedQueries;
 
     @Nonnull
     private final Map<String, JsonData> metaFields;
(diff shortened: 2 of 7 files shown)


reta (approved):

lawofcycles:
Thank you for reviewing!

Issue #2112, opened 2026-08-19
[BUG] currentSize is wrongly calculated in BulkInjester.flush()


gaobinlong:
### What is the bug?

See https://github.com/opensearch-project/opensearch-java/blob/86e3fd294020a1db2fc94d0e3a0dd2901c01482c/java-client/src/main/java/org/opensearch/client/opensearch/_helpers/bulk/BulkIngester.java#L381, 
 currentSize is the bytes size of the current buffered operations, but `operations` is a List, size() means the operations count.

### How can one reproduce the bug?

code bug

### What is the expected behavior?
Calculate the currentSize correctly.

PR #2113, opened 2026-08-19
Fix currentSize calculation in BulkIngester

gaobinlong:
### Description
`currentSize` in BulkIngester tracks the total byte size of buffered operations, used by canAddOperation() to enforce the maxSize flush threshold. After each flush, line 381 reset it incorrectly:

   // Before — wrong: operations.size() returns a count (e.g. 3), not bytes (e.g. 30,000)
   currentSize = operations.size();

This caused canAddOperation() to compare bytes against a count, making the maxSize-based flush trigger effectively broken — the buffer would almost never flush based on byte size alone.

The fix has two parts:

   1. Make IngesterOperation generic with Context

   IngesterOperation previously used a raw RetryableBulkOperation field, causing unchecked cast warnings and requiring callers to go through .repeatableOperation(). It is now IngesterOperation<Context> with typed fields and delegation methods (operation(), context(), isSendable(),canRetry(), retries()).

   2. Store IngesterOperation<Context> in the operations list

   Previously the list stored RetryableBulkOperation<Context> (without size). By storing IngesterOperation<Context> instead, each entry carries its precomputed byte size. currentSize is now decremented inline as operations are removed during flush:

   // After — correct: subtract the actual byte size of each sent operation
   currentSize -= op.size();

### Issues Resolved
https://github.com/opensearch-project/opensearch-java/issues/2112

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Fixed
 - Fix `unitTest` task not running the tests in the `test` source set ([#2074](https://github.com/opensearch-project/opensearch-java/pull/2074))
 - Run model tests against both JSON mappers instead of picking one at random ([#2085](https://github.com/opensearch-project/opensearch-java/pull/2085))
+- Fix currentSize calculation in BulkIngester
 
 ## [Unreleased 3.x]
 ### Added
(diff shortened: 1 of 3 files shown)


reta (approved):
Thanks @gaobinlong , this is cherrypick of https://github.com/elastic/elasticsearch-java/pull/1167, right?

gaobinlong:
> Thanks @gaobinlong , this is cherrypick of [elastic/elasticsearch-java#1167](https://github.com/elastic/elasticsearch-java/pull/1167), right?

Not fully, made small improvement that add a public retries() method into IngestOperation to simplify code.

PR #2149, opened 2026-09-13
Make HitsMetadata.hits optional to handle empty or filtered responses

### LINKED ISSUE #1729: [BUG] Getting 0 results with filter_path fails to deserialize
### What is the bug?
Not 100% sure this is a bug or a missing feature, or if theres an known workaround, but when `filter_path` is provided on `/_search` to omit any fields in the `_source`, an empty query result will fail to deserialize with `org.opensearch.client.util.MissingRequiredPropertyException: Missing required property 'HitsMetadata.hits'` because the `hits.hits` array is missing from the response

Make a query to `{indexName}/_search?filter_path=*,-hits.hits._source.someField` to only exclude `someField` on the `_source` doc, with a query body that results in no hits. The client is constructed as:

```
new OpenSearchClient(ApacheHttpClient5TransportBuilder
                .builder(HttpHost.create(endpoint))
                .build())
``` 

Presumably using the default JSON mapper (`JacksonJsonpMapper`)

The opensearch response I'm getting is:

```
{
  "took": 6,
  "timed_out": false,
  "_shards": {
    "total": 2,
    "successful": 2,
    "skipped": 0,
    "failed": 0
  },
  "hits": {
    "total": {
      "value": 0,
      "relation": "eq"
    },
    "max_score": null
  }
}
```

Which fails to deserialize with `org.opensearch.client.util.MissingRequiredPropertyException: Missing required property 'HitsMetadata.hits'` 

Ideally the missing array property would be deserialized into an empty array, is there a workaround for this?

Client version: `org.opensearch.client:opensearch-java:3.2.0`, on Java 21

Opensearch is running in a docker container: `opensearchproject/opensearch:3.2.0`

This is the calling code (in Kotlin) - the query is contructed as a string, but there's no issue deserializing when there are some results

```
    val jsonpMapper = client._transport().jsonpMapper()

    val searchResponseDeserializer = SearchResponse.createSearchResponseDeserializer(
        JsonpDeserializer.of(MySearchDocument::class.java)
    )
    
    val request = Requests.builder()
        .endpoint("/$indexName/_search?typed_keys=true&filter_path=-hits.hits._source.omittedField")
        .method("POST")
        .json(finalQuery)
        .build()

    val responseBody: Body? = client.generic().execute(request).body.getOrNull()

    try {
        val searchResponse: SearchResponse<MySearchDocument> = responseBody
            ?.let { b -> Bodies.json(b, searchResponseDeserializer, jsonpMapper) }
            ?: throw RuntimeException("Failed to parse search response")
        return searchResponse
    } catch (e: RuntimeException) {
        logger.error(e) { "OpenSearch: Failed to search: ${responseBody?.bodyAsString()}" }
        throw e
    }
```

elgamal-ahmed:
Closes #1729.

A search that matches nothing, or that uses filter_path to drop hits.hits, returns a hits object with no hits array, which failed to deserialize because HitsMetadata.hits was generated as a required property. This marks the property optional in the codegen overrides and regenerates the client. The underlying cause is the OpenAPI spec marking hits required, so correcting it in the override layer is the in-repo fix; the getter stays non-null and returns an empty list when the array is absent, so this is not a breaking API change.

Tests: added HitsMetadataTest covering deserialization of a response with and without the hits array (against both JSON mappers) and that an unset list is not serialized. ClassStructureTest used HitsMetadata as its example of a required list, so that assertion now uses Profile, whose only required field is a list.


reta:
Thanks for the pull request @elgamal-ahmed , the issue with `filter_path` is that **any** required property could be filtered out (and essentially become non-required) (by the specification [1], `hits` is a required property and has to be present, so the client does a right thing here). To summarize - we need to find a way how to deal with  `filter_path` , not changing the ad-hoc properties optionality.

> A search that matches nothing

Empty response must include empty `hits` array and should be properly deserialized

elgamal-ahmed:
Good point, thanks. You're right that a genuine zero-hit response includes an empty `hits` array, and that `filter_path` can prune any required property, not just this one, so changing the optionality of `hits` was the wrong lever. I reverted it: `hits` is required again, and a response missing `hits.hits` still fails deserialization by default. There is also a new test showing that a genuine zero-hit response deserializes as before.

For `filter_path`, the client already has an escape hatch, `ApiTypeHelper.DANGEROUS_disableRequiredPropertiesCheck`, but it crashed on exactly these responses. The PR now makes it work:

- A missing required list or map becomes an empty undefined list or map instead of throwing a `NullPointerException`. `isDefined()` returns false for it, so callers can still tell a pruned array from a real empty one.
- A missing required primitive such as `took` or `timed_out` gets its default value instead of failing on unboxing, as the javadoc of the escape hatch already says. The javadoc now also warns that this default can't be told apart from a real value.

For the case in #1729, which uses the generic client and decodes the response on the caller's thread:

```java
try (ApiTypeHelper.DisabledChecksHandle h = ApiTypeHelper.DANGEROUS_disableRequiredPropertiesCheck(true)) {
    SearchResponse<MyDoc> response = Bodies.json(body, searchResponseDeserializer, mapper);
}
```

I kept typed-client support out of this PR. Generated request classes don't expose `filter_path`, and the Apache HttpClient 5 and RestClient transports ignore query parameters set through `TransportOptions` (`AwsSdk2Transport` does forward them). Also, the flag only applies where the response is decoded on the caller's thread. `ApacheHttpClient5Transport`, even for sync calls, and `AwsSdk2Transport` with an async SDK client decode on another thread. I'm happy to follow up on either, for example typed-client support for `filter_path` that relaxes the checks for those requests only, if that direction sounds right to you.

(An earlier version of this comment described a new mapper attribute. I took it out to keep this PR to the fix itself.)

PR #2150, opened 2026-09-13
Add output_format option to date ingest processor

### LINKED ISSUE #1512: [BUG] There is no "output_format" in DateProcessor ingest pipeline
### What is the bug?
As stated in the document: https://opensearch.org/docs/latest/ingest-pipelines/processors/date/
There should be a property for `output_format` on DateProcesser ingest on opensearch-java library but there is none.

use latest library from mvn:
```xml
<dependency>
    <groupId>org.opensearch.client</groupId>
    <artifactId>opensearch-java</artifactId>
    <version>2.22.0</version>
</dependency>
```

How can I create a pipeline like this in java?
```json
{
  "description": "Convert date string to epoch millis",
  "processors": [
    {
      "date": {
        "field": "start_time",
        "formats": ["date_time"],
        "output_format": "epoch_millis",
        "target_field": "start_time"
      }
    }
  ]
}
```

should support `output_format` for the DateProcessor

Is there at least a workaround for this?

elgamal-ahmed:
Adds the missing `output_format` option to the `date` ingest processor so pipelines that set it can be modeled by the client. Closes #1512.

`output_format` is an optional date-format string the processor applies when writing the parsed date to the target field (default `yyyy-MM-dd'T'HH:mm:ss.SSSZZ`), documented for the date processor but absent from the generated `DateProcessor`. I added it as an optional string to `ingest._common.DateProcessor` in the bundled `java-codegen/opensearch-openapi.yaml` and regenerated the client with `./gradlew :java-codegen:run`, so the generated change is limited to `DateProcessor.java`. Heads up that the same field is also missing from the canonical `opensearch-api-specification`, so it will need to be added there too for the fix to survive the weekly spec sync.

Tests: added `DateProcessorTest` covering serialization and round-trip deserialization with and without `output_format`, plus a `toBuilder` copy check; all pass under both JSON mappers.


java-codegen/opensearch-openapi.yaml
```diff
@@ -60418,6 +60418,9 @@ components:
                 The locale to use when parsing the date, relevant when parsing month names or week days.
                 Supports template snippets.
               type: string
+            output_format:
```
reta, line 60421: Please submit a change against  https://github.com/opensearch-project/opensearch-api-specification, [1]

[1] https://github.com/opensearch-project/opensearch-api-specification/blob/main/spec/schemas/ingest._common.yaml#L239

reta (changes requested):

elgamal-ahmed:
Makes sense. I've opened opensearch-project/opensearch-api-specification#1238 to add `output_format` to the date processor schema upstream. Closing this in favor of that one. Thanks for the pointer.

PR #2160, opened 2026-09-20
Thread document type through gRPC typed search (fix ClassCastException)

### LINKED ISSUE #2123: gRPC transport: typed search silently deserializes hits as Object.class, causing ClassCastException on source access
## Summary

`GrpcTransport.performSearch()` discards the caller's document type entirely: it always deserializes hit `_source` values with `Object.class` and then unchecked-casts the result to `SearchResponse<TDocument>`. Unlike the REST transport - where `SearchRequest._ENDPOINT` instances carry a response deserializer bound to the concrete document class passed to `client.search(request, MyType.class)` - the gRPC path returns hits whose sources are generic `JsonData`/Map values. Any typed access to `hit.source()` fails at runtime with `ClassCastException`.

Static-analysis finding against current `master`; not executed here.

## Location

- File: `java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/GrpcTransport.java`
- Function: `performSearch(SearchRequest request)` (~lines 275-297):

```java
// Convert response — use Object.class as default; the actual deserialization
// is handled by the endpoint's response deserializer in the transport layer
return (SearchResponse<TDocument>) ...SearchResponseConverter.fromProto(
    protoResponse,
    jsonpMapper,
    (Class<TDocument>) Object.class
);
```

- Materialization point: `translation/SearchResponseConverter.deserializeSource()` (~line 168-172) calls `jsonpMapper.deserialize(parser, tDocumentClass)` — with `Object.class` this produces `JsonData`, and nothing re-deserializes afterwards.

## Problem

Two claims in the code do not hold together:

1. The comment states "the actual deserialization is handled by the endpoint's response deserializer in the transport layer". But `performSearch` receives only the `SearchRequest`; the `Endpoint` (which is exactly what carries the typed response deserializer) is dropped in `performRequest()` at ~line 146 (`performSearch((SearchRequest) request)`). No later stage exists that would convert the already-materialized generic hits into `TDocument`.
2. Because generics are erased, `(SearchResponse<TDocument>)` succeeds silently; the type error surfaces later as a `ClassCastException` inside user code iterating `response.hits().hits()` and calling methods on `source()`.

The REST transport honors the typed contract because `SearchRequest._ENDPOINT` used by `OpenSearchClient.search(req, clazz)` embeds `clazz` in its response parser. The gRPC branch bypasses it.

## Trigger / Reproduction

Based on static analysis; no runtime run performed:

```java
OpenSearchTransport grpcTransport = GrpcTransport.builder(...).build();
OpenSearchClient client = new OpenSearchClient(grpcTransport);

SearchResponse<Product> resp = client.search(s -> s.index("products")
        .query(q -> q.matchAll(m -> m)), Product.class);

for (Hit<Product> hit : resp.hits().hits()) {
    Product p = hit.source();   // compiles; actually a JsonData
    p.getName();                // ClassCastException here
}
```

## Expected Behavior

Either:
- extract the document class from the endpoint's response deserializer (or from the `SearchRequest`'s typed deserializer reference) and pass it to `SearchResponseConverter.fromProto`, or
- if typed search is not yet supported over gRPC, restrict `isEndpointSupported()` so typed searches fall back to REST under `HybridTransport`, and/or document the limitation loudly instead of failing late with CCE.

## Actual Behavior

Silent success with wrongly-typed contents; failure deferred to first element access in user code.

## Impact

Every gRPC-transport user of the strongly-typed search API (the primary usage pattern shown in the client docs) gets runtime `ClassCastException`s instead of mapped documents. This makes the new `java-client-grpc` module unusable for typed search workloads without users knowing why, since compile-time types look correct.

## Suggested Direction

Thread the document class through: e.g. have `performRequest` detect `SearchRequest._ENDPOINT`'s `responseDeserializer` (opensearch-java endpoints expose their `JsonpDeserializer`) or add an overload of `performSearch(SearchRequest, Endpoint)` that pulls `tDocumentClass` from the deserializer when it is a `ObjectDeserializer<TDocument>`-style instance, defaulting to `JsonData` with a logged warning otherwise.

## Evidence

- Hard-coded `Object.class` + unchecked cast quoted above.
- `SearchResponseConverter.fromProto/deserializeSource`: single-pass materialization using the passed class; no deferred conversion exists.
- `performRequest()` routes solely on request identity and drops the `Endpoint` argument before calling `performSearch`.

harrisleesh:
### Description

The gRPC transport's typed search path discarded the caller's document type. `GrpcTransport.performRequest` dropped the `Endpoint` (which carries the response deserializer bound to the concrete `TDocument`) and `performSearch` always deserialized hit `_source` values with a hard-coded `Object.class`, then unchecked-cast the result to `SearchResponse<TDocument>`. This meant a strongly-typed `client.search(request, MyType.class)` over gRPC could never return properly typed hits — `hit.source()` was actually a generic `JsonData`/`Map`, surfacing as a `ClassCastException` at first field access in user code.

Compounding this, routing matched the endpoint by identity against the raw `SearchRequest._ENDPOINT`, but `OpenSearchClient.search(request, clazz)` passes a wrapping `EndpointWithResponseMapperAttr`. So a typed search over a pure `GrpcTransport` actually threw `UnsupportedOperationException` (and under `HybridTransport` silently fell back to REST), meaning typed gRPC search never really worked.

This PR threads the caller's document deserializer through the gRPC search path, the same way the REST endpoint does:

- `GrpcTransport` now unwraps `DelegatingJsonEndpoint` wrappers to match the underlying registered endpoint by identity (fixes routing for wrapped/typed endpoints), and extracts the `JsonpDeserializer<TDocument>` that `client.search(..., clazz)` stores on the endpoint, passing it to `performSearch`.
- `SearchResponseConverter` deserializes hit `_source` values with that deserializer instead of `Object.class`. A `Class<TDocument>` overload is retained for convenience and falls back to `Object`/`JsonData` when no typed deserializer is present, preserving prior behavior for untyped callers.
- Small additive accessors were added to `DelegatingJsonEndpoint` (`endpoint()`) and `EndpointWithResponseMapperAttr` (`attrName()`, `attrValue()`) so the transport can recover the wrapped endpoint and its response-mapper attribute.

### Issues Resolved

Fixes #2123

### Testing

Added `GrpcTypedSearchTest`, which stands up an in-process gRPC `SearchService` returning a canned hit and drives a real `client.search(s -> ..., Movie.class)` end-to-end, asserting `hit.source()` is a `Movie` with the expected fields.

- Before the fix: fails (the wrapped typed endpoint is rejected by routing → `UnsupportedOperationException`; even when reached, `_source` deserializes to `JsonData` → `ClassCastException`).
- After the fix: passes; hits deserialize into the requested document type.

`./gradlew :java-client-grpc:test` passes (full module, no regressions).

### Check List
- [x] New functionality includes testing.
- [x] New functionality has been documented.
- [x] Commits are signed per the DCO using `--signoff`.


java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/GrpcTransport.java
```diff
@@ -60,11 +69,24 @@ public class GrpcTransport implements OpenSearchTransport {
...
     }
 
+    /**
+     * Unwraps any {@link DelegatingJsonEndpoint} wrappers (such as the typed
+     * {@link EndpointWithResponseMapperAttr} produced by {@code client.search(request, MyType.class)})
+     * to recover the underlying registered endpoint so it can be matched by identity.
+     */
+    private static Endpoint<?, ?, ?> rootEndpoint(Endpoint<?, ?, ?> endpoint) {
+        Endpoint<?, ?, ?> current = endpoint;
+        while (current instanceof DelegatingJsonEndpoint) {
```
reta: This was a miss, thanks @harrisleesh

java-client-grpc/src/main/java/org/opensearch/client/transport/grpc/GrpcTransport.java
```diff
@@ -139,12 +186,14 @@ public <RequestT, ResponseT, ErrorT> ResponseT performRequest(
...
+        // Route to the appropriate gRPC handler. The caller may have wrapped the endpoint (e.g. the
+        // typed search endpoint carries the document deserializer), so match on the underlying endpoint.
+        Endpoint<?, ?, ?> root = rootEndpoint(endpoint);
+        if (root == BulkRequest._ENDPOINT) {
             return (ResponseT) performBulk((BulkRequest) request);
         }
-        if (endpoint == SearchRequest._ENDPOINT) {
-            return (ResponseT) performSearch((SearchRequest) request);
+        if (root == SearchRequest._ENDPOINT) {
+            return (ResponseT) performSearch((SearchRequest) request, searchDocumentDeserializer(endpoint));
```
reta: I think we could explore a simple option here:
 - we know we handle search response and we could use `NamedDeserializer<TDocument>`
```
           final NamedDeserializer<?> deserializer = new NamedDeserializer<>(
                "org.opensearch.client:Deserializer:_global.search.TDocument"
            );
            return (ResponseT) performSearch((SearchRequest) request, deserializer);
```
- the `NamedDeserializer<TDocument>` would need to expose underlying named `JsonpDeserializer<T>`, we could add a method there:

```
public class NamedDeserializer<T> implements JsonpDeserializer<T> {
     ....
     public JsonpDeserializer<T> named(JsonpMapper mapper) {
        return mapper.attribute(name);
    }
}
```
- and that's about it, `performSearch` would need to change `documentDeserializer` -> `namedDeserializer.named(jsonpMapper)` 

WDYT? It is still not the best option, but much less changes and public API exposure

java-client-grpc/src/test/java/org/opensearch/client/transport/grpc/GrpcTypedSearchTest.java
```diff
@@ -0,0 +1,128 @@
...
+import org.opensearch.client.opensearch.core.search.Hit;
+import org.opensearch.protobufs.services.SearchServiceGrpc;
+
+/**
+ * Verifies that a strongly-typed search issued through {@link OpenSearchClient#search} over the
+ * gRPC transport deserializes hit {@code _source} values into the caller-supplied document type
+ * (rather than a generic {@code Object}/{@code JsonData}), i.e. that the {@code Class<TDocument>}
+ * contract is honored end-to-end. Reproduces issue #2123.
+ */
+public class GrpcTypedSearchTest {
```
reta: We don't need this test, `GrpcSearchIT` should have covered it, we need to improve it a bit to test hybrid and grcp transports together (we miss this issue due to hybrid  -> rest fallback, but would have not using grpc transport directly)

java-client-grpc/build.gradle.kts
```diff
@@ -53,6 +53,7 @@ dependencies {
 
     // Test dependencies
     testImplementation("io.grpc", "grpc-testing", grpcVersion)
+    testImplementation("io.grpc", "grpc-inprocess", grpcVersion)
```
reta: No need (see please https://github.com/opensearch-project/opensearch-java/pull/2160/changes#r4094673842)

